PR #26, opened 2017-04-12
fix DOCKER_DEPLOY env var, .env logic

milescrabill:
@peterbe r?

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -2,7 +2,7 @@
 set -eo pipefail
 
 # copy the .env template to .env if not there already
-[ ! -f .env ] && cp .env-dist .env
+[ ! -f .env ] || cp .env-dist .env
 
 # default variables
 export DEVELOPMENT=1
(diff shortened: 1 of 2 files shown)


bin/test
```diff
@@ -2,7 +2,7 @@
 set -eo pipefail
 
 # copy the .env template to .env if not there already
-[ ! -f .env ] && cp .env-dist .env
+[ ! -f .env ] || cp .env-dist .env
```
peterbe: That doesn't make sense. No?
The first part says "Does .env NOT exist?"
It should be `&&` meaning, "If it does NOT exist, then copy from .env-dist"

circle.yml
```diff
@@ -8,8 +8,7 @@ machine:
 
 dependencies:
   pre:
-    # should just set env vars in circleci
-    - touch .env
+    - cp .env-dist .env
```
peterbe: Now I'm confused. By GitHub. 
[That line is already in there](https://github.com/mozilla/tecken/blob/master/circle.yml)

PR #28, opened 2017-04-12
healthcheck that redis store is LRU configured

peterbe:


peterbe:
@jezdez is the right way to write my own custom healthchecks for dockerflow?

peterbe:
Obsoleted by https://github.com/mozilla/tecken/pull/36

jezdez:
@peterbe Still answering since I simply was on holidays... healthchecks are useful for the loadbalancer to check if the web app works. If your web app can't work without running in allkeys-lru mode then this is the right place to put it. OTOH, I would argue that this should be written as a [Django system check](https://docs.djangoproject.com/en/1.11/topics/checks/) (marked as "deploy time only") and then calling `django-admin check --deploy` [during deploy](https://docs.djangoproject.com/en/1.11/ref/django-admin/#check). That the healthcheck system in dockerflow also uses those is just a matter of re-using an already existing API.

PR #36, opened 2017-04-18
configure redis store's eviction policy on startup

peterbe:


tecken/apps.py
```diff
@@ -19,3 +20,15 @@ def ready(self):
...
 
         markus.configure(settings.MARKUS_BACKENDS)
+
+        connection = get_redis_connection('store')
+        maxmemory_policy = connection.info()['maxmemory_policy']
+        if maxmemory_policy != 'allkeys-lru':
+            logger.warning(
+                "The 'store' Redis cache was not configured to be an "
+                "LRU cache because the maxmemory-policy was set to '{}'. "
+                "Now changed to 'allkeys-lru'.".format(
```
jezdez: Here I wouldn't state, that it has already been changed when the actual change only happens in the next line which may still raise an error if the Redis server doesn't accept a config set call. IIRC some Redis hosters don't allow this type of change in favor of web ui config changes.

peterbe: You are right. I'm glad I tested this. In ElastiCache Redis you can't send any `CONFIG` commands. Not even to *read* config. 
So I'm going to document how you have to do it on the infra side. 
The code here above will be changed slightly so it only tries to do this if `settings.DEBUG`.

peterbe:
@jezdez You were totally right! `config` is not allowed in ElastiCache. But we've already figured out how to do it with infra ops. And I've vaguely documented it even though I don't really know how Cloud OPs really automate this. 

Also, I changed the code so that it corrects the config only in docker when `settings.DEBUG` is true and if it not it sends a message on `logger.error()`. 

This feels right now.

PR #127, opened 2017-05-31
fixes bug 1367513 - add mozilla-django-oidc

peterbe:


willkg:
I wouldn't use mozilla-django-oidc 0.10. I'd wait until they do a new release which includes a ton of fixes.

peterbe:
@willkg Is the current one insecure? 
The current one works after all.

willkg:
I don't think I've fixed any security issues, but I've fixed a ton of other things. I'm holding off on switching Standup and Socorro until I'm done with mozilla-django-oidc work and they've done a release.

peterbe:
Sigh. I wrote a long comment but it disappeared when I moved this window. Oh Nightly!

There are bugs. I was going to list them. Including the landed patches that haven't been released. 

To be able to regain momentum on Symbol Upload I just need to be able to sign in. Once mozilla-django-oidc gets better I'll upgrade to that too.

PR #147, opened 2017-06-07
Fixes bug 1370960 - Copy .env-dist if .env doesn't exist

willkg:
This copies the .env-dist file over if .env doesn't exist. It also adds a header
indicating which .env-dist got copied which is helpful sometimes.

r?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,55 +22,58 @@ help:
 .docker-build:
 	make build
 
-build:
+.env:
+	./bin/cp-env-file.sh
+
+build: .env
 	docker-compose build deploy-base
 	docker-compose build dev-base
 	touch .docker-build
 
-clean: stop
+clean: .env stop
 	docker-compose rm -f
 	rm -rf coverage/ .coverage
 	rm -fr .docker-build
 
-migrate:
+migrate: .env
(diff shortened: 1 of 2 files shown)


willkg:
Yup! Oops.

willkg:
^^^ That adds the missing script.

peterbe:
There's probably a lot more we can do. For example; perhaps instead of copying it should say 
"Error! You haven't made your own .env file yet. Copy the .env-dist to .env and fill it in with ..."
But I'm not totally clear what needs to be done. Because we use localstack you don't need credentials. But there are other things like managing your own Auth0 credentials etc. 

Also, I think the `.env-dist` file is a bit messy as is. There are a bunch of things in there that is best to leave as defaults in `settings.py` instead. For example, if you `cp .env-dist .env`, today, you're stuck with certain defaults like `ALLOWED_HOSTS` which doesn't really need to be set by a human developer. Perhaps the file should only mention AWS, OIDC and Sentry credentials. And perhaps there should be a big fat comment that says "To change anything from the default in 'settings.py' type it into this file and prefix it with DJANGO_".

peterbe:
/me filed https://bugzilla.mozilla.org/show_bug.cgi?id=1371365

PR #173, opened 2017-06-15
bug 1369733 - default to 3 gunicorn workers

peterbe:


milescrabill:
I'm inclined not to do it this way because it makes it more difficult for ops to configure. The command line argument to Gunicorn takes precedence over all over forms of configuration, and doing it this way (not set with an env var, for example) requires a code change in tecken in order to adjust worker counts. Preferably we would drop this `--workers` from the command line args and I will set it in my host configuration.

peterbe:
For the record, I intent to overhaul that almost entirely. The end result is not to come up with a different number than 3 or 4, but a bunch of other parameters. Similar to [antenna](https://github.com/mozilla-services/antenna/blob/master/bin/run_web.sh). And surely that's a good thing to keep in code so that docker runtime (when you don't run with DEBUG and Django's runserver) exhausts the same workers and configuration. 

The intent here was just to correct the default as a stop gap right before we start load testing the first batch. 

I still support the idea of (ab)using environment variables though. Suppose that we end up preferring `gunicorn --worker-elfs=89` we should probably code that as `gunicorn --worker-elfs="${SOME_VAR:89}"` or some such.

peterbe:
@milescrabill 
I know this doesn't make all gunicorn configurations configurable by environment variables. Partly because I don't know what the ideal way of writing that bash-fu and partly because I kinda like that the default is set in code. For example, the `--worker-class` stuff. I think that is nice to have as a hardcoded default. 

Also, I haven't tested this. I just did what's done with other variables in that file.

r?

PR #196, opened 2017-06-27
fixes bug 1367456 - Filename pattern enforcement in uploads

peterbe:


peterbe:
why isn't circle running?

PR #214, opened 2017-07-12
fixes bug 1106313 - fetch missing symbols from Microsoft

peterbe:
@luser I'm not expecting a full review but I'm going to highlight some areas (with comments) where I'd love a second pair of eyes.


docker/build_dump_syms.sh
```diff
@@ -0,0 +1,22 @@
...
+export CXX="g++-4.8" CC="gcc-4.8"
+
+CXXFLAGS=-O2 gyp -f ninja --depth=. ./dump_syms.gyp
+ninja -C out/Default
+
+ls -l out/Default
+
+# Put the final binaries in /stackwalk in the container
+mkdir -p /dump_syms
+cp out/Default/dump_syms /dump_syms/
```
peterbe: I took this from your `.travis.yml` and some tips from @willkg 's code where he installs stackwalker in socorro. 
Ok?

docs/download.rst
```diff
@@ -65,6 +65,46 @@ which used to manage reporting symbols that can't be found during
 processing. See https://bugzilla.mozilla.org/show_bug.cgi?id=1361809
 
 
+Microsoft on-the-fly Symbol Lookups
+===================================
+
+Under certain conditions, if a symbol can not be found in S3, we might
+try to look it up from Microsoft's download server
+(``https://msdl.microsoft.com/download/symbols/``) if the symbol file
+ends in ``.pdb`` and filename ends in ``.sym``.
```
peterbe: Is this correct?
It filters out a bunch of files that are called things like `foo.pdb/HEXHEX/foo.dll` which I see a lot of in the S3 prod logs.

tecken/download/tasks.py
```diff
@@ -0,0 +1,192 @@
...
+            if std_err and not std_out:
+                raise DumpSymsError(
+                    f'dump_syms extraction failed for {url}. '
+                    f'Error: {std_err!r}'
+                )
+
+        # Let's go ahead and upload it now, if it hasn't been uploaded
+        # before.
+        upload_microsoft_symbol(symbol, debugid, std_out)
+
```
peterbe: All this meat above is basically a fancy refactoring of https://gist.github.com/luser/92d5bc88478665554898

One important feature difference, that your gist didn't get right, is that it forces the generated `.pdb` file to be in lowercase. If you do `cabextract -d /tmp FOO.pd_` it will create `/tmp/foo.pdb`. Not `/tmp/FOO.pdb` so you can't just assume the same name by replacing the `_` suffix with a `d`.

tecken/download/views.py
```diff
@@ -80,7 +107,27 @@ def download_symbol(request, symbol, debugid, filename):
...
+
+            # The querying of Microsoft's server is potentially slow.
+            # That's why this call is down in a celery task.
+            # But there is hope! And the client ought to be informed
+            # that if they just try again in a couple of seconds/minutes
+            # it might just be there.
+            delayed_lookup = True
+
+    response = http.HttpResponseNotFound(
+        'Symbol Not Found Yet' if delayed_lookup else 'Symbol Not Found'
```
peterbe: What do you think about this crazy idea (thanks @lonnen for inspiring this idea).

Instead of responding with `404` we can instead respond with `302` (or some other 3xx) code. But when we set the `Location` header we add a little extra to the URL like `&was-recently-put-in-background-task=1` and when the client comes back to us again, this time we put in a little sleep so that `curl` doesn't fail on an "infinite" redirect loop **too quickly**. 

It would be nice because then things like stackwalker wouldn't need to do anything smarter to retry based on the output.

tecken/upload/models.py
```diff
@@ -52,13 +52,21 @@ class FileUpload(models.Model):
...
+    upload = models.ForeignKey(Upload, null=True)
     bucket_name = models.CharField(max_length=100)
     key = models.CharField(max_length=300)
     # True if this overwrote an existing key
     update = models.BooleanField(default=False)
     # True if the file was gzip compressed before being uploaded
     compressed = models.BooleanField(default=False)
     size = models.PositiveIntegerField()
+    # Used when this upload belongs to a Microsoft proxy download
+    microsoft_download = models.BooleanField(default=False)
```
peterbe: This way we can later, easily, query how many/which symbol uploads came from this Microsoft background task thing.

PR #238, opened 2017-07-18
build frontend separately in Circle

peterbe:


peterbe:
@milescrabill What do you think about this?
The frontend static files are not needed to do Docker development. It's a build artifact that is only needed *after* the tests have passed and *before* the deployment. 

I ran a Circle with SSH enabled and verified that the files got created:
```
(venv-system) ubuntu@box543:~$ ls tecken/frontend/build/
asset-manifest.json  favicon.ico  index.html  manifest.json  service-worker.js	static
```

peterbe:
@jgmize So I think this is actually a good stepping stone in the right direction. 
Ideally it would be better to have a dedicated Dockerfile just for building the static files (i.e. `cd frontend && yarn run build && ls -l build`) but that's perhaps something we can do later. 
I still kinda like the idea of executing the static file building in Circle because it's automated and it's a step that's only ever (beyond some debugging scenarios maybe) needed right before taking the build artifacts into production. 

Also, it would be awesome to get rid of [this](https://github.com/peterbe/tecken/blob/cb26e42b340f986dd4820dd911f91e1f9d50fe0e/Dockerfile#L37-L99) and keep the Dockerfile "purer" and simpler. By simpler I mean being able to just point to a trusted node+yarn container on dockerhub or something like that.

PR #242, opened 2017-07-19
bug 1381621 - frontend docker container

peterbe:


peterbe:
@jgmize Arh! I completely forgot to ask you for a review on this. It's basically everything you thought me plus some cleaning up. 

I'm really happy with how this works. I can't think of anything I'd like to change. 

However, the thing that still really sucks is [the ugliness of installing node inside a Python docker image](https://github.com/mozilla-services/tecken/pull/238#issuecomment-316440353). If you can think of a better way to do that, I'm aaaaaall ears.

PR #269, opened 2017-07-28
fixes bug 1385261 - Unnecessary containers is started in Circle

peterbe:


docker-compose.yml
```diff
@@ -48,9 +48,6 @@ services:
       - db
       - redis-store
       - redis-cache
-      - statsd
-      - localstack-s3
-      - frontend
     command: web-dev
```
willkg: Instead of doing this, can you create a `test` service like this?:

```
test:
  extends:
    service: base
  ports:
    - "8000:8000"
  links:
    # Only add services we need to run the tests.
    - db
    - redis-store
    - redis-cache
```

Then you can change `make run` back to what it originally was and change the test script to run in the `test` container:

```
docker-compose run -e DEVELOPMENT -e DJANGO_CONFIGURATION $CI_ENV test test $@
```

peterbe: Thanks!
See the followup commit I made. 

Now, when I `make run` it starts `localstack-s3` and `statsd` as daemons in the background (just like postgres and redis). 
But when I run `make test` it only starts the absolutely minimal. Now `make test` runs a lot faster than before. 
Best of both worlds.

PR #374, opened 2017-09-06
circleci 2.0, remove staticfrontend and dev containers, add dependencies on db to docker-compose

milescrabill:
this is still WIP, but has some of the ideas I was talking about in https://bugzilla.mozilla.org/show_bug.cgi?id=1393465

thoughts so far?


milescrabill:
Multi-stage builds require Docker v17.05.0-ce.

docker-compose.yml
```diff
@@ -124,6 +121,7 @@ services:
       - "35729:35729"
     volumes:
       - $PWD/frontend:/app
+    command: "yarn start | cat"
```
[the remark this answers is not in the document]
milescrabill: We can still do multiline bash scripts in the docker-compose.yml with comments:

```
command: >
         bash -c "# this is a comment
                  echo 'the code can keep going'"
```

I agree that cryptic one-liners are bad, but I think that using external bash scripts in a docker-compose configuration that will only be used for development can create confusion and contributes to a nest of scripts in the bin directory.

Dockerfile
```diff
@@ -1,3 +1,18 @@
+FROM node:6 as frontend
+
+# these build args are turned into env vars
+# and used in bin/build_frontend.sh
+ARG FRONTEND_SENTRY_PUBLIC_DSN=UNSET_DSN
+ENV FRONTEND_SENTRY_PUBLIC_DSN=${FRONTEND_SENTRY_PUBLIC_DSN}
+ARG CI=false
+ENV CI=${CI}
```
peterbe: This repetition is gross but I guess necessary evil. Well done for figuring it out!

Dockerfile
```diff
@@ -1,3 +1,18 @@
+FROM node:6 as frontend
```
peterbe: So, does this beast come with a decent and recent version of yarn? 

One thing that strikes me as a novice is that it's there's no comments. 
The appropriate comment could be,
```
First, build the frontend code static assets based on Node 6. The culmination 
is the execution of `bin/build_frontend.sh`. 
```

tecken/settings.py
```diff
@@ -542,6 +542,8 @@ def post_setup(cls):
 
     @property
     def VERSION(self):
+        # this was breaking in ci
+        return {}
```
peterbe: That's ok with me. I'll clean it up later.

Dockerfile
milescrabill: as of the current: https://github.com/nodejs/docker-node/blob/7701eea6fe125530894b3e83f1e9d385e9ee509f/6.11/Dockerfile#L41

Dockerfile
milescrabill: unfortunately, I believe this is indeed a necessary evil.

PR #393, opened 2017-09-12
bug 1399140 - longer gunicorn timeout

peterbe:
The default is 30s. So not only does this make it configurable, it also changes the default to 60 seconds.


peterbe:
Let's sit on this for a while.

peterbe:
@milescrabill Actually this little code change isn't bad. I don't like that we have to do this but in hindsight it's actually not a bad thing to have. 
I've changed the default so that it's now the same as Gunicorn's default. So nothing changes but as of this change we can at least control it. 
r?

PR #533, opened 2017-10-30
copy static assets build dir correctly

milescrabill:


milescrabill:
So, I `docker pull`'d the latest `mozilla/tecken:latest` and looked for `/app/frontend/build`:

```
docker run -it mozilla/tecken:latest /bin/bash
app@bb65da2800ef:~$ ls /app/frontend/
README.md  asset-manifest.json	favicon.ico  index.html  manifest.json	package.json  public  service-worker.js  src  static  yarn.lock
```

Not there, but all of the files expected to be in that dir are instead in `/app/frontend/` given how the COPY was executed.

Given that, I'm not sure how it's working. It's possible that the the script that copies the frontend files out of the container is failing because `/app/frontend/build` doesn't exist, and so it isn't overwriting the contents of the dir that nginx is serving on the host.

PR #542, opened 2017-11-01
bug 1413683 - Check Prettier in CI

peterbe:


peterbe:
@milescrabill 
This isn't of great importance but I'm curious and if it can be done it's nice to have it done. 
Basically trying to enforce that all .js files are formatted with Prettier. The ultimate command is `docker-compose run frontend lint`. 

1) Can you take a quick look (avoiding all the .js files) to check my "docker syntax" and stuff.
2) How would I be able to start parallel jobs in Circle so that the linting is separated from the testing. The documentation examples look easy but there's all that docker install prerequisites I can't figure out other than having to copy it into a separate step in the circle yml file. Ideas?

milescrabill:
@peterbe to do this in parallel, I think we need to do something along the lines of "workflows" [0], which would require a bit of re-architecting of our circleci conf. I think we should absolutely do that, but we'll be blazing new ground, so there will be an R&D component to be had.

We'll also want to make sure we're appropriately caching and saving things in between jobs such that that doesn't end up taking even _longer_ [1].

[0] https://circleci.com/docs/2.0/workflows/
[1] https://circleci.com/blog/how-to-build-a-docker-image-on-circleci-2-0/

peterbe:
(iterating some irc discussion...)

Your impression is that it'd be non-trivial to turn this into a parallel workflow. Ok, let's not break new ground here yet then. It'd be a different story if two parallel'able workflow tasks didn't depend on so much docker situps. 

Beyond that, can you comprehend why the bash script can't be found? It works locally to `docker-compose run frontend lint` but perhaps my macOS volume mounting is what is so different from my laptop and circle. 

Also, Ideally I think the file`frontend/bin/run_frontend.sh` should `live in `bin/` with all the other `.sh` files. I just couldn't figure out how to get that to work. Any idea?

milescrabill:
Your instincts are good here. I think the volume mount is what's breaking things again. See: https://github.com/mozilla-services/tecken/blob/master/docker-compose.yml#L117.

PR #545, opened 2017-11-02
only volume mount explicitly

peterbe:
@milescrabill I think I figured it out. At least it all makes sense locally. 
If you run `docker-compose run test-ci test` it runs the `test-ci` directive from `docker-compose.yml`. So, that'll run with the files "as of the latest `make build`". 

To test this, I do:
```
$ export CI=true
$ make test
```
The tests pass. So I mess with, for example, `tests/test_api.py` and put in a `raise Exception("hi")` and run: `make test` again and the tests pass again. 

However, when I do this WITHOUT setting `export CI=true` the tests will re-read the host's file and the second time the tests fail because of the temporary `raise Exception("hi")` I put in.

### DIFF
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -10,7 +10,7 @@ export DJANGO_CONFIGURATION=Test
 
 # run docker compose with the given environment variables
 
-if [[ ! -z "${CI}" ]]; then
+if [[ -n "${CI}" ]]; then
     # pass CI env vars into docker containers for codecov submission
     echo "Getting Codecov environment variables"
     export CI_ENV=`bash <(curl -s https://codecov.io/env)`
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -8,18 +8,14 @@ set -eo pipefail
 export DEVELOPMENT=1
 export DJANGO_CONFIGURATION=Test
 
-
 # run docker compose with the given environment variables
 
 if [[ ! -z "${CI}" ]]; then
-    echo "IN CI!!!!!"
     # pass CI env vars into docker containers for codecov submission
     echo "Getting Codecov environment variables"
     export CI_ENV=`bash <(curl -s https://codecov.io/env)`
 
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -56,8 +56,18 @@ services:
...
     command: test
 
+  test-ci:
+    extends:
+      service: base
+    links:
+      - db
+      - redis-store
+      - redis-cache
+
```
peterbe: I tried to be "neater" here to not have to list the `links: ...` two times but docker-compose doesn't allow that. Now I violate the DRY principle but it's fine. 
Note also, that there's no need for a `command` in the `test-ci` directive.

pytest.ini
```diff
@@ -14,4 +15,4 @@ DJANGO_CONFIGURATION = Test
 
 
 [pytest-watch]
-ignore = frontend .git docs test-with-nginx
+ignore = frontend .git docs test-with-nginx .cache htmlcov
```
peterbe: These changes are actually irrelevant. I just noticed it whilst working on this. Sorry.

docker-compose.yml
jgmize: The explicit links are not necessary: https://docs.docker.com/compose/networking/#links

bin/test.sh
```diff
@@ -8,10 +8,14 @@ set -eo pipefail
 export DEVELOPMENT=1
 export DJANGO_CONFIGURATION=Test
 
-# pass CI env vars into docker containers for codecov submission
-[ ! -z ${CI+check} ] && \
-    echo "Getting Codecov environment variables" && \
+# run docker compose with the given environment variables
+
+if [[ ! -z "${CI}" ]]; then
```
jgmize: FYI you could replace `! -z` with the equivalent `-n` here

peterbe: Ohh... I never get these right. I always just google stackoverflow and trial-and-error till it works. It always gets tricky when things are set but set to an empty string.

PR #625, opened 2017-12-20
fixes bug 1426495 - Upgrade to Django 2.0

peterbe:
This won't pass until there's [a new version of mozilla-django-oidc](https://github.com/mozilla/mozilla-django-oidc/pull/203).


peterbe:
Note-to-self; locally I managed to get the whole test suite to pass if I install Django 2.0 **and** willkg's branch for mozilla-django-oidc. I haven't tried to run migrations or actually runserver yet.

peterbe:
Also, perhaps we should look to using the [new Django 2.0 URL pattern tricks](https://docs.djangoproject.com/en/2.0/releases/2.0/#simplified-url-routing-syntax).

PR #915, opened 2018-06-18
fixes bug 1433565 - document symbolication v5 symbolicated stacks

peterbe:


docs/symbolication.rst
```diff
@@ -110,7 +110,9 @@ What you get back is a JSON output that looks like this:
               {
                 "frame": 0,
                 "module_offset": "0xb2e3f7",
-                "module": "xul.pdb"
+                "module": "xul.pdb",
```
willkg: At the top of this section, it uses a url that ends with a `4`. Should it be 5 given this is v5?

Also, a few paragraphs above this, it says "The ``version`` key needs to be a 4." Is that still true?

peterbe: Ugh. A rushed job. This is what happens when projects are dead. I'll make a new PR.

peterbe: https://github.com/mozilla-services/tecken/pull/916

PR #925, opened 2018-06-21
optional 'm' after FUNC or PUBLIC, fixes #924

### LINKED ISSUE #924: Support symbol file format change in upstream Breakpad (optional 'm' after FUNC or PUBLIC)
Migrated from https://bugzilla.mozilla.org/show_bug.cgi?id=1470092

peterbe:
@luser The code kinda speaks for itself. I just made it so that the symbolication works with *and without* the 'm' after 'FUNC' and 'PUBLIC'. 
Left a code comment in there to note that some day we might actually use the 'm' in the symbolication.

### DIFF
--- a/tecken/symbolicate/views.py
+++ b/tecken/symbolicate/views.py
@@ -5,6 +5,7 @@
 import datetime
 import time
 import logging
+import re
 from bisect import bisect
 from functools import wraps
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


tecken/symbolicate/views.py
```diff
@@ -715,6 +716,17 @@ def load_symbol(self, filename, debug_id):
         t0 = time.time()
         stream = self.get_download_symbol_stream(filename, debug_id)
 
+        # We are only interested in lines that match this regex.
+        # The 'm' is special. It's an extra *optional* prefix which,
+        # we currently omit.
+        # Origin: https://bugzilla.mozilla.org/show_bug.cgi?id=1470092
```
luser: I would link the upstream Breakpad issue or changeset here instead:
https://bugs.chromium.org/p/google-breakpad/issues/detail?id=751
https://chromium.googlesource.com/breakpad/breakpad/+/b1226959a25b6a5311801d6f204b088c706e7c25

tecken/symbolicate/views.py
```diff
@@ -725,31 +737,41 @@ def load_symbol(self, filename, debug_id):
...
                         )
                     )
                     continue
-                address = int(fields[1], 16)
-                symbol = fields[4]
+                address = int(fields[0], 16)
+                symbol = fields[3]
                 func_symbols[address] = symbol
 
         # Prioritize PUBLIC symbols over FUNC symbols # XXX why?
```
luser: I noticed this comment while looking at the code. Is this actually true? If so this is the opposite of what we actually want (FUNC symbols should take priority over PUBLIC symbols.) If that's the case we should open a new issue to fix that separately.

peterbe: This is a relic from https://github.com/mozilla/Snappy-Symbolication-Server/blob/6a3347dc32d5ecaccaafd3219f341d7b7c2d0cf0/snappy/DiskCache_DiskCache.py#L380-L382 which I just copied (although I call the dict variables differently). So is this wrong? Then we should definitely start a new ticket.

PR #928, opened 2018-06-22
upgrade python:3.6-slim docker

peterbe:
It seems the world of Debian totally changed as of this upgrade to the python:3.6-slim docker image. There is no more gcc-4.8 or g++-4.8. Only gcc-6 and g++-6. 
I'm a bit out of my depth here but this PR seems to work. Using the latest and greatest is good for security.


peterbe:
I'm going to land this. We don't really *use* `dump_syms` yet anyway. This PR at least upgrades our Python docker image which is important.

PR #984, opened 2018-08-10
Reenable New Relic

sciurus:
This reverts commit b6c427fa4e94a26d796f93ca4ea545f37e02cadd.

While we are trying to understand symbols performance better with its new workload, it would be helpful to have data from New Relic again.

This is needed to re-enable New Relic in https://github.com/mozilla-services/cloudops-deployment/pull/2391


requirements.txt
```diff
@@ -48,6 +48,8 @@ pytz==2018.3 \
     --hash=sha256:5bd55c744e6feaa4d599a6cbd8228b4f8f9ba96de2c38d56f08e534b3c9edf0d \
     --hash=sha256:887ab5e5b32e4d0c86efddd3d055c1f363cbaa583beb8da5e22d2fa2f64d51ef \
     --hash=sha256:410bcd1d6409026fbaa65d9ed33bf6dd8b1e94a499e32168acfc7b332e4095c0
+newrelic==2.106.1.88 \
```
peterbe: That's now a very ancient version. Latest version is 4.2.0.100

But if you like, we can land this and let Pyup upgrade it. But that might be time consuming too.

bin/pre_deploy.sh
```diff
@@ -14,6 +14,6 @@ else
 fi
 
 echo "-----> PRE-DEPLOY: Running Django migration..."
-./manage.py migrate --noinput
+newrelic-admin run-program ./manage.py migrate --noinput
```
peterbe: I doubt this. We don't need newrelic to understand the db migrations.

requirements.txt
sciurus: Good call checking that. I'll bump it here.

bin/post_compile.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# The post_compile script is run in a sub-shell, so we need to source the
+# buildpack's utils script again, so we can use set-env/set-default-env:
+# https://github.com/heroku/heroku-buildpack-python/blob/master/bin/utils
+source $BIN_DIR/utils
+
+# Override the hostname that is displayed in New Relic, so the Dyno name
+# (eg "web.1") is shown, rather than "Dynamic Hostname". $DYNO is quoted
+# so that it's expanded at runtime on each dyno, rather than now.
+set-env NEW_RELIC_PROCESS_HOST_DISPLAY_NAME '$DYNO'
```
peterbe: I think this was got deleted as part of https://github.com/mozilla-services/tecken/commit/b6c427fa4e94a26d796f93ca4ea545f37e02cadd because it turned up when doing `grep`. I don't think this file gets used anywhere.

bin/run.sh
```diff
@@ -45,7 +45,7 @@ case $1 in
     exec python manage.py runserver 0.0.0.0:${PORT}
     ;;
   worker)
-    exec celery -A tecken.celery:app worker -l info
+    exec newrelic-admin run-program celery -A tecken.celery:app worker -l info
```
peterbe: Why can't this use that `$CMD_PREFIX` thing? Like we do for the line with `gunicorn`.

bin/pre_deploy.sh
sciurus: 👍

bin/post_compile.sh
sciurus: I don't see it anymore either, and clearly we're not running on Heroku.

bin/run.sh
sciurus: Makes sense.

sciurus:
This is ready for re-review.

PR #1040, opened 2018-08-29
enable python in Renovate

peterbe:
@rarkins As a piece of feedback on [the documentation](https://renovatebot.com/docs/python/) I think it's not clear why Python support is not enabled by default. This project as a `requirements.txt` and the config was basically just extending `config:base`. The documentation talks about how to *disable* but not how to enable it.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -7,5 +7,8 @@
   },
   "dockerfile": {
     "pinDigests": true
+  },
+  "python": {
+    "enabled": true
   }
 }


rarkins:
@peterbe python/pip should actually be enabled by default. I'm guessing that you're not receiving PRs because we don't support the hashes yet. Could you raise a feature request in the Renovate repo describing what logic you would expect Renovate to use to keep them updated?

peterbe:
[There. Done.](https://github.com/renovatebot/renovate/issues/2444)

peterbe:
Let's not merge this until we figure out why Renovate wasn't doing Python for us. The change in this PR shouldn't have to exist I suspect.

peterbe:
Going to stick with Pyup for the moment.

PR #1045, opened 2018-08-30
.pyup.yml syntax

### LINKED ISSUE #1044: .pyup.yml syntax
Hi! After a bit of digging, it turns out the issue with the PyUp parsing of .pyup.yml is because the attributes of each requirement file need to be double tabbed or space here: https://github.com/mozilla-services/tecken/blob/fc634ee014f7f75fd18acfc99af1d17ebb840f6d/.pyup.yml#L7

You have:
```
  ...
  - docs-requirements.txt:
    pin: False
```

And you need:
```
  ...
  - docs-requirements.txt:
      pin: False
```

For correct parsing by PyUp.

peterbe:
Fixes #1044

### DIFF
--- a/.pyup.yml
+++ b/.pyup.yml
@@ -4,4 +4,4 @@ requirements:
   - requirements.txt
   - tests/systemtest/requirements.txt
   - docs-requirements.txt:
-    pin: False
+      pin: False


peterbe:
@Jwomers What tools exist that I can run locally that checks that integrity or inspect/dry-run what my yaml file does?

PR #1421, opened 2018-12-18
healthcheck celery

peterbe:


peterbe:
@sciurus All that rabbit-hole digging... This is my final output towards the problem. 

We should, of course, discuss better how to spot this using the Datadog metrics. In particular, `uploads_created_update` is a metric [here](https://app.datadoghq.com/dash/339351/tecken-performance?screenId=339351&screenName=tecken-performance&from_ts=1542585600000&is_auto=false&live=true&page=0&tile_size=l&to_ts=1545177600000&fullscreen_widget=354222387) that, as you can see "laid dormant" for days. If you know a good way to set up an alert on that that would be awesome. 

Also, can you think of a good way to trigger a check of `curl -XPOST https://symbols.mozilla.org/__task_tester__ && curl https://symbols.mozilla.org/__task_tester__` post-deployment that would help too.

sciurus:
I set up an alert at https://app.datadoghq.com/monitors/7636848 . We can tackle the check during deploy as part of moving to GCP.

PR #1481, opened 2019-01-11
Incorrect module returned after a frame with module ID -1

### LINKED ISSUE #1475: Incorrect module returned after a frame with module ID -1
When a stack frame has a module of `-1`, following stack frames will have an incorrect module returned.

Example input (contrived as a demonstration):
```
{
  "memoryMap": [
    ["mozglue.pdb", "E160ACAB977F93F52AF4015ADD8B844E1"],
    ["kernelbase.pdb", "D396875654E9416CBA16E51F8B0A8B1E2"],
    ["ws2_32.pdb", "5D9C92DA00D24235AD321A8810C80B022"]
  ],
  "stacks": [
    [
      [0, 21180], [1, 131446], [-1, 0]
    ],
    [
      [0, 21180], [1, 131446], [2, 23643], [-1, 0], [1, 131446]
    ]
  ]
}
```

This request has 2 stacks that should look like:

- Stack 0
    - `mozglue!patched_LdrLoadDll`
    - `kernelbase!NlsConvertStringToIntegerW`
    - `(unknown)0x0`
- Stack 1
    - `mozglue!patched_LdrLoadDll`
    - `kernelbase!NlsConvertStringToIntegerW`
    - `ws2_32!NSPROVIDER::Initialize(...)`
    - `(unknown)0x0`
    - `kernelbase!NlsConvertStringToIntegerW`

But the actual stack returned is:

- Stack 0
    - `mozglue!patched_LdrLoadDll`
    - `kernelbase!NlsConvertStringToIntegerW`
    - `kernelbase!0x0`
- Stack 1
    - `kernelbase!patched_LdrLoadDll`
    - `kernelbase!NlsConvertStringToIntegerW`
    - `ws2_32!NSPROVIDER::Initialize(...)`
    - `ws2_32!0x0`
    - `ws2_32!NlsConvertStringToIntegerW`

Notice how unknown modules make the previous module "gets stuck" resulting in 2 frames with incorrect modules.

The `module=-1` case is needed because some stack frames don't land in any known module. In this case, the offset is considered to be an absolute address, not an offset.

Relevant code that generates these values: 

- [CombinedStacks.cpp | CreateJSStackObject(...)](https://searchfox.org/mozilla-central/rev/b4ebbe90ae4d0468fe6232bb4ce90699738c8125/toolkit/components/telemetry/other/CombinedStacks.cpp#193)
- [ProcessedStack.cpp | GetStackAndModules(...)](https://searchfox.org/mozilla-central/rev/b4ebbe90ae4d0468fe6232bb4ce90699738c8125/toolkit/components/telemetry/other/ProcessedStack.cpp#120)

peterbe:
Fixes #1475

@CarlCorcoran r? (Not expecting you to manually end-to-end test this like I did. What I want your review for is the wording. In particular it's a rather odd thing to use a string `(unknown)`. Why can't that be a more structured thing like `null` instead? 

Here's how I test it locally (beyond unit tests)

First I created a .zip like this:

```bash
./bin/make-a-zip.py mozglue.pdb/E160ACAB977F93F52AF4015ADD8B844E1 kernelbase.pdb/D396875654E9416CBA16E51F8B0A8B1E2 ws2_32.pdb/5D9C92DA00D24235AD321A8810C80B022
```
And then I uploaded that on http://localhost:3000/uploads/upload

```
▶ curl -XPOST -d '{"jobs":[{"memoryMap":[["mozglue.pdb","E160ACAB977F93F52AF4015ADD8B844E1"],["kernelbase.pdb","D396875654E9416CBA16E51F8B0A8B1E2"],["ws2_32.pdb","5D9C92DA00D24235AD321A8810C80B022"]],"stacks":[[[0,21180],[1,131446], [-1, 0]]]}]}' http://localhost:8000/symbolicate/v5 | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   817  100   590  100   227  29826  11475 --:--:-- --:--:-- --:--:-- 31052
{
  "results": [
    {
      "stacks": [
        [
          {
            "module_offset": "0x52bc",
            "module": "mozglue.pdb",
            "frame": 0,
            "function": "static long patched_LdrLoadDll(wchar_t *, unsigned long *, struct _UNICODE_STRING *, void * *)",
            "function_offset": "0x1ec"
          },
          {
            "module_offset": "0x20176",
            "module": "kernelbase.pdb",
            "frame": 1,
            "function": "NlsConvertStringToIntegerW",
            "function_offset": "0x2eb6"
          },
          {
            "module_offset": "0x0",
            "module": "(unknown)",
            "frame": 2
          }
        ]
      ],
      "found_modules": {
        "mozglue.pdb/E160ACAB977F93F52AF4015ADD8B844E1": true,
        "kernelbase.pdb/D396875654E9416CBA16E51F8B0A8B1E2": true,
        "ws2_32.pdb/5D9C92DA00D24235AD321A8810C80B022": null
      }
    }
  ]
}
```

### DIFF
--- a/tecken/symbolicate/views.py
+++ b/tecken/symbolicate/views.py
@@ -36,6 +36,9 @@ downloader = SymbolDownloader(
     file_prefix=settings.SYMBOL_FILE_PREFIX,
 )
 
+# What to put for the module if a stack points to a negative module index.
+UNKNOWN_MODULE = "(unknown)"
+
 # This lists all the possible exceptions that the SymbolDownloader
 # might raise that we swallow in runtime.
 # Any failure to download a symbol from S3, that is considered operational,
@@ -297,7 +300,7 @@ class SymbolicateJSON:
                     response_stack.append(
                         {
                             "module_offset": module_offset,
-                            "module": symbol_filename,
+                            "module": UNKNOWN_MODULE,
                             "frame": j,
                         }
                     )
--- a/tests/test_symbolicate.py
+++ b/tests/test_symbolicate.py
@@ -85,6 +85,25 @@ PUBLIC 3338 0 _ZL10do_preload
 PUBLIC 3340 0 _ZL14xpcomFunctions
 
     """,
+    "ws2_32.sym": """
+MODULE windows x86 9A8C8930C5E935E3B441CC9D6E72BB990 ws2_32.sym
+
+FUNC eb0 699 0 main
+FUNC 1550 cf 0 Output
+1550 6f 60 1
+15bf 2e 62 1
+15ed 12 65 1
+15ff 12 65 1
+1611 e 97 1
+FUNC 1620 d6 0 int SprintfLiteral<1024ul>(char (&) [1024ul], char const*, ...)
+1620 6f 32 3
+168f 36 34 3
+16c5 10 24 3
+16d5 11 25 3
+16e6 10 37 3
+FUNC 1700 149 0 mozilla::BinaryPath::GetFile(char const*, nsIFile**)
+PUBLIC 0 0 _mh_execute_header
+    """,
 }
 
 
@@ -105,6 +124,7 @@ def default_mock_api_call(self, operation_name, api_params):
         filename = api_params["Key"].split("/")[-1]
         if filename in SAMPLE_SYMBOL_CONTENT:
             return {"Body": BytesIO(SAMPLE_SYMBOL_CONTENT[filename].encode("utf-8"))}
+        print((filename, SAMPLE_SYMBOL_CONTENT.keys()))
         raise NotImplementedError(api_params)
 
     raise NotImplementedError(operation_name)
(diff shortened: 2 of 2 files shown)


CarlCorcoran:
Ooh excellent, thanks a ton for the fast response!

`(unknown)` is theoretically a valid module name, so I'd suggest using `<unknown>` instead, which is not a valid module name on any platform I can think of, therefore clients can detect that it's a special message generated by Tecken. In theory Tecken could use `<`/`>` to pass any kind of message in this field without compromising the rest of the symbolication.

In some ways I prefer `null` over `"<unknown>"`, but using the string is a simpler, less impact solution for any existing Tecken clients that expect that field to be non-null. This way the field is still always populated, and clients can be aware of this magic value if they want to be.

peterbe:
Ok. I've changed it from `(unknown)` to `<unknown>`. I'm going to land this if CI passes. The only thing I lack is that I wish there could be a better code-comment explanation for all of this. 

In particular, the fact that whatever software makes the stacks sometimes spits out -1 for the module index and what that means and why it's better to allow that to be OK than to pre-flight validate it. And the reasoning, mentioned above, about why `<unknown>` is better than `null` or `(unknown)`. Basically, a summary of all the things you've tried to explain to me. 

If you have 5min to write this down I can take care of putting it into the code.

CarlCorcoran:
Here are 2 comments:

Line 39:
```
# The way Firefox serializes its CombinedStacks object, a negative module index
# indicates that the stack frame's program counter was not found in any known
# module. In this case, no symbolication occurs, but we support this case as a
# convenience so clients don't need to filter this case out of stack frames and
# re-add them after symbolication.
#
# We don't return a null module name in order to further simplify the logic for
# clients. By always returning a valid module name string, clients aren't forced
# to always check for this condition.
#
# Angle brackets '<' and '>' cannot appear in real module names, so clients can
# test for this string to detect this condition 100% reliably.
UNKNOWN_MODULE = "<unknown>"
```

Line 298:
```
                if module_index < 0:
                    # Exit early. See declaration of UNKNOWN_MODULE for details
                    # about this case.
```

PR #1809, opened 2019-05-20
Finish unfinished paragraph in benchmarking docs

### LINKED ISSUE #1695: the benchmarking documentation is truncated
See [docs/benchmarking.rst](https://github.com/mozilla-services/tecken/blob/master/docs/benchmarking.rst), which just ends with "But basically the idea is that every benchmark is started by querying a".

willkg:
Pretty sure this finishes that section sufficiently.

Fixes #1695.


willkg:
self-r+

PR #1816, opened 2019-05-29
bug 1552901: Use oidcprovider in local dev

jwhitlock:
Use testprovider from [mozilla-parsys/docker-test-mozilla-django-oidc](https://github.com/mozilla-parsys/docker-test-mozilla-django-oidc) for authentication in the local development environment.

You may need to comment out ``DJANGO_OIDC_RP_CLIENT_ID`` and other parameters in your ``.env``, to get the new defaults.  Restart with ``make build run``, go to ``http://localhost:3000``, and authenticate. The setup is very similar to Socorro.

The docs are updated to reflect the new default, and that extra steps are required beyond the client ID and secret to get Auth0 working now.

### DIFF
--- a/tecken/settings.py
+++ b/tecken/settings.py
@@ -665,6 +665,26 @@ class Localdev(Base):
     API_UPLOADS_BATCH_SIZE = 10
     API_FILES_BATCH_SIZE = 20
 
+    #
+    # Default to the test oidcprovider container for Open ID Connect
+    #
+    # Client ID and secret must match oidcprovider database
+    OIDC_RP_CLIENT_ID = values.IntegerValue(1)
+    OIDC_RP_CLIENT_SECRET = values.Value("bd01adf93cfb")
+    # Load oidcprovider on public port 8081, without /etc/hosts changes
+    OIDC_OP_AUTHORIZATION_ENDPOINT = values.URLValue(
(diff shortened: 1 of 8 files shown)


tecken/settings.py
```diff
@@ -665,6 +665,26 @@ def VERSION(self):
...
+    )
+    # The backend connects to oidcprovider on docker port 8080
+    # Django's URL validator, used in URLValue, doesn't like docker hostnames
+    OIDC_OP_TOKEN_ENDPOINT = values.Value("http://oidcprovider:8080/openid/token")
+    # Same as token switch from URLValue to Value
+    OIDC_OP_USER_ENDPOINT = values.Value("http://oidcprovider:8080/openid/userinfo")
+    # Allow non-SSL connection to oidcprovider
+    OIDC_VERIFY_SSL = values.BooleanValue(False)
+    # Disable NotBlockedInAuth0Middleware
+    ENABLE_AUTH0_BLOCKED_CHECK = values.BooleanValue(False)
```
willkg: This makes me a bit nervous. If this wasn't in settings at all, then what is happening in dev, stage, and prod?

willkg: Oh, my bad--this is the `Localdev` set. Nevermind.

PR #1821, opened 2019-06-04
bug 1556775: add "make setup"

willkg:
This adds "make setup". It only sets up the db at the moment. We should add S3 and GCS, but I think we can do that in future PRs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,16 +10,17 @@ default:
 	@echo "Welcome to the tecken\n"
 	@echo "The list of commands for local development:\n"
 	@echo "  build            Builds the docker images for the docker-compose setup"
+	@echo "  setup            Initializes services"
+	@echo "  run              Runs the whole stack, served on http://localhost:8000/"
+	@echo "  stop             Stops the docker containers"
+	@echo ""
 	@echo "  clean            Stops and removes all docker containers"
-	@echo "  migrate          Runs the Django database migrations"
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -44,9 +45,9 @@ clean: .env stop
 	rm -rf coverage/ .coverage
 	rm -fr .docker-build
 
-.PHONY: migrate
-migrate: .env
-	docker-compose run web python manage.py migrate --run-syncdb
+.PHONY: setup
+setup: .env
+	docker-compose run web /app/bin/setup-services.sh
```
willkg: `make setup` will wipe and recreate the db, so I don't think we need `migrate` anymore.

pytest.ini
```diff
@@ -1,7 +1,7 @@
 [pytest]
 norecursedirs = .git .* frontend docs htmlcov .cache
 testpaths = tests
-addopts = -rsxX --showlocals --tb=native --no-migrations
+addopts = -rsxX --showlocals --tb=native --no-migrations -p no:cacheprovider
```
willkg: This fixes the cache warning that pytest kicks up because it can't save state.

willkg:
I think this is missing the thing that the socorro one is missing--it's not starting the services first. I think if you do a `make build` and then a `make setup` without starting the services, it'll fail. I'll double check that and fix it.

willkg:
Nevermind--it's fine.

I'm going to land this as is for now and we can improve it going forward.

PR #1825, opened 2019-06-05
Overhaul make-tag

willkg:
This fixes a bunch of things in make-tag to make it easier to use:

1. it figures out the correct remote since we don't all have "origin"
   as the remote name
2. it tells you more about what it's going to do before it does it
3. it prompts you before it does things--this gives you a chance to
   say, "OH NO!!! DON'T DO THAT!!!" before triggering a deploy


bin/make-tag.py
```diff
@@ -5,17 +5,30 @@
...
+def get_remote_name():
+    # Figure out remote to push tag to
+    remote_output = _check_output("git remote -v".split())
+
+    for line in remote_output.splitlines():
+        line = line.split("\t")
+        if "mozilla-services" in line[1]:
+            return line[0]
+
+    raise Exception("Can't figure out remote name for mozilla-services.")
```
willkg: This susses out the remote name so we can all have different remote names and we're fine.

willkg:
This doesn't affects dev, stage, or prod. I'm going to land it and use it now. self-r+

PR #1828, opened 2019-06-06
Upgrade js-yaml

willkg:


frontend/package.json
```diff
@@ -23,7 +23,8 @@
   "resolutions": {
     "braces": ">=2.3.1",
     "merge": ">=1.2.1",
-    "handlebars": ">=4.0.13"
+    "handlebars": ">=4.0.13",
+    "js-yaml": ">3.13.1"
```
peterbe:
```suggestion
    "js-yaml": ">=3.13.1"
```

peterbe: Actually, don't accept that verbatim. You'll need to re-run `yarn` after this manual edit to `package.json` so the yarn.lock updates too.

willkg:
Yay! self-r+

PR #1830, opened 2019-06-07
bug 1552950: Add GCS Emulator

jwhitlock:
**This is work in progress**, to preview my approach so far. Feel free to ignore or give feedback here or on [bug 1552950](https://bugzilla.mozilla.org/show_bug.cgi?id=1552950). This PR will probably not be the one submitted for review.

Add ``gcs-emulator``, an emulator for Google Cloud Storage, analogous to ``minio`` for AWS S3.

## Working
* Create the bucket during startup
* Upload symbols from a .zip, create file records

## To Fix
* [x] Uploads are not committed, so not stored on disk
* [x] ~~Uploads have null Download URL and Bucket Region~~ - not a bug
* [x] Downloads don't work, since uploads not committed
* [x] Download by direct URL

## Not yet tested
* [x] Upload via URL

## Possible Refactors
* Make interfaces on ``StorageBucket`` to handle differences between S3 and GCS, and real and emulated services, rather than handling differences in the client code. For example, currently to check if the bucket defined in ``DJANGO_UPLOAD_DEFAULT_URL`` exists:

```
from botocore.exceptions import ClientError
from google.cloud.exceptions import NotFound
from tecken.storage import StorageBucket
from django.conf import settings

bucket = StorageBucket(settings.DJANGO_UPLOAD_DEFAULT_URL)
exists = False
if bucket.is_google_cloud_storage:
    try:
        bucket.client.get_bucket(bucket.name)
    except NotFound:
        pass
    else:
        exists = True
else:
    try:
        bucket.client.head_bucket(Bucket=bucket.name)
    except ClientError as exception:
        if exception.response["Error"]["Code"] != "404":
            raise
    else:
        exists = True
```

This could be written as:

```
from tecken.storage import StorageBucket
from django.conf import settings

exists = StorageBucket(settings.DJANGO_UPLOAD_DEFAULT_URL).exists()
```

## To test

Add to ``.env``:
```
DJANGO_SYMBOL_URLS=https://gcs-emulator:4443/tecken
DJANGO_UPLOAD_DEFAULT_URL=https://gcs-emulator:4443/tecken
```

## Next Steps
* [x] Test current behavior of ``minio`` in local dev
* [x] Test current behavior of GCS in local dev
* [x] Survey use of ``StorageBucket``, investigate the refactor to interfaces.
* [x] Debug non-committed uploads

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -45,6 +45,7 @@ services:
       - minio
       - oidcprovider
       - statsd
+      - "gcs-emulator:gcs-emulator.127.0.0.1.nip.io"
     volumes:
       - $PWD:/app
     command: web-dev
@@ -112,6 +113,7 @@ services:
     links:
       - db
       - redis-cache
+      - "gcs-emulator:gcs-emulator.127.0.0.1.nip.io"
     volumes:
       - $PWD:/app
     command: worker-purge
@@ -153,3 +155,15 @@ services:
     image: local/tecken_oidcprovider
     ports:
       - "8081:8080"
+
+  # Fake Google Cloud Storage
+  gcs-emulator:
+    # TODO: Switch to official image, when it is ready (bug 1563036)
+    image: jwhitlock/gc-fake-storage:test
+    environment:
+      - GCS_FAKE_EXTERNAL_URL=https://gcs-emulator.127.0.0.1.nip.io:4443
+      - GCS_FAKE_PUBLIC_HOST=storage.gcs-emulator.127.0.0.1.nip.io:4443
+    ports:
+      - "4443:4443"
+    volumes:
+      - /storage
--- a/docs/dev.rst
+++ b/docs/dev.rst
@@ -163,6 +163,46 @@ public by default with:
 
     $ gsutil defacl set public-read gs://my-gcs-bucket
 
+Emulated Google Cloud Storage
+=============================
+
+The ``gcs-emulator`` container provides an emulated Google Cloud Storage service
+as well as a public host for downloads. It supports the operations needed by
+Tecken, such as creating and listing buckets, uploading and reading files, and
+serving files via the public host. Further GCS features may not be supported,
+but can be added to `fsouza/fake-gcs-server`_.
+
+To use the emulator, set the URLs in your ``.env`` file:
+
+.. code-block:: shell
+
+    DJANGO_SYMBOL_URLS=https://gcs-emulator.127.0.0.1.nip.io:4443/tecken
+    DJANGO_UPLOAD_DEFAULT_URL=https://gcs-emulator.127.0.0.1.nip.io:4443/tecken
+
+The bucket (``tecken`` in this example) will be created when the web server
+starts.
+
+Tecken changes the Google Cloud Storage client libraries to use the
+``gcs-emulator`` URLs rather than the official URLs. It is not possible to mix
+real and emulated GCS usage in the same Tecken instance, so ensure that all
+the URL settings (``DJANGO_SYMBOL_URLS``, ``DJANGO_UPLOAD_DEFAULT_URL``,
+``DJANGO_UPLOAD_TRY_SYMBOLS_URL``, and ``DJANGO_UPLOAD_URL_EXCEPTIONS``)
+are consistantly set to real or emulated GCS URLs.
+
+The public host at https://storage.gcs-emulator.127.0.0.1.nip.io:4443 uses a
+self-signed certificate.  You will need to add an SSL exception when downloading
+a symbol file for the first time.
+
+Files are persisted in a volume created by ``docker-compose``. To clear old
+uploads, delete the ``gcs-emulator`` container and volumes:
+
+.. code-block:: shell
+
+    $ docker-compose rm --force --stop -v gcs-emulator
+
+A fresh volume with no files will be created on the next startup of ``gcs-emulator``.
+
+.. _`fsouza/fake-gcs-server`: https://github.com/fsouza/fake-gcs-server
 
 Documentation
 =============
--- a/tecken/apps.py
+++ b/tecken/apps.py
@@ -116,16 +116,16 @@ class TeckenAppConfig(AppConfig):
             if not url:
                 continue
             netloc = urlparse(url).netloc
-            is_minio = "minio" in netloc
-            is_emulated_gcs = "gcs-emulator" in netloc
-            if is_minio:
+            is_emulated_s3 = StorageBucket.URL_FINGERPRINT["emulated-s3"] in netloc
+            is_emulated_gcs = StorageBucket.URL_FINGERPRINT["emulated-gcs"] in netloc
+            if is_emulated_s3:
                 bucket = StorageBucket(url)
                 try:
                     bucket.client.head_bucket(Bucket=bucket.name)
                 except ClientError as exception:
                     if exception.response["Error"]["Code"] == "404":
                         bucket.client.create_bucket(Bucket=bucket.name)
-                        logger.info(f"Created minio bucket {bucket.name!r}")
+                        logger.info(f"Created emulated S3 bucket {bucket.name!r}")
                     else:
                         # The most common problem is that the S3 doesn't match
                         # the AWS credentials configured.
--- a/tecken/apps.py
+++ b/tecken/apps.py
@@ -8,6 +8,7 @@ from urllib.parse import urlparse
 from botocore.exceptions import ClientError
 import markus
 from django_redis import get_redis_connection
+from google.cloud.exceptions import NotFound
 
 from django.conf import settings
 from django.apps import AppConfig
@@ -96,10 +97,15 @@ class TeckenAppConfig(AppConfig):
 
     @staticmethod
     def _check_storage_urls():
-        """If you use minio to functionally test S3, since it's
+        """Ensure emulated storage buckets exist.
+
+        If you use minio to functionally test S3, since it's
         ephemeral the buckets you create disappear after a restart.
         Make sure they exist. That's what we expect to happen with the
         real production S3 buckets.
+
+        The emulated GCS may also have missing buckets that can be created
+        during initialization.
         """
         _all_possible_urls = set(
             list(settings.SYMBOL_URLS)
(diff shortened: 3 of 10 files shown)


jwhitlock:
Upload by zip file works for both ``minio`` and real GCS credentials. In both cases, the upload file has ``null`` for both Download URL and Bucket Region, so that isn't a bug with ``gcs-emulator``.

For both backends, the existence check works, so that is a bug with ``gcs-emulator``. However, the existence check fails for ``minio`` when the ``UPLOAD_DEFAULT_URL`` is changed, so this may be an issue when the backend is changed in production.

The ``minio`` download link worked and returned the file.  The GCS link gave an ``AccessDenied`` error, because ``Anonymous caller does not have storage.objects.get access``. If the bucket was configured with this permission, it should work.

jwhitlock:
These are our storage operations:

- General Information:
  - Does a bucket exist?
  - Does an object exist?
  - What is the public URL of an object?
  - What is the size and other metadata of an object?
- Reading
  - What are the lines of the uncompressed content of an object?
- Adding data
  - Create a bucket
  - Upload an object
- Granting Access
  - Generate a "presigned" URL for a private object

The details for how to do these operations varies based on:
- Is it S3 or GCS?
- Is it a public or private S3 server?
- Is it an emulated service for local dev, or an external service?
- Do we want to fail fast with a short timeout?

The handlers that take care of the variations are spread over the project. One reason for this, rather than being implemented as methods on ``StorageBucket``, might be that some of the checks are cached with a [memoize decorator](https://github.com/peterbe/django-cache-memoize) that works best with plain functions.

My gut says to refactor this code before adding the GCS emulator, but it isn't a blocker, and there is still the upload bug to fix.

jwhitlock:
There are differences between the upload strategy of the Go and Python client (and maybe other clients), and the emulated GCS server only supported the Go strategy. I've adjusted it to support both (see https://github.com/fsouza/fake-gcs-server/pull/39 for details), and now large files are correctly uploaded.

The next bit to fix is the download URLs, which currently go to https://storage.googleapis.com. This may mean adjusting the client or using similar fixups to the ``minio`` URLs.

jwhitlock:
With the public host set, downloading works as intended. I added ``make clear-caches``, since the cache got in the way a few times. I added a persistent, unnamed volume so that the storage survives rebuilding the ``gcs-emulator`` container.

I tested upload via a download URL, and this worked as well. It appears the ``gcs-emulator`` behaves in a similar way to the real Google Cloud Storage.

willkg:
This is awesome work! Let's talk about this at the All Hands especially possible Tecken refactorings.

jwhitlock:
Expect a few more commits to come today, for tests and docs.

I think this could be split into two PRs, one for ``make clear-caches`` and one for the emulated GCS, both squashed to a single commit. Neither of these changes are urgent, and would benefit from the All Hands discussions.

jwhitlock:
Rebased to squash into two commits, and remove the ``WIP`` tag. This change turned out smaller than I first expected.

docker-compose.yml
```diff
@@ -153,3 +155,14 @@ services:
     image: local/tecken_oidcprovider
     ports:
       - "8081:8080"
+
+  # Fake Google Cloud Storage
+  gcs-emulator:
+    image: jwhitlock/gc-fake-storage:test
```
willkg: This is in your namespace and has a "test" tag. I think we either want to make this more durable now by moving the stuff to a better place or write up a bug to do it soon as a follow-up PR.

The socorro_app is here:

https://hub.docker.com/r/mozilla/socorro_app/

Maybe it makes sense to put it in the same place?

I know with the oidc_testprovider, the fact that it doesn't have a versioned tag creates problems when one tries to update. I think we'd want a versioned tag.

jwhitlock: I'll have to think about moving the image. My plan is that the upstream project starts officially supporting a dockerized server. I hope that this branched image only lasts another 30 days. That makes me reluctant to start down the path of getting push access to Mozilla's account or delegating creation to ops.

willkg: That makes sense. Once this lands, we should write up a bug to make sure this gets figured out this summer with some kind of durable solution.

docs/dev.rst
```diff
@@ -163,6 +163,46 @@ public by default with:
...
+The public host at https://storage.gcs-emulator.127.0.0.1.nip.io:4443 uses a
+self-signed certificate.  You will need to add an SSL exception when downloading
+a symbol file for the first time.
+
+Files are persisted in a volume created by ``docker-compose``. To clear old
+uploads, delete the ``gcs-emulator`` container and volumes:
+
+.. code-block:: shell
+
+    $ docker-compose rm --force --stop -v gcs-emulator
```
willkg: Is this something we should instead put in `bin/setup-services.sh` which gets run when you do `make setup`? The thinking being that there's related data in postgres and maybe redis and it should all get wiped and rebuilt at the same time.

tecken/apps.py
```diff
@@ -96,32 +97,47 @@ def _redis_store_eviction_policy():
...
-        """If you use minio to functionally test S3, since it's
+        """Ensure emulated storage buckets exist.
+
+        If you use minio to functionally test S3, since it's
         ephemeral the buckets you create disappear after a restart.
         Make sure they exist. That's what we expect to happen with the
         real production S3 buckets.
+
+        The emulated GCS may also have missing buckets that can be created
+        during initialization.
```
willkg: We have similar issues in Socorro. I'm of the mind that data storage things should exist before starting Tecken--Tecken should not be required to create them. This reduces the permissions Tecken needs to run.

I think setting this up should be the job of `make setup`.

Does that create new problems and/or doesn't really work in this situation?

willkg: Also, we could keep this and change it in a future cleanup PR. I don't mean to suggest we have to change it in this one.

tecken/apps.py
```diff
@@ -96,32 +97,47 @@ def _redis_store_eviction_policy():
...
-                    logger.info(f"Created minio bucket {bucket.name!r}")
-                else:
-                    # The most comment problem is that the S3 doesn't match
-                    # the AWS credentials configured.
-                    # If that's the case, this will raise a 403 Forbidden
-                    # ClientError.
-                    raise
+            netloc = urlparse(url).netloc
+            is_minio = "minio" in netloc
+            is_emulated_gcs = "gcs-emulator" in netloc
```
willkg: "minio" and "gcs-emulator" are magic strings here. I think I would lift them as "constants" with a comment along the lines of "must match docker-compose service names" or something like that.

tecken/dockerflow_extra.py
```diff
@@ -55,7 +55,7 @@ def check_url(url, setting_key):
         if url in checked:
             return
         bucket = StorageBucket(url)
-        if not bucket.private:
+        if bucket.is_emulated_gcs or not bucket.private:
```
willkg: Why this check here? Do we do the same for minio?

Also, since private bucket has nothing to do with emulated environment, I'd split this up and add a comment.

```
if bucket.is_emulated_gcs:
    # something something exception with emulated environments
    return
if bucket.private:
    return
```

willkg:
Possible future work:

* put the gcs-emulator docker image in a more durable place--we may not need to do this but if it doesn't get figured out upstream this summer, we should do this
* refactor the storage classes so there's a "storage API" and a bunch of classes that implement it
* see if we can reduce checks for "is this the emulated gcs environment" across the codebase; it'd be great if we had it in only one place when setting up a client

docs/dev.rst
jwhitlock: If ``make setup`` means "reset the world" as well as "bootstrap my environment", then yes. It doesn't make sense for a script in the web container to call docker-compose though, I think this would be at the Makefile level.

tecken/apps.py
jwhitlock: I think it may be a step too far for this PR. I think either local dev buckets would need to be hard-coded, or bash would need to call Python utilities to parse URLs, and optionally verify and create buckets. Maybe a management command is a compromise? I'll keep it in mind for the storage refactor.

docs/dev.rst
willkg: `make setup` isn't really about bootstrapping. It's about creating a default initial state for things including data services. That's an important step in bootstrapping a local dev environment, but it's also really handy in throwing away all the state of said environment.

We can't call docker-compose from a script inside a container--that's bad. But doing it in the `Makefile` as you suggest would be ok.

tecken/dockerflow_extra.py
jwhitlock: I want to verify that this is needed, it is possible that something broke (tests? bootstrapping?) before this change, but that a later fix resolved it.

While I'm doing that, do you know why do we skip verifying the existence of private buckets?

tecken/apps.py
willkg: We do this in Socorro for the db, es, and pubsub. The way we did it in Socorro is nice because those give developers additional tools for scripting things. It's been really helpful.

For s3, we have an aws cli wrapper. I'd love to rewrite how the s3 one works--the aws cli wrapper is pretty annoying.

tecken/dockerflow_extra.py
willkg: I can't think of any reason why we'd want to verify urls for private buckets, but not public ones. That's puzzling.

jwhitlock: oops, it is the other way around. This verifies public but not private buckets. I'm guessing that doesn't make it any clearer why.

jwhitlock: The code is able to verify buckets in the GCS emulator. I must have added this before the automatic bucket creation code.

I'll remove the change, which will leave ``dockerflow_extra.py`` unchanged for this PR, and the skipped private bucket checks as a mystery.

tecken/apps.py
jwhitlock: I think I understand better, having looked at the Socorro implementation, which does a lot more with ``make setup``. I think the way forward would be to move ``_check_storage_urls`` to a management command, and call it from the setup script. That allows us to reuse code in scripts and at runtime, and gets away from "at startup" initialization. Since that changes existing code around minio buckets, I'd like to focus on that in a future cleanup PR.

jwhitlock: [bug 1556775](https://bugzilla.mozilla.org/show_bug.cgi?id=1556775) is open, so I added [a comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1556775#c7) about the new additions to ``make setup``

willkg: :+1:

docker-compose.yml
jwhitlock: Filed as [bug 1563036](https://bugzilla.mozilla.org/show_bug.cgi?id=1563036).

jwhitlock:
This process was longer and messier than I hoped, but I think it is time to merge and move on to the next long, messy bit. A lot of discussion was threaded with the review, so to summarize:

* This adds a gcs-emulator service that provides a subset of the Google Cloud Storage API. It allows creating and listing buckets, and creating, listing, reading, and deleting objects. It does not include authentication or encryption, and several other features.
* This modifies the official GCS library to talk to the gcs-emulator instead of the official API endpoints. It does this in a way that prevents accessing the emulated and official API at the same time. I expect it will be a while, if ever, before the [upsteam libraries](https://github.com/googleapis/google-cloud-python) make this change easier.
* The gcs-emulator image is currently [jwhitlock/gc-fake-storage:test](https://cloud.docker.com/repository/docker/jwhitlock/gc-fake-storage/general), which is built from the [test branch of my fork](https://github.com/jwhitlock/gc-fake-storage/tree/test) of https://github.com/teone/gc-fake-storage. The long term plan is for the upstream project to merge the Docker build configuration. See https://github.com/fsouza/fake-gcs-server/issues/45 and [bug 1563036](https://bugzilla.mozilla.org/show_bug.cgi?id=1563036).
* This PR follows the pattern of auto-creating emulated S3 buckets at server startup to create the emulated GCS buckets as well. The long term plan is to move both to ``make setup``. See [bug 1556775](https://bugzilla.mozilla.org/show_bug.cgi?id=1556775).
* The code that handles the difference in storage backends is spread over the code base. It may be better to create a generic object storage interface on ``StorageBucket``, and hide the details of the backends from the client code. See [bug 1564452](https://bugzilla.mozilla.org/show_bug.cgi?id=1564452).

PR #1835, opened 2019-06-28
Fix shields in README

willkg:
This fixes the shields in the README so they're less clunky.


willkg:
Yay! Looks much better now.

willkg:
self-r+

PR #1839, opened 2019-07-23
bug 1564452: Add StorageBucket.exists(), .backend()

jwhitlock:
This PR, part of [bug 1564452](https://bugzilla.mozilla.org/show_bug.cgi?id=1564452), adds some methods to ``StorageBucket`` so that client code can use object storage without being as concerned whether it is S3 or GCS, emulated or "real".

* ``exists()``: Returns ``True`` if the bucket exists, ``False`` if it does not, and raises a backend-specific exception for issues like authentication and authorization issues.
* ``backend()``: Set to a string like ``"s3"`` or ``"emulated-gcs"``, to classify the backend if needed.

This expands the code and tests in ``StorageBucket``, and simplifies the code and tests for the code that checks for bucket existence.

This first PR took more time than I planned, as I got my S3 and GCS accounts setup, and made a bunch of test requests to investigate the API, and became more familiar with the code. According the lines added and deleted, it looks like the complexity mostly moved, rather than things being simplified. I hope as more API methods are implemented, we can see larger gains in simplicity for client code and tests.

### DIFF
--- a/tecken/dockerflow_extra.py
+++ b/tecken/dockerflow_extra.py
@@ -2,9 +2,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, you can obtain one at http://mozilla.org/MPL/2.0/.
 
-from google.api_core.exceptions import GoogleAPIError
-from botocore.exceptions import ClientError, EndpointConnectionError
-
 from django.core import checks
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
@@ -14,7 +11,7 @@ from dockerflow.health import (
     ERROR_MISCONFIGURED_REDIS,
     ERROR_REDIS_PING_FAILED,
 )
-from tecken.storage import StorageBucket
+from tecken.storage import StorageBucket, StorageError
 
 
 def check_redis_store_connected(app_configs, **kwargs):
@@ -63,24 +60,15 @@ def check_storage_urls(app_configs, **kwargs):
                     checks.Error(
                         f"Unable to connect to {url} (bucket={bucket.name!r}), "
                         f"because bucket not found",
-                        id="tecken.health.E003",
+                        id="tecken.health.E001",
                     )
                 )
-        except (GoogleAPIError, ClientError) as exception:
+        except StorageError as error:
             errors.append(
                 checks.Error(
                     f"Unable to connect to {url} (bucket={bucket.name!r}), "
-                    f"due to {type(exception).__name__}: {exception}",
+                    f"due to {error.backend_msg}",
                     id="tecken.health.E002",
-                )
-            )
-        except EndpointConnectionError:
-            errors.append(
-                checks.Error(
-                    f"Unable to connect to {url} (bucket={bucket.name!r}, "
-                    f"found in settings.{setting_key}) due to "
-                    f"EndpointConnectionError",
-                    id="tecken.health.E001",
                 )
             )
         else:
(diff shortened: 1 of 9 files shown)


tecken/dockerflow_extra.py
```diff
@@ -57,44 +57,34 @@ def check_url(url, setting_key):
...
+                    id="tecken.health.E002",
+                )
+            )
+        except EndpointConnectionError:
+            errors.append(
+                checks.Error(
+                    f"Unable to connect to {url} (bucket={bucket.name!r}, "
+                    f"found in settings.{setting_key}) due to "
+                    f"EndpointConnectionError",
+                    id="tecken.health.E001",
```
willkg: Can't `.exists()` throw an exception that wraps the actual exception? Then we don't need the S3 and GCP specific handling here.

tecken/download/views.py
```diff
@@ -151,7 +151,7 @@ def download_symbol(request, symbol, debugid, filename, try_symbols=False):
             # from the host.
             if (
                 settings.DEBUG
-                and StorageBucket.URL_FINGERPRINT["emulated-s3"] in url
+                and StorageBucket(url).backend == "emulated-s3"
                 and "http://minio:9000" in url
                 and request.get_host() == "localhost:8000"
             ):  # pragma: no cover
```
willkg: Is the issue that this code is solving that inside the web container, the minio container has one name and from the docker host, it has another. Is that right?

tests/test_storage.py
```diff
@@ -2,46 +2,189 @@
...
+def test_init(url, expected):
+    """The URL is processed during initialization."""
+    bucket = StorageBucket(url)
+    assert bucket.backend == expected["backend"]
+    assert bucket.base_url == expected["base_url"]
+    assert bucket.endpoint_url == expected["endpoint_url"]
+    assert bucket.name == expected["name"]
+    assert bucket.prefix == expected["prefix"]
+    assert bucket.private == expected["private"]
+    assert bucket.region == expected["region"]
```
willkg: Nicely done!

tecken/dockerflow_extra.py
jwhitlock: Yes! Python 3's exception chaining makes this nice for uncaught exceptions. For the dockerflow_extra checks, I'll need to get some details into the error message, but I think I can copy the client's exception string, rather than wrap the exception.

tecken/download/views.py
jwhitlock: That's the way I read it too. Minio may have a configuration that will make this go away, or we can do the nip.io trick, or maybe we hold our nose until the S3 code goes away.

willkg: I'm +1 on whatever you want to do. Given our current status, I wouldn't hold this PR up for "the best solution to all things".

jwhitlock: I'm adding it to my list, but not changing for this PR.

jwhitlock:
These new commits add ``StorageError``, raised when a backend-specific client raises an exception. In general, the code that uses ``StorageBucket`` shouldn't handle these errors, but instead let them propagate. However, the dockerflow code turns them into error messages, so there's enough human-level info in the exception to aid in debugging. The dockerflow error handlers is simpler, and one error case has gone away, so I've reassigned the error IDs.

The storage code was 8 lines away from 100% code coverage, so I added a few more tests as well. This will make it easier to ensure my new code is properly covered.

PR #1841, opened 2019-07-30
Update Python requirements and constraints

willkg:


willkg:
I went through all the make rules and this seemed fine. I'm going to land this and we'll know more when it hits stage. self-r+

PR #1842, opened 2019-07-30
Upgrade to Python 3.7

willkg:
This pins to Python 3.7.4 using slim and stretch. We need to keep using
stretch because it has the gcc-6 package which we need to compile
`dump_syms`. In order to get off of stretch, we'd need to fix all that.
It's probably doable, but I didn't want to tackle it today.


willkg:
I went through the make rules and this looked fine. I want to see it on stage for a bit. self-r+

PR #1849, opened 2019-08-08
Switch from make-tag to release.py

willkg:
This switches to `release.py` so we have the same deploy steps across all
the projects we maintain.


willkg:
We can move the pytest configuration into `setup.cfg`, too, but then that orphans `pytest-watch` configuration. I've never used `pytest-watch` before, so I don't know what it's for. I figured I'd just add another configuration file and someone with opinions could sort it all out later.

willkg:
This is a minor change even though it looks more substantial. It's also really hard to test without deploying. `release.py` is based on `make-tag` and the changes I made to `make-tag` in this repository a while back.

Makefile
```diff
@@ -101,10 +101,6 @@ django-shell: .env .docker-build
 docs:
 	@bin/build-docs-locally.sh
 
-.PHONY: tags
```
jwhitlock: I like this style of declaring ``.PHONY`` along with the commands, rather than collecting all the ``.PHONY`` declarations on one line. However, it didn't prevent someone from misspelling the make target ``tag`` as ``tags``. Deleting is one way to fix it 😁

willkg: Oops--that's probably my fault.

willkg:
Thank you!

PR #1855, opened 2019-08-13
Fix uploads date-fmt format usage

willkg:
In date-fmt, ' is an "escape character", but "escaped" things actually
need to be quoted. This adds the end '.


willkg:
self-r+

PR #1873, opened 2019-10-23
bug 1590767: update oidc-testprovider to v0.9.3

willkg:
This only affects the local dev environment. Now that we have this, I can do this from the shell:

```
docker-compose exec oidcprovider /code/manage.py createuser USER PASSWORD EMAIL
```


willkg:
This only affects the local dev environment and if there are issues, we can fix it with a PR without affecting anything else. self-r+

PR #1874, opened 2019-10-24
bug 1585319: Remove Google Cloud Storage (GCS), default to S3

jwhitlock:
This PR removes GCS, which is mostly reversing two PRs:

* https://github.com/mozilla-services/tecken/pull/1226 - Added GCS, made it the new testing default
* https://github.com/mozilla-services/tecken/pull/1830 - Added GCS emulator

These PRs helped identify the code, documentation, test, and requirements to change. The changes were done manually, because a year of updates and bug fixes prevents a clean ``git revert``.

This could potentially be a smaller change, especially in the test code, if the ``StorageBucket`` API was expanded first ([bug 1564452](https://bugzilla.mozilla.org/show_bug.cgi?id=1564452)). However, the purpose of that change was to lessen the application and test code that was concerned with S3 / GCS differences, so it less urgent now.

My test process:

```shell
make build   # Recreate with new requirements
make test
make setup
make shell   # Add superuser with upload powers
make run     # Confirm upload of symbols to http://minio:9000
```

The stage and production environments were not switched to GCS, and will continue to use S3.


tecken/base/symboldownloader.py
```diff
@@ -200,15 +193,9 @@ def invalidate_cache(self, symbol, debugid, filename):
...
-                    exists_in_source.invalidate(
-                        bucket, self._make_key(prefix, symbol, debugid, filename)
-                    )
-                else:
-                    exists_in_source.invalidate(
-                        source, self._make_key(prefix, symbol, debugid, filename)
-                    )
+                exists_in_source.invalidate(
+                    source, self._make_key(prefix, symbol, debugid, filename)
+                )
```
jwhitlock: This is typical of the application changes. A ``if`` statement is used for the GCS code path, and a ``else`` for the S3 path. Removing GCS gets rid of the ``if`` statement and de-dents the code.

tests/test_download.py
```diff
@@ -51,49 +52,7 @@ def reload_downloaders(urls, try_downloader=None):
...
-    response = client.head(url)
-    assert response.status_code == 200
-    assert response.content == b""
-
-    assert response["Access-Control-Allow-Origin"] == "*"
-    assert response["Access-Control-Allow-Methods"] == "GET"
-
-
-def test_client_happy_path_s3(client, botomock, metricsmock, settings):
+def test_client_happy_path(client, botomock):
```
jwhitlock: In many cases, a test using GCS is followed by a test using S3. It is enough to delete the GCS test and rename the S3 test to drop ``_s3`` from the function name.

tests/test_download.py
```diff
@@ -125,38 +84,29 @@ def mock_api_call(self, operation_name, api_params):
...
-
-    def mock_get_blob(key):
-        if key == "prefix/v0/xul.pdb/44E4EC8C2F41492B9369D6B9A059577C2/xul.sym":
-            # As if it exists happily
-            blob = gcsmock.mock_blob_factory(key)
-            blob.public_url = "https://googleapis.example.com/private/" + key
-            return blob
-
-        raise NotImplementedError(key)
+def test_client_legacy_product_prefix(client, botomock):
```
jwhitlock: In other cases like this, a GCS-only test was later added, so the test needs to be converted to use ``botomock`` and a ``mock_api_call`` inner function.

I also noticed that ``metricsmock`` was added as an unused fixture. I've removed these as I found them.

tests/test_upload.py
```diff
@@ -101,177 +101,11 @@ def test_get_key_content_type(settings):
     assert get_key_content_type("foo.HTML") == "text/html"
 
 
-@pytest.mark.django_db
-def test_upload_archive_happy_path(
```
jwhitlock: The S3 variant of ``test_upload_archive_happy_path`` appears later in the file. To try to keep the diff sane for review, I removed this GCS variant test and left the S3 variant where it was.

tests/test_upload.py
```diff
@@ -409,17 +230,14 @@ def mock_upload_from_file(file):
 
 
 @pytest.mark.django_db
-def test_upload_archive_happy_path_s3(
+def test_upload_archive_happy_path(
```
jwhitlock: Here's the S3 variant of ``test_upload_archive_happy_path``.

tests/test_upload.py
```diff
@@ -573,117 +391,12 @@ def mock_api_call(self, operation_name, api_params):
...
     client,
     botomock,
     fakeuser,
-    metricsmock,
     upload_mock_invalidate_symbolicate_cache,
     upload_mock_update_uploads_created_task,
-    settings,
 ):
-    settings.UPLOAD_DEFAULT_URL = "https://s3.example.com/private/prefix/"
-    settings.UPLOAD_TRY_SYMBOLS_URL = "https://s3.example.com/try/prefix"
```
jwhitlock: When GCS was the default for tests, it was necessary to set these settings for S3 test variants. With S3 as the default again, it is not needed, and in some cases, the settings fixture is not needed either.

tests/test_upload.py
```diff
@@ -1006,199 +652,18 @@ def mock_api_call(self, operation_name, api_params):
...
-        assert response.status_code == 201
-
-        assert Upload.objects.all().count() == 3
-        assert FileUpload.objects.all().count() == 1
-        # This time it doesn't need to look up the size a third time
-        assert len(lookups) == 2
-
-
-@pytest.mark.django_db
-def test_upload_archive_key_lookup_cached_s3(
```
jwhitlock: The differ loses its place here, making it a little harder to review. What's going on:

* ``test_upload_archive_key_lookup_cached`` removed (GCS)
* ``test_upload_archive_key_lookup_cached_s3`` renamed to ``test_upload_archive_key_lookup_cached``
* ``test_upload_archive_key_lookup_cached_without_metadata`` removed (GCS)
* ``test_upload_archive_key_lookup_cached_without_metadata_s3`` renamed to ``test_upload_archive_key_lookup_cached_without_metadata``

The ``_without_metadata`` test is so similar that the diff gets confused, but it syncs again after these two tests.

tests/test_upload.py
```diff
@@ -1274,20 +741,16 @@ def mocked_create_blob(key):
 
 
 @pytest.mark.django_db
-def test_upload_archive_key_lookup_cached_without_metadata_s3(
+def test_upload_archive_key_lookup_cached_without_metadata(
```
jwhitlock: The differ is back in sync here.

tecken/base/symboldownloader.py
willkg: :+1:

tests/test_download.py
willkg: :+1:

jwhitlock:
Thanks!

PR #1894, opened 2019-11-04
bug 1592143: add new domains to ALLOW_UPLOAD_BY_DOWNLOAD_DOMAINS

willkg:
Taskcluster is moving to a new infrastructure. This adds two more
domains to the `ALLOW_UPLOAD_BY_DOWNLOAD_DOMAINS` setting allowing symbols
upload tasks to upload by download from the new infrastructure.

### DIFF
--- a/tecken/settings.py
+++ b/tecken/settings.py
@@ -559,7 +559,13 @@ class Base(Core):
     # When you "upload by download", the URL's domain needs to be in this
     # whitelist. This is to double-check that we don't allow downloads from
     # domains we don't fully trust.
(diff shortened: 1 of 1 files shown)


willkg:
To verify this, I'd make sure I didn't do a typo. Then we can test it in stage/prod.

willkg:
I don't know why this is set as the default and not something in the infrastructure. I can't think of a compelling reason for either option. Might as well keep it in code until one comes up.

willkg:
Thank you!

PR #1895, opened 2019-11-14
Add "pip check" to build step

willkg:


Dockerfile
```diff
@@ -54,7 +54,9 @@ COPY requirements.txt /tmp/
 COPY requirements-constraints.txt /tmp/
 # Switch to /tmp to install dependencies outside home dir
 WORKDIR /tmp
-RUN pip install --no-cache-dir -r requirements.txt
+RUN pip install -U 'pip>=10' && \
```
jwhitlock: Was updating pip needed? The one included in the Python base image was sufficient for other projects. And, if you are updating pip, then ``--disable-pip-version-check`` seems unneeded (except to avoid a network request).

willkg: I changed it to match all the other projects we work on.

willkg:
Thank you!

PR #1905, opened 2019-12-11
Update to Django 2.2 LTS

jwhitlock:
* Django 2.1.14→[2.2.8](https://docs.djangoproject.com/en/2.2/releases/2.2/): Latest Long Term Support (LTS) version.
* sqlparse 0.3.0: New requirement in Django 2.2

The project builds, tests, and runs locally. With warnings on (``PYTHONWARNINGS=always pytest tests --capture=no``), there are two warnings about deprecated settings, but this is a known issue (https://github.com/jazzband/django-configurations/issues/233) and requires no change when updating to 3.0.


jwhitlock:
Thanks!

PR #1908, opened 2019-12-30
Update handlerbars to 4.5.3

willkg:


willkg:
I went through the ui in my local dev environment and everything seemed to work as far as I can tell. self-r+

PR #1912, opened 2020-01-02
Update to Python 3.7.6, update python dependencies, move some things around

willkg:
This does a bunch of things:

* move requirements files into `requirements/` like our other projects
* moves a bunch of dependencies from `requirements/default.txt` to `requirements/constraints.txt` which are really dependencies of dependencies
* move `Dockerfile` to `docker/`
* update to Python 3.7.6
* update python dependencies in `default.txt`


willkg:
self-r+

I'm going to merge this now and it can sit on dev for the weekend and we can verify it on Monday.

PR #1964, opened 2020-02-28
minor cleanup of scaffolding

willkg:
This is some minor cleanup of project scaffolding: make, docker-compose.yml, and lint rules. This makes Tecken a little more like Socorro and friends.


willkg:
This gets Tecken working in CI again and also fixes some nits where it's different than Socorro. self-r+

PR #1990, opened 2020-03-27
Update to CircleCI config 2.1

willkg:
As a side-effect, this switches CircleCI pipelines on.


willkg:
self-r+

willkg:
As we upgrade the CircleCI configuration number, so, too, we upgrade our souls.

PR #2016, opened 2020-04-13
bug 1618942: add upload system tests

willkg:
This tweaks some systemtest bits and adds the scripts to test uploading symbols ZIP files and uploading symbols ZIP files by download url.

To test:

```
1. `make shell`
2. `cd systemtests`
3. `./setup_tests.sh`
4. `./test_env.sh stage`
5. `./test_env.sh prod`
```


systemtests/bin/fetch-crashids.py
```diff
@@ -0,0 +1,100 @@
...
+import click
+import requests
+
+
+CRASHSTATS = "https://crash-stats.mozilla.org/"
+MAX_PAGE = 1000
+
+# Indicators that the crash report probably doesn't have a good stack for
+# symbolication
+MEH_INDICATORS = [
```
willkg, line 22: Variable named on a meh Friday.

systemtests/bin/fetch-crashids.py
```diff
@@ -0,0 +1,100 @@
...
+        else:
+            print(result["uuid"])
+
+        crashids += 1
+        if crashids > num_results:
+            break
+
+
+if __name__ == "__main__":
+    fetch_crashids()
```
willkg, line 100: We have a couple of implementations of `fetch_crashids` in Socorro, but I needed something in Tecken and I wanted it to be self-contained and I didn't want to install more dependencies. That's why I wrote this.

Some future Tecken developer with different opinions of the world may decide differently in which case looking at the Socorro stuff is probably a good idea.

systemtests/bin/make-symbols-zip.py
```diff
@@ -0,0 +1,193 @@
...
+                sym_filename = sym_filename[3:]
+
+            url = urljoin(SYMBOLS_URL, sym_filename)
+            if is_try:
+                url = url + "?try"
+            sym_file = os.path.join(sym_dir, sym_filename)
+            download_sym_file(url, sym_file)
+
+            # Build the new zip file
+            build_zip_file(tmp_zip_path, sym_dir)
```
willkg, line 186: This has a rough runtime thing where it creates a ZIP file every time it adds a SYM file. SYM files are text, so they compress pretty well. We could probably improve this by guessing the compression ratio and adding a bunch of SYM files until we're 75% of the way there, then adding-and-checking.

I didn't do that because it adds complexity and I don't see myself running this script very often. Given that, I'd rather it be simpler than more performant.

systemtests/bin/upload-symbols.py
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env python
```
willkg, line 1: I decided to do this and `upload_symbols_by_download.py` as separate scripts even though they test the same API endpoint. It was either that or have subcommands and two modes. I started with that and then decided I wasn't saving much by combining the two.

systemtests/test_env.sh
```diff
@@ -46,6 +52,24 @@ esac
 echo "HOST: ${HOST}"
 echo ""
 
+# DESTRUCTIVE TESTS
+if [ "${DESTRUCTIVE_TESTS}" == "1" ]; then
+    # Test upload
+    for FN in ./data/zip-files/*.zip
```
willkg, line 58: Need an echo line here.

willkg:
This is a script that devs will run on their local machines. It's important that the test is:

1. easy to set up
2. well documented
3. easy to maintain/extend
4. easy to run

I think this gets 95% of the way there. I think it meets my immediate needs for ongoing Tecken maintenance and development.

self-r+

PR #2017, opened 2020-04-14
Update kind-of to 6.0.3 or later

willkg:


willkg:
I ran this locally and then flipped through the Tecken web interface. We have no JS testing and no smoketest plan, so it's ad hoc. Seems fine as far as I can tell.

self-r+

PR #2019, opened 2020-04-14
Fix prettier run in circleci

willkg:
prettier is getting run in a different way in circle than it is in the local dev environment. That's causing problems when I upgrade prettier.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,9 @@ jobs:
       - run:
           name: Run lint check
           command: |
+            # NOTE(willkg): Use -ci services since they doesn't mount . as a volume
             docker-compose run linting-ci lint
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -40,7 +40,9 @@ jobs:
       - run:
           name: Run lint check
           command: |
+            # NOTE(willkg): Use -ci services since they doesn't mount . as a volume
             docker-compose run linting-ci lint
+            docker-compose run frontend-ci lint
```
willkg, line 45: We should do all the linting in this step with containers built via our Docker instructions.

willkg:
The frontend_lint item was removed in this PR, but since it was removed in this PR, I think something is confused and thinks it's still an outstanding check.

self-r+

PR #2026, opened 2020-04-14
bug 1628802: fix permissions issues in docker containers

willkg:
This fixes permission issues when running the docker containers on Linux where the uid/gid of the host user doesn't match the app user.

To test:

1. add `USE_UID` and `USE_GID` to `.env`
2. `make build`
3. `make setup`
4. `make lintfix`
5. `make test`
6. `make run` (though this is broken because of a different issue that I'll get to next)

None of those should result in files that are owned by root or some other user.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,7 @@ frontend/build/:
 
 .PHONY: build-frontend
 build-frontend:
-	docker-compose run -e CI web ./bin/build_frontend.sh
+	docker-compose run --no-deps -e CI web ./bin/build_frontend.sh
 
 .PHONY: lint
 lint: .env .docker-build
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,17 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+# Include my.env and export it so variables set in there are available
+# in the Makefile.
+include .env
+export
+
+# Set these in the environment to override them. This is helpful for
+# development if you have file ownership problems because the user
+# in the container doesn't match the user on your host.
+USE_UID ?= 10001
+USE_GID ?= 10001
+
 .PHONY: help
 help: default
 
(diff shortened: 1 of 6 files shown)


willkg:
This works locally (Linux). I'll test this out on MacOS next and fix anything that comes up.

Makefile
```diff
@@ -19,11 +30,8 @@ default:
 	@echo "  redis-store-cli  Opens a Redis CLI to the store Redis server"
 	@echo "  clear-caches     Clear Redis caches"
 	@echo "  shell            Opens a Bash shell"
-	@echo "  currentshell     Opens a Bash shell into existing running 'web' container"
 	@echo "  test             Runs the Python test suite"
 	@echo "  testshell        Runs a shell in the test environment"
-	@echo "  gunicorn         Runs the whole stack using gunicorn on http://localhost:8000/"
-	@echo "  django-shell     Django integrative shell"
```
willkg, line 26: I took out some make rules that can easily be run manually and don't come up often.

Makefile
```diff
@@ -54,17 +62,11 @@ clear-caches:
 
 .PHONY: setup
 setup: .env
-	docker-compose run web /app/bin/setup-services.sh
+	docker-compose run --rm web /app/bin/setup-services.sh
```
willkg, line 65: This was leaking docker containers. This should reduce the number of containers created and orphaned.

Makefile
```diff
@@ -54,17 +62,11 @@ clear-caches:
...
 .PHONY: shell
 shell: .env .docker-build
-	# Use `-u 0` to automatically become root in the shell
-	docker-compose run --user 0 web bash
-
-.PHONY: currentshell
-currentshell: .env .docker-build
-	# Use `-u 0` to automatically become root in the shell
-	docker-compose exec --user 0 web bash
+	docker-compose run --rm web bash
```
willkg, line 69: This was leaking docker containers. This should reduce the number of containers created and orphaned.

Makefile
```diff
@@ -84,32 +86,27 @@ stop: .env
...
-django-shell: .env .docker-build
-	docker-compose run web python manage.py shell
-
 .PHONY: docs
 docs:
-	@bin/build-docs-locally.sh
+	bin/build-docs-locally.sh
+
+frontend/build/:
+	make build-frontend
```
willkg, line 105: This reduces a step in local development. Having said that, when this is run in CI, it causes problems because the check for this directory happens on the host and not in the container. I'm going to leave it here, but it needs some thinking.

willkg:
Everything I did to test this worked on MacOS.

self-r+

PR #2027, opened 2020-04-15
bug 1630120: fix frontend in local dev environment

willkg:
This fixes the frontend container to run with the UID/GID set in the `.env` file so it doesn't create files as root.

This adds a script to build the yarn.lock file and dealing with the `frontend/node_modules/` directory that has "stuff" in it and can cause problems if left around.

This demotes react-scripts back to 3.4.0 which works--3.4.1 seems to skip running the dev server with no errors.

This rebuilds the `yarn.lock` file.

To test:

1. `make build`
2. `make run`
3. go to http://localhost:3000 and click through the site


willkg:
self-r+

PR #2029, opened 2020-04-15
bug 1552947: remove credentials argument

willkg:
The credentials argument defaults to "same-origin" now, so we don't need
to specify it.


willkg:
self-r+

PR #2030, opened 2020-04-16
bug 1596109: fix ACAO header with static assets

willkg:
I'm pretty sure static assets are served by whitenoise which is adding this header:

```
Access-Control-Allow-Origin: *
```
Since we're not hosting static assets on a CDN, we don't need this so we should disable it.


willkg:
We'll test it in stage. self-r+

PR #2032, opened 2020-04-30
bug 1634372: fix make-stacks to handle unknown frames

willkg:
This fixes `make-stacks` to handle unknown frames and modules correctly.

To test:

1. `make shell`
2. `cd systemtests/`
3. `./bin/make-stacks.py save stackdir CRASHID` for a bunch of crash reports

### DIFF
--- a/systemtests/bin/make-stacks.py
+++ b/systemtests/bin/make-stacks.py
@@ -56,22 +56,30 @@ def build_stack(data):
         return {}
 
     modules = []
+    modules_list = []
     for module in json_dump.get("modules", []):
         debug_file = module.get("debug_file", "")
         debug_id = module.get("debug_id", "")
+        if not debug_file or not debug_id:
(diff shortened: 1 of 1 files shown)


systemtests/bin/make-stacks.py
```diff
@@ -56,22 +56,30 @@ def build_stack(data):
...
         debug_file = module.get("debug_file", "")
         debug_id = module.get("debug_id", "")
+        if not debug_file or not debug_id:
+            continue
+        # Add the module information to the map
         modules.append((debug_file, debug_id))
-
-    modules_list = [m["filename"] for m in json_dump.get("modules", [])]
+        # Keep track of which modules are at which index
+        modules_list.append(module["filename"])
```
willkg, line 68: We need to make sure these two lists are in lock step because we're potentially skipping modules.

PR #2057, opened 2020-05-28
bug 1641212: remove "download_from_microsoft" code

willkg:
A while back, we started implementing automatic download of missing Microsoft symbols. That feature was never finished. Since then, another group has taken on the task of uploading symbols and has a system for doing that, so we no longer need this feature.

This removes it.


willkg:
To test:

1. rebuild the docker images
2. run the system tests
3. go into the ui and click through all the tabs and choices

There are a couple of interstitial pages that aren't interesting anymore. I thought about spending time figuring out how to remove them, but then decided it wasn't worth it.

willkg:
One thing this does is drop a field from the `upload_fileupload` table. I'm pretty sure that table is huge, so I wonder if that migration would take a long time to run and/or fail.

I need to talk to Brian before landing this.

willkg:
I talked to Brian. We decided to defer the table update, so I'm going to remove that from this PR before landing.

docker/Dockerfile
```diff
@@ -38,14 +37,7 @@ RUN groupadd --gid $groupid app && \
...
     apt-transport-https build-essential curl git libpq-dev \
-    gettext libffi-dev jed
-
-# Install dump_syms
-RUN DEBIAN_FRONTEND=noninteractive \
-    apt-get install -y --no-install-recommends \
-    gyp ninja-build binutils-gold gcc-6 g++-6 pkg-config cabextract
-COPY ./docker/build_dump_syms.sh /tmp
-RUN /tmp/build_dump_syms.sh
+    gettext libffi-dev
```
willkg, line 40: I took jed out because I don't use it and we don't need it in the image.

docker/Dockerfile
```diff
@@ -13,9 +13,8 @@ COPY . /app
 WORKDIR /app
 RUN bin/build_frontend.sh
 
-# NOTE(willkg): stick with stretch because it has gcc-6 which we need
-# to build dump_syms.
-FROM python:3.7.7-slim-stretch@sha256:3b072d391496e120bd25025462489b8a437f17135dda96f765de5d0a0e159fa3
+
+FROM python:3.7.7-slim@sha256:1d96a34649e53ee768aff6d75918aa7ebb7e0b21b55a41795a95f27bb7268026
```
willkg, line 17: Since we're not building `dump_syms` anymore, we can use buster (debian stable).

docker/Dockerfile
```diff
@@ -38,14 +37,7 @@ RUN groupadd --gid $groupid app && \
...
     apt-get install -y --no-install-recommends \
     apt-transport-https build-essential curl git libpq-dev \
-    gettext libffi-dev jed
-
-# Install dump_syms
-RUN DEBIAN_FRONTEND=noninteractive \
-    apt-get install -y --no-install-recommends \
-    gyp ninja-build binutils-gold gcc-6 g++-6 pkg-config cabextract
-COPY ./docker/build_dump_syms.sh /tmp
-RUN /tmp/build_dump_syms.sh
```
willkg, line 48: This builds `dump_syms`. I removed this step from the Dockerfile as well as the `build_dump_syms.sh` script.

tecken/download/views.py
```diff
@@ -179,50 +152,12 @@ def download_symbol(request, symbol, debugid, filename, try_symbols=False):
...
-                code_id=code_id,
-                missing_symbol_hash=missing_symbol_hash,
-            )
-
-            # The querying of Microsoft's server is potentially slow.
-            # That's why this call is down in a celery task.
-            # But there is hope! And the client ought to be informed
-            # that if they just try again in a couple of seconds/minutes
-            # it might just be there.
-            delayed_lookup = True
```
willkg, line 224: I can remove other uses of this variable.

tecken/download/views.py
```diff
@@ -325,11 +243,7 @@ def missing_symbols_csv(request):
 
     # By default, only do those updated in the last 24h
     qs = MissingSymbol.objects.filter(
-        modified_at__gte=date,
-        # This is a trick to immediately limit the symbols that could
-        # be gotten from a Microsoft download.
-        symbol__iendswith=".pdb",
-        filename__iendswith=".sym",
```
willkg, line 332: Now that I'm re-reading this, I think we need to re-add it. It's not related to the `download_from_microsoft` feature.

willkg:
self-r+

PR #2087, opened 2020-06-01
bug 1641212: remove intermediary downloads listing page

willkg:
There used to be two "downloads" pages and an index page that linked to both. When I removed the download_from_microsoft feature, there was only one download page and we didn't need an index page anymore. This removes the index page.


willkg:
self-r+

PR #2095, opened 2020-06-02
Add note about redis==3.4.1

willkg:
Redis deprecated HMSET. We're still using 3.4.2 in production. Redis 4 added variadic arguments to HSET which is what you're supposed to switch to. That's all super....

However, updating the Python redis library to 3.5.0 deprecates hmset so you have to switch to hset, but we can't use variadic arguments because the version of Redis we're using doesn't support it. Ergo, we're hosed.

We have a few of options:

1. Update Redis in stage/prod to something >= 4.
2. Rewrite the symbolicate API view code to not use hmset.
3. Pin the library to 3.4.1 and push this off until another day.

I'm choosing option 3 for now.


willkg:
self-r+

PR #2098, opened 2020-06-08
Rebuild yarn.lock

willkg:
This rebuilds yarn.lock to pick up updates on a variety of libraries including a security fix in websocket-extensions (0.1.3 -> 0.1.4).


willkg:
I tested this locally. I'll re-test on stage. self-r+

PR #2100, opened 2020-06-16
bug 1646133: switch to better language

willkg:
This does two things:

1. switch to better language where appropriate and helpful
2. ditch the blockade plugin which was a modified form of pytest-blockage that we weren't actually using


willkg:
self-r+

PR #2101, opened 2020-06-17
bug 1627052: rewrite symbolication API docs

willkg:
This overhauls the symbolication API docs to make them more user-focused. It should be easier to read the docs and understand how to use the API now.

Further, this fixes the docs so they use the sphinx-rtd-theme when building locally. This is what we use on RTD, so it makes it easier to edit and author them locally.


willkg:
I'm going to land this and then verify it on RTD and make any changes in a future PR.

self-r+

PR #2134, opened 2020-07-02
Tweak Makefile

willkg:
Replace explicit help docs with ones generated from Makefile making it
easier to maintain.


willkg:
self-r+

PR #2135, opened 2020-07-06
bug 1650917: update bugzilla links to Tecken product

willkg:
We created a Tecken product in Bugzilla and all Tecken bugs should get filed there. This updates the links accordingly.


willkg:
self-r+

PR #2136, opened 2020-07-07
bug 1649531: handle empty string debug filename and id

willkg:
This fixes symbolication to handle empty strings for the debug filename
and debug id in the memoryMap portion of the symbolication request.
Previously, it would raise an error. It no longer raises an error.
Instead, the code doesn't check cache and doesn't try to download the
module and treats it as if the module isn't available.

I also fixed make-stacks so it no longer drops modules where the debug
filename or id aren't known.

### DIFF
--- a/tecken/symbolicate/views.py
+++ b/tecken/symbolicate/views.py
@@ -165,7 +165,7 @@ class SymbolicateJSON:
         needed_modules_lookups = {
             key: modules_lookups[key]
             for key in modules_lookups
-            if key not in self.all_symbol_offsets
+            if key[0] and key[1] and key not in self.all_symbol_offsets
         }
 
         if needed_modules_lookups:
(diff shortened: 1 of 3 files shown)


tecken/symbolicate/views.py
```diff
@@ -165,7 +165,7 @@ def symbolicate(self, stacks, memory_map):
         needed_modules_lookups = {
             key: modules_lookups[key]
             for key in modules_lookups
-            if key not in self.all_symbol_offsets
+            if key[0] and key[1] and key not in self.all_symbol_offsets
```
willkg, line 168: `key[0]` and `key[1]` can be the empty string. If they are, we don't want to try to look them up. We want to leave them as `None`.

willkg:
I ran the tests and also fixed make-stacks and built a bunch of stacks that have `["", ""]` modules in them. I tested it with those stacks and everything looks ok.

self-r+

PR #2137, opened 2020-07-08
bug 1651190: maintain missingsymbol table

willkg:
This removes some code about total counts that we don't really need. This also adds a Django management command and test for removing old entries in the missingsymbol table.

To test:

1. run tests
2. create entries in missingsymbol and run `manage.py cleanse_missingsymbol`

### DIFF
--- a/tecken/api/views.py
+++ b/tecken/api/views.py
@@ -1177,15 +1177,7 @@ def downloads_missing(request):
     start = (page - 1) * batch_size
     end = start + batch_size
 
-    # The MissingSymbol class has a classmethod called `total_count`
-    # which returns basically the same as `MissingSymbol.objects.count()`
-    # but it comes from a counter in the cache instead.
-    if any(v for v in form.cleaned_data.values()):
-        # Use the queryset
-        total_count = qs.count()
-    else:
-        # No specific filtering was done, we can use the increment counter.
-        total_count = MissingSymbol.total_count()
+    total_count = qs.count()
 
     context["aggregates"] = {"missing": {"total": total_count}}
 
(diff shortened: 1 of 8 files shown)


tecken/api/views.py
```diff
@@ -1177,15 +1177,7 @@ def downloads_missing(request):
...
 
-    # The MissingSymbol class has a classmethod called `total_count`
-    # which returns basically the same as `MissingSymbol.objects.count()`
-    # but it comes from a counter in the cache instead.
-    if any(v for v in form.cleaned_data.values()):
-        # Use the queryset
-        total_count = qs.count()
-    else:
-        # No specific filtering was done, we can use the increment counter.
-        total_count = MissingSymbol.total_count()
```
willkg, line 1188: After we land the code for deleting records, we can probably remove the line that looks at the 30-day count, too. I claim we only need "total" and "1-day" in the webpage.

tests/test_download.py
```diff
@@ -761,37 +761,38 @@ def mock_api_call(self, operation_name, api_params):
 
 
 @pytest.mark.django_db
-def test_missingsymbol_model_counts():
-    # The .total_count won't change just because you manually
-    # create, edit, or delete MissingSymbol instances. The increments are
-    # done explicitly in the code where upserts are made.
-    assert MissingSymbol.total_count() == 0
+def test_cleanse_missingsymbol_delete_records():
```
willkg, line 764: This is an unfortunate diff. I removed the `test_missingsymbol_model_counts` function completely. I added a totally different and unrelated `test_cleanse_missingsymbol_delete_records` test.

willkg:
This is nice, but won't do anything until we add the ability to do scheduled tasks.

self-r+

PR #2138, opened 2020-07-10
bug 1638412: add cleanse_upload command

willkg:
This deletes `upload_upload` and `upload_fileupload` records for files that have expired in S3.

### DIFF
--- a/tecken/upload/management/commands/cleanse_upload.py
+++ b/tecken/upload/management/commands/cleanse_upload.py
@@ -36,15 +36,16 @@ class Command(BaseCommand):
         uploads = Upload.objects.filter(try_symbols=is_try, created_at__lte=cutoff)
         file_uploads = FileUpload.objects.filter(upload__in=uploads)
 
+        # Delete fileupload first because it's got a foreignkey to upload
         if is_dry_run:
-            upload_count = file_uploads.count()
+            fileupload_count = file_uploads.count()
         else:
-            upload_count = file_uploads.delete()[0]
+            fileupload_count = file_uploads.delete()[0]
 
         if is_dry_run:
-            fileupload_count = uploads.count()
+            upload_count = uploads.count()
         else:
-            fileupload_count = uploads.delete()[0]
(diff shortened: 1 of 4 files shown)


willkg:
@sciurus : These are the lifecycles we have set on the S3 bucket for public symbols, right?

* `v1/` -> 2 years
* `try/` -> 28 days

sciurus:
Yes, those are the days until deletion.

tecken/upload/management/commands/cleanse_upload.py
```diff
@@ -0,0 +1,50 @@
...
+        self.stdout.write(
+            f"cleanse_upload: fileupload try={is_try}: Deleted {ret[0]} records older than {cutoff}"
+        )
+        ret = uploads.delete()
+        self.stdout.write(
+            f"cleanse_upload: upload try={is_try}: Deleted {ret[0]} records older than {cutoff}"
+        )
+
+    def handle(self, *args, **options):
+        today = timezone.now()
```
willkg, line 42: I wrote this (and the other cleanse command for missing download records) figuring we'd just run it. However, I think we want to run it the first time in a "it's not permanent" kind of mode so we can verify that it's going to do what we think it's going to do.

I should add a `--dry-run` mode. I'll do that in this PR.

tecken/upload/management/commands/cleanse_upload.py
```diff
@@ -0,0 +1,75 @@
...
+        )
+
+    def delete_records(self, is_dry_run, is_try, cutoff):
+        uploads = Upload.objects.filter(try_symbols=is_try, created_at__lte=cutoff)
+        file_uploads = FileUpload.objects.filter(upload__in=uploads)
+
+        if is_dry_run:
+            upload_count = file_uploads.count()
+        else:
+            upload_count = file_uploads.delete()[0]
```
willkg, line 42: Oops--this is reversed. :(

willkg:
self-r+

PR #2144, opened 2020-07-30
bug 1656075: move contributing docs to CONTRIBUTING.rst

willkg:
This moves a bunch of documentation around contributing to `CONTRIBUTING.rst` in the root of the repository which GitHub will do special things with.

This also adds the MPLv2 preamble to all the JS files which should have it.


CONTRIBUTING.rst
```diff
@@ -0,0 +1,150 @@
...
+
+  $ make docs
+
+Then view ``docs/_build/html/index.html`` in your browser.
+
+
+Testing
+=======
+
+See the testing documentation at https://tecken.readthedocs.io/en/latest/dev.html .
```
willkg, line 150: Have to use a url here because this file gets viewed in two contexts: as part of the documentation on RTD, but also in GitHub where the renderer doesn't know anything about the other docs.

willkg:
self-r+

I may follow this up with another PR depending on how things build on RTD and show up in GitHub.

PR #2145, opened 2020-07-31
bug 1655944: fix bad token responses so the client gets them

willkg:
This fixes the issue where the client is trying to upload a big symbols file using a bad token. The Tecken middleware will see the auth header, recognize it's a bad token, and then return a 403 with some message. However, if the client is using chunked transfer encoding and the request body hasn't been fully sent, the connection gets dropped before the client gets the 403 response.

This fixes that by forcing the middleware to slurp the entire request body (if possible) before sending the 403 response.

This adds a system test that tests the situation in stage and prod environments.

To test:

1. run tests to make sure they still pass
2. run the system test against the stage environment to verify it fails in the "UPLOAD WITH BAD TOKEN TEST--this should return a 403 and error and not a RemoteDisconnected" test
3. run the system test against the local dev environment to make sure it doesn't run that test and that all the system tests pass

### DIFF
--- a/systemtests/bin/list-firefox-symbols-zips.py
+++ b/systemtests/bin/list-firefox-symbols-zips.py
@@ -15,13 +15,14 @@ import requests
 
 SYMBOLS_ZIP_SUFFIX = ".crashreporter-symbols.zip"
 SYMBOLS_FULL_ZIP_SUFFIX = ".crashreporter-symbols-full.zip"
-INDEX = "https://index.taskcluster.net/v1/"
-QUEUE = "https://queue.taskcluster.net/v1/"
+INDEX = "https://firefox-ci-tc.services.mozilla.com/api/index/v1/"
+QUEUE = "https://firefox-ci-tc.services.mozilla.com/api/queue/v1/"
 NAMESPACE = "gecko.v2.mozilla-central.revision.REV.firefox"
 
 
 def index_namespaces(namespace, limit=1000):
-    resp = requests.post(INDEX + "namespaces/" + namespace, json={"limit": limit})
+    url = INDEX + "namespaces/" + namespace
+    resp = requests.post(url, json={"limit": limit})
     for namespace in resp.json()["namespaces"]:
         yield namespace["namespace"]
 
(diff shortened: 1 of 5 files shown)


willkg:
@jwhitlock : Can you eyeball this and let me know if this looks right?

systemtests/bin/list-firefox-symbols-zips.py
```diff
@@ -15,13 +15,14 @@
 
 SYMBOLS_ZIP_SUFFIX = ".crashreporter-symbols.zip"
 SYMBOLS_FULL_ZIP_SUFFIX = ".crashreporter-symbols-full.zip"
-INDEX = "https://index.taskcluster.net/v1/"
-QUEUE = "https://queue.taskcluster.net/v1/"
+INDEX = "https://firefox-ci-tc.services.mozilla.com/api/index/v1/"
```
jwhitlock, line 18: Good catch!

systemtests/bin/upload-symbols.py
```diff
@@ -82,7 +83,10 @@ def upload_symbols(ctx, base_url, auth_token, symbolsfile):
                         headers={"auth-token": auth_token},
                         timeout=CONNECTION_TIMEOUT,
                     )
-                if resp.status_code != 201:
+                if resp.status_code == 429:
+                    click.echo(click.style("429--sleeping for 5", fg="yellow"))
+                    time.sleep(5)
+                elif resp.status_code != 201:
```
willkg, line 89: I should add explicit handling for 403 here and in `upload-symbols-by-download.py`.

Also, it'd be better if the script took an expected status code so it could "yay" or "nay" based on the expectation since we expect uploading with a bad token to fail.

willkg:
^^^ That redoes the response handling so we can pass the status code we're expecting. That's the minimal functionality we need for the system test results to be easier to eyeball.

willkg:
^^^ And that fixes linting and the errant exit. Bleh. It's ready for review again.

willkg:
Thank you!

PR #2167, opened 2020-08-03
bug 1655944: fix force_full_request_body_read function

willkg:
In a chunked transfer encoding scenario, the content length will be 0
so using that to guard against reading forever won't work. Underneath
the Django request .read() is gunicorn's read which will handle
dropped connections, timeouts, and other possible data transfer issues.

Given that, we can just loop until there's nothing left to read and
that'll be ok.


willkg:
self-r+

PR #2176, opened 2020-08-05
bug 1656313: update upload docs

willkg:
This improves the upload documentation to make it easier for people who
are using the upload API to implement their clients. This documents the
various status codes and what the client should do when it encounters
them.

This slims down some of the explanatory text that isn't particularly
helpful and clarifies some of the text around what decisions an upload
client needs to make.


willkg:
self-r+

I'll see how this renders on RTD and fix any issues I notice.

PR #2179, opened 2020-08-17
bug 1555567: docs pass 2

willkg:
This tweaks the index to better separate user-centric docs from dev/admin-centric docs. This also removes the docker tips and tricks chapter.


willkg:
self-r+

PR #2208, opened 2020-09-15
bug 1665127: add notes about user-agent

willkg:
We really want people to specify the User-Agent HTTP header so we can
better know who we need to talk to when changing API things.


willkg:
self-r+

PR #2212, opened 2020-09-24
Clean up docs-related things

willkg:
This removes a bunch of documentation and stuff related to running Tecken in a prod-like environment. I haven't had to do that in the year and change that I've maintained Tecken and I think it's out-of-date and I'd rather remove it than maintain it. If we need to d that in the future, we can figure it out again.

This also redoes how docs get built so that they're built using the Docker image. That makes things easier and more consistent.

This also updates the Sphinx Makefile to a modern one which doesn't have the tab issues (among other things).


Makefile
```diff
@@ -68,16 +68,16 @@ clear-caches:  ## | Clear Redis caches.
 
 .PHONY: redis-cache-cli
 redis-cache-cli: .env .docker-build  ## | Open Redis CLI to cache Redis server.
-	docker-compose run redis-cache redis-cli -h redis-cache
+	docker-compose run --rm redis-cache redis-cli -h redis-cache
```
willkg, line 71: The `--rm` reduces accumulation of docker container state which overwhelms my hard drive.

Makefile
```diff
@@ -88,22 +88,22 @@ testshell: .env .docker-build  ## | Open shell in test environment.
 	bin/test.sh --shell
 
 .PHONY: docs
-docs:  ## | Build docs.
-	bin/build-docs-locally.sh
+docs: .env .docker-build  ## | Build docs.
+	docker-compose run --rm --user ${USE_UID} linting docs
```
willkg, line 92: I used the `linting` container here. What we should probably do is rename it to `tools` or `utils` since it's now used for building docs and linting.

docs/Makefile
```diff
@@ -1,232 +1,20 @@
-# Makefile for Sphinx documentation
+# Minimal makefile for Sphinx documentation
```
willkg, line 1: Sphinx creates a Makefile when you initialize a Sphinx project. The Makefile we had is from running that a long time ago. So I updated it.

docs/Makefile
```diff
@@ -1,232 +1,20 @@
...
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    = -W
-SPHINXBUILD   = sphinx-build
-SPHINXAPIDOC  = sphinx-apidoc
-PAPER         =
+SPHINXOPTS    =
+SPHINXBUILD   = python -msphinx
+SPHINXPROJ    = Socorro
```
willkg, line 7: I cheated and updated it by copying the one from Socorro. I'll fix that.

docs/dev.rst
```diff
@@ -204,132 +204,30 @@ corresponding user in the oidcprovider service like this:
 
    docker-compose exec oidcprovider /code/manage.py createuser yourname yourpassword yourname@example.com
 
-Running ``gunicorn`` locally
```
willkg, line 207: I have no idea if any of this works because I've never used it. I didn't want to go through it and update it. I figure if I haven't used it since I started maintaining Tecken, it's worth removing.

willkg:
self-r+

PR #2213, opened 2020-09-25
bug 1667213: switch to django admin

willkg:
This enables the Django admin, nixes the User Management react app in favor of the Django user management bit, and adds a LogEntry viewer for auditing things.


frontend/src/App.js
```diff
@@ -84,6 +82,17 @@ const App = observer(
...
     };
 
+    adminLink = () => {
+      // Need to figure out the right url that works in the local dev environment
+      // which does this proxy thing
+      let here = window.location;
+      let url =
+        here.protocol + "//" + here.hostname + ":" + here.port + "/admin";
+      // Replace proxy webpack host with django host in local dev environment
+      url = url.replace("http://localhost:3000/", "http://localhost:8000/");
```
willkg, line 92: This kind of stuff is kind of gross, but there's no other way to deal with the proxy in the local dev environment.

tecken/base/admin.py
```diff
@@ -0,0 +1,75 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
willkg, line 1: I cribbed this from Socorro. It lets us see the audit log created by any changes that happen in the Django admin.

tecken/useradmin/admin.py
```diff
@@ -0,0 +1,50 @@
...
+
+# Unregister the original UserAdmin and register our better one
+try:
+    admin.site.unregister(User)
+except TypeError:
+    pass
+
+
+@admin.register(User)
+class UserAdminBetter(UserAdmin):
```
willkg, line 21: I cribbed this from Socorro. It shows users, but the bits that are helpful for us. We don't use the "username" field, so it's better to show users by email address. Further, it's nice to see number of uploads and api tokens and which groups they're in.

tecken/useradmin/management/commands/superuser.py
```diff
@@ -31,12 +31,16 @@ def handle(self, *args, **options):
                 username = default_username_algo(email)
             user = User.objects.create(username=username, email=email)
             user.set_unusable_password()
-            self.stdout.write(self.style.WARNING("User created"))
+            self.stdout.write(self.style.WARNING("New user created"))
 
-        user.is_superuser = not user.is_superuser
```
willkg, line 36: This used to enable/disable superuser status, but we don't need that. In a local dev environment, we need a command that creates a user account and assigns superuser/staff to it. That simplifies a ton of things when developing/testing.

PR #2214, opened 2020-09-25
bug 1667213: fix static assets in Django

willkg:
The static assets weren't loading for the Django admin. This reworks a
bunch of the bits related to building Tecken to simplify it between CI
and non-CI environments, remove things that aren't helpful/used, and add
Django staticfiles support.


Makefile
```diff
@@ -91,13 +91,6 @@ testshell: .env .docker-build  ## | Open shell in test environment.
 docs: .env .docker-build  ## | Build docs.
 	docker-compose run --rm --user ${USE_UID} linting docs
 
-frontend/build/:
-	make build-frontend
-
-.PHONY: build-frontend
-build-frontend:  ## | Build frontend static files.
```
willkg, line 98: Year-ago-me added this to patch over difficulties running the tests after building Tecken. However, it doesn't build the frontend static files at all--it's a complete lie. Year-ago-me wasted today-me's time!

The problem is that the tests for some reason need stuff in `/app/frontend/build/` to exist. That exists in the `tecken:build` Docker image, but when we use `docker-compose` to run anything, we volume mount the repository directory as `/app` and thus the `/app/frontend/build/` that's in the image disappears in a puff of magic smoke.

So year-ago-me thought, "Well that sucks. I'll just create an empty directory there."

Today-me went through and did a different way to fix this and put that "if the directory don't exist, create it" bit in `bin/test.sh` and removed this completely misleading "Build frontend static files" make rule of insincerity.

bin/build_frontend.sh
```diff
@@ -5,29 +5,18 @@ set -eo pipefail
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-# Builds the React frontend, *only* if we are in CI.
+# Builds the React frontend.
 
-# True if the environment variable 'CI' is NOT empty.
-# In CircleCI, this variable is "true" and in local dev, it's "".
-if [[ ! -z "${CI}" ]]; then
```
willkg, line 12: This whole CI/not-CI thing shaves 18s off of building the image for me. However, it creates this situation where the local development environment is wildly different than what we build in CI for stage/prod which makes it really hard to figure out what's going on when debugging static files and what's actually getting built.

I wasted a whole day today. I think it's worth losing 18s per build to have a local dev environment building images that are what we're building for stage/prod.

bin/run.sh
```diff
@@ -74,9 +74,7 @@ case $1 in
       bash
     else
       # python manage.py collectstatic --noinput
-      coverage erase
-      coverage run -m pytest "${@:2}"
-      coverage report -m
```
willkg, line 79: The copious coverage output is copious and not something I ever look at. I don't hate coverage, but I don't want to run it all the time.

tecken/settings.py
```diff
@@ -807,42 +807,45 @@ class Prod(Stage):
     """Configuration to be used in prod environment"""
 
 
-class Prodlike(Prod):
```
willkg, line 810: I nixed all the Prodlike stuff a couple of commits ago and missed this configuration.

tecken/settings.py
```diff
@@ -807,42 +807,45 @@ class Prod(Stage):
...
+    CACHES = {
+        "default": {
+            "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+            "LOCATION": "unique-snowflake",
+        },
+        "store": {
+            "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+            "LOCATION": "unique-snowflake",
+        },
+    }
```
willkg, line 832: I guessed on a lot of these configuration variables.

willkg:
self-r+ in anguish and misery

PR #2216, opened 2020-09-25
bug 1667213: add status page to django admin

willkg:
This adds a status page to the Django admin. In order to do that, I had to add the ability to have static pages in the Django admin.

### DIFF
--- a/tecken/base/admin.py
+++ b/tecken/base/admin.py
@@ -2,9 +2,20 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, you can obtain one at http://mozilla.org/MPL/2.0/.
 
+import json
+from urllib.parse import urlparse, urlunparse
+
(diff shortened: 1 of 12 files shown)


tecken/base/admin.py
```diff
@@ -73,3 +84,90 @@ def has_change_permission(self, request, obj=None):
...
+    )
+    try:
+        redis_cache_info = get_redis_connection("default").info()
+    except NotImplementedError:
+        redis_cache_info = {"redis_version": "fakeredis"}
+    context["versions"].append(
+        {"key": "Redis Cache", "value": redis_cache_info["redis_version"]}
+    )
+
+    context["versions"].sort(key=lambda x: x["key"])
```
willkg, line 171: The one version this misses that was in the previous version is the version of React. I don't think it's all that helpful to know the React version and there are other ways to figure it out.

willkg:
self-r+

PR #2217, opened 2020-09-26
bug 1667213: reduce race condition between tests and services

willkg:
The tests are kicking off before the backing services have started up.

What we should really be doing is using urlwait or something like that
which will cause everything to pause while the service is coming up, but
because of the way configuration works, we may not have the service urls
in the environment so we'll have to go with this "start it up a little
bit before" method for now.


willkg:
self-r+

PR #2218, opened 2020-09-26
Repository cleanup

willkg:
This moves a few things around to make room for another microservice.

Specifically:

1. remove the `pytest.ini` file and fold contents into `setup.cfg`
2. move `logo.png` into `docs/`
3. move `tests/` into `tecken/`


willkg:
self-r+

PR #2219, opened 2020-09-28
More cleanup

willkg:
This renames `run.sh` to `entrypoint.sh` and reworks it to focus on services.

This removes the shell implementation of "waitfor" replacing it with urlwait.

This redoes linting so it uses a `bin/run_lint.sh` script for python lint and lintfix and runs in regular or ci containers.

This redoes tests so that it uses a `bin/run_test.sh` script that sets things up and uses the test-c or test container depending on whether it's CI or not.

This adds better auto-generated usage for `bin/entrypoint.sh`.

To test, run through the make rules.


Makefile
```diff
@@ -81,22 +81,38 @@ psql: .env .docker-build  ## | Open psql cli.
...
+.PHONY: lintci
+lintci: .env .docker-build  ## | Lint code in test-ci container.
+	docker-compose run --rm --no-deps test-ci bash ./bin/run_lint.sh
+	docker-compose run --rm frontend-ci lint
+
 .PHONY: lintfix
 lintfix: .env .docker-build  ## | Reformat code.
-	docker-compose run --rm linting lintfix
+	docker-compose run --rm --no-deps test bash ./bin/run_lint.sh --fix
 	docker-compose run --rm frontend lintfix
```
willkg, line 118: We don't run this in CI, so we don't need a CI equivalent.

willkg:
self-r+

PR #2220, opened 2020-09-30
bug 1667132: add table counts to site status admin

willkg:
Table counts will make it easier to see growth of data in the db.


willkg:
self-r+

PR #2221, opened 2020-09-30
Fixes to systemtests

willkg:
This improves the error message for a user when the stack to be
symbolicated isn't valid JSON.

This fixes the comment in the `sym_files_to_download.csv` to be
correct--the response HTTP status code doesn't have anything to do with
tokens.


willkg:
self-r+

PR #2248, opened 2020-10-02
bug 1668518: switch to pip-compiled requirements.txt file

willkg:


docs/dev.rst
```diff
@@ -57,37 +57,20 @@ Managing dependencies
...
-For example, to add ``foobar`` version 5::
-
-  $ hashin -r requirements/default.txt foobar==5
-
-If ``foobar`` has any dependencies that would also be installed, you must add
-them to the constraints file::
-
-  $ hashin -r requirements/constraints.txt bazzbiff==4.0
+   $ make shell
+   $ pip-compile --generate-hashes requirements.in
```
willkg, line 66: @relud So one funny thing about this is that I need the Tecken Docker image to be built in order to run this inside the image. Otherwise I don't have the system dependencies installed to run `pip-compile` without it kicking up errors.

Is that how you add new dependencies in ingestion-edge? If not, how do you do it?

relud: I start a `pip install pip-tools` in a venv or the docker container i'm running

requirements.txt
```diff
@@ -1,278 +1,209 @@
-toml==0.10.1 \
-    --hash=sha256:bda89d5935c2eac546d648028b9901107a595863cb36bae0c73ac804a9b4ce88 \
-    --hash=sha256:926b612be1e5ce0634a2ca03470f95169cf16f939018233a670519cb4ac58b0f
+#
+# This file is autogenerated by pip-compile
+# To update, run:
+#
+#    pip-compile --generate-hashes requirements.in
```
relud, line 5: the `requirements.in` argument is unnecessary, as it's automatically detected. it's still preserved when passed, so removing it here means you don't have to type it when you update it.
```suggestion
#    pip-compile --generate-hashes
```

Some dependencies have python-version-specific requirements, and if a dev runs a globally installed `pip-compile`, it may not be on the same version of python as the docker container. So I recommend enforcing the correctness of this in CI, e.g.
https://github.com/mozilla/bigquery-etl/blob/3725a64/.circleci/config.yml#L37-L46

docs/dev.rst
relud, line 66: to be consistent, gotta remove `requirements.in` in both places.
```suggestion
   $ pip-compile --generate-hashes
```

docs/dev.rst
```diff
@@ -57,37 +57,20 @@ Managing dependencies
 Python dependencies
 -------------------
 
-Python dependencies for all parts of Socorro are split between two files:
+Python dependencies are maintained in the ``requirements.in`` file and
+"compiled" with dependencies of dependencies in the ``requirements.txt`` file.
```
relud, line 61:
```suggestion
"compiled" with hashes and dependencies of dependencies in the ``requirements.txt`` file.
```

relud (approved):
r+wc

docs/dev.rst
willkg: :+1:

docs/dev.rst
willkg: :+1:

docs/dev.rst
willkg: :+1:

.circleci/config.yml
```diff
@@ -32,6 +32,12 @@ jobs:
             # build tecken containers
             make build
 
+      - run:
+          name: Verify requirements.txt contains correct dependencies
+          command: |
+            pip-compile --quiet --generate-hashes requirements.in
```
relud, line 38:
```suggestion
            pip-compile --quiet --generate-hashes
```

relud:
```suggestion
            docker-compose run --rm --no-deps test bash -c 'pip-install pip-compile && pip-compile --quiet --generate-hashes'
```

requirements.txt
willkg: I had seen this when I was looking at how you used pip-tools, but then completely forgot to add it. Thank you! This will save hours of time at some point in the future when I'm all confused about what's going on.

bin/run_verify_reqs.sh
```diff
@@ -0,0 +1,12 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Verifies that the requirements file is built by the version of Python that
+# runs in the container.
+
+cp requirements.txt requirements.txt.orig
+pip-compile --quiet --generate-hashes
+diff requirements.txt requirements.txt.orig
```
willkg, line 12: This uses `diff` rather than `git diff` because `.git` isn't in the container. I wanted to run this script in the container because then it guarantees it's using the right version of Python and I don't have to deal with what I had before where we had multiple container specifications and I'd forget to update them all at the same time and then "strange things occur".

willkg:
Thank you!

PR #2249, opened 2020-10-06
bug 1669498: handle upload-by-download download url failure

willkg:
If the specified url in an upload-by-download request can't be downloaded, the `UploadByDownloadForm` will probably catch it when it does a HEAD request. It's vaguely possible that if the HEAD succeeds, the following GET might fail.

This adds some minor handling in the case where the HEAD succeeds and the GET fails. It returns an HTTP 400 telling the user the error. The user should check the download works and retry.


willkg:
self-r+

PR #2250, opened 2020-10-06
bug 1669519: always set a timeout when using requests

willkg:
The requests library defaults to an infinite timeout, so it's possible that a request could cause a Python process to hang forever. That's super bad for Tecken, so this fixes requests uses to always have a default timeout and always retry.


willkg:
Note that I didn't change this in the systemtests. I figured I'd leave that code as is since the retryable bits might paper over problems I want to stick out in the systemtests.

willkg:
self-r+

PR #2252, opened 2020-10-20
Update object-path to >=0.11.5

willkg:
This gets us past a security vulnerability with object-path < 0.11.5.

This also updates mobx and mobx-react because yarn couldn't resolve the
pinned versions of those dependencies with other things and it seemed
easier to update the pinned versions than spend time to sort it all out.


willkg:
I tested this locally and all the pages in the webapp seem to work fine.

self-r+

PR #2255, opened 2020-10-29
bug 1636210: implement eliot microservice for symbolication

willkg:
This implements a new microservice called Eliot which does symbolication.

It uses the Symbolic Python library for parsing sym files, generating symcache files, and looking up symbols.

It has an LRU disk cache for symcache files and a disk cache manager process to manage that cache.

It's built using the Falcon web framework. It currently has a very anemic `/`, but we can fix that going forward if we want. Maybe redirect to Tecken or the docs?

To test:

1. run `make test` which will run all the Tecken webapp tests and then the Eliot tests
2. perform symbolication against http://localhost:8050

Things to do:

* [x] review code
* [ ] figure out what metrics we want to generate and implement those -- might push this off to another PR
* [ ] figure out if INFO level logging has the information we want in it -- might push this off to another PR


docker-compose.yml
```diff
@@ -1,3 +1,4 @@
+---
```
willkg, line 1: yamllint kept complaining about this, so I fixed it.

docker/config/test.env
```diff
@@ -31,3 +34,13 @@ DJANGO_SYMBOL_FILE_PREFIX=v0
 DJANGO_UPLOAD_DEFAULT_URL=https://s3.example.com/private/prefix/
 DJANGO_UPLOAD_TRY_SYMBOLS_URL=https://s3.example.com/try/prefix
 DJANGO_UPLOAD_URL_EXCEPTIONS={"*peterbe.com": "https://s3.example.com/peterbe-com"}
+
+
+# Eliot settings
+# --------------
+
+HOST_ID=testnode
```
willkg, line 42: I think this is wrong. Should probably be `ELIOT_HOST_ID`.

docs/configuration.rst
```diff
@@ -360,3 +344,28 @@ You'll remain logged in to ``oidcprovider``, and the account will persist until
...
+
+Disk cache manager
+------------------
+
+The Eliot disk cache manager is run as a single process by Circus.
+
+.. autocomponent:: eliot.cache_manager.AppConfig
+   :hide-classname:
+   :namespace: eliot
+   :case: upper
```
willkg, line 371: It's super easy to document Eliot configuration because it uses Everett and Everett has a Sphinx extension.

eliot-service/eliot/cache_manager.py
```diff
@@ -0,0 +1,366 @@
...
+        try:
+            while True:
+                for event in inotify.read(timeout=timeout):
+                    LOGGER.debug(event)
+                    event_flags = flags.from_mask(event.mask)
+                    # Ignore changes to directories
+                    if flags.ISDIR in event_flags:
+                        continue
+
+                    path = self.cache_dir / event.name
```
willkg, line 274: The `event.name` is relative to the watched thing which in this case is a directory. If/when we rework this to support a directory tree, we'll have to do a lookup of the watch descriptor in the event back to the path that's being looked at and then use that to generate the full path here.

eliot-service/eliot/cache_manager.py
```diff
@@ -0,0 +1,366 @@
...
+                    elif flags.MODIFY in event_flags:
+                        self.process_modify(path, event)
+                    elif flags.DELETE in event_flags:
+                        self.process_delete(path, event)
+                    elif flags.ACCESS in event_flags:
+                        self.process_access(path, event)
+                    else:
+                        LOGGER.debug(f"ignored {path} {event}")
+
+                yield
```
willkg, line 290: YIELD!

eliot-service/eliot/cache_manager.py
```diff
@@ -0,0 +1,366 @@
...
+
+        finally:
+            inotify.rm_watch(watch_descriptor)
+
+    def run_loop(self):
+        """Run cache manager in a loop."""
+        for event in self.event_generator():
+            pass
```
willkg, line 298: `self.event_generator()` doesn't yield events, so this is confusing and should get fixed.

eliot-service/eliot/cache_manager.py
```diff
@@ -0,0 +1,366 @@
...
+
+    def process_modify(self, path, event):
+        """Process a modify event."""
+        LOGGER.debug(f"modify {path}")
+        self.update_bookkeeping(path)
+
+    def process_access(self, path, event):
+        """Process an access event."""
+        LOGGER.debug(f"access {path}")
+        self.mark_access(path)
```
willkg, line 328: These `process_*` methods will get more complex when we support watching directory trees, so it's best to keep the structure this way even if it's repetitive.

eliot-service/eliot/downloader.py
```diff
@@ -0,0 +1,157 @@
...
+
+        """
+        raise NotImplementedError
+
+
+class HTTPSource(Source):
+    """Source for HTTP/HTTPS requests."""
+
+    def __init__(self, source_url):
+        self.source_url = source_url.rstrip("/") + "/"
```
willkg, line 58: One of the things this makes difficult is if the key should go between two parts of a `source_url` and maybe the `source_url` should be more of a template. Then you could do this:

```
source_url="https://symbols.mozilla.org/%(debug_filename)s/%(debug_id)s/%(filename)s?try"
```

eliot-service/eliot/downloader.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        key = self._make_key(debug_filename, debug_id, filename)
+        url = "%s%s" % (self.source_url, key)
+
+        resp = self.session.get(url, allow_redirects=True)
+        if resp.status_code != 200:
+            METRICS.incr("sym_download", tags=["statuscode:%d" % resp.status_code])
+
+            # If the status_code is 404, that's a legitimate FileNotFound. Anything else
+            # is either fishy or a server error.
```
willkg, line 97: Note that the session is already set up to retry a whole series of conditions, so this is more like "anything else at this point after retrying a bunch of ephemeral universe hooey".

eliot-service/eliot/health_resource.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class BrokenResource:
+    """Handle ``/__broken__`` endpoint."""
+
+    def on_get(self, req, resp):
+        """Implement GET HTTP request."""
+        METRICS.incr("broken.count")
+        # This is intentional breakage
+        raise Exception("intentional exception")
```
willkg, line 27: We need to make sure that nginx is set up to protect this endpoint behind basic auth. I'll add that to the project plan.

eliot-service/eliot/health_resource.py
```diff
@@ -0,0 +1,110 @@
...
+        d = OrderedDict([("name", name), ("msg", msg)])
+        self.errors.append(d)
+
+    def is_healthy(self):
+        """Return whether this represents a healthy state."""
+        return len(self.errors) == 0
+
+    def to_dict(self):
+        """Convert state to a dict."""
+        return OrderedDict([("errors", self.errors), ("info", self.statsd)])
```
willkg, line 81: We don't need this. It was something I copied from Antenna which is a bit more complex.

eliot-service/eliot/index_resource.py
```diff
@@ -0,0 +1,16 @@
...
+"""
+
+import falcon
+
+
+class IndexResource:
+    def on_get(self, req, resp):
+        resp.content_type = "text/html"
+        resp.status = falcon.HTTP_200
+        resp.body = "<html><body>Eliot Index</body></html>"
```
willkg, line 16: We might want this to point to docs or something. We can figure it out later.

eliot-service/eliot/symbolicate_resource.py
```diff
@@ -0,0 +1,464 @@
...
+        :arg debug_id: the debug id
+        :arg data: bytes
+
+        :returns: symcache or None
+
+        """
+        try:
+            ndebug_id = symbolic.normalize_debug_id(debug_id)
+        except symbolic.ParseDebugIdError:
+            # If the debug id isn't valid, then there's nothing to pars3, so
```
willkg, line 201: `pars3` is interesting.

eliot-service/tests/test_cache.py
```diff
@@ -0,0 +1,106 @@
...
+        ("..//foo", "%(tmp)s/foo"),
+        # Test bad characters
+        ("/foo()*$", "%(tmp)s/foo____"),
+        # Test multiple / .
+        ("foo/bar/baz/baz2", "%(tmp)s/foo_bar_baz_baz2"),
+    ],
+    ids=counter(),
+)
+def test_diskcache_key_to_filepath(tmpdir, key, filepath):
+    diskcache = DiskCache(str(tmpdir))
```
willkg, line 36: Oops--DiskCache now takes two directories. These tests need to be fixed.

willkg:
The first of the month is coming, so I'm going to land this now. It won't affect deploys for the existing Tecken infrastructure and doesn't push out anything new. If I'm horribly mistaken about that, I have time to fix that before November comes and dependabot dumps a bunch of stuff on us.

I have a few things to think through. I'll write up bugs for those and proceed in future PRs.

PR #2267, opened 2020-11-20
bug 1674406: reimplement metrics in Eliot and add new ones

willkg:
This reimplements how metrics work in Eliot and adds the ones we need
for the bug.

The new system centralizes all the metrics that Eliot generates into a
single file. It adds an extension for extracting that data and
auto-generating documentation. It also shapes the metrics data as a
Markus MetricsFilter so it validates any metrics being emitted.

Adding a new metric involves adding it to ELIOT_METRICS and then adding
the code to emit the metric. That's about how much effort it was before,
so I think this seems promising. I'm inclined to redo Tecken webapp
accordingly.

### DIFF
--- a/docs/exts/eliot_metrics.py
+++ b/docs/exts/eliot_metrics.py
@@ -1,0 +1,109 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, you can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""Generates the documentation for Eliot metrics."""
+
+import importlib
+import sys
+import textwrap
+
+from docutils import nodes
+from docutils.parsers.rst import Directive
+from docutils.statemachine import ViewList
+
+
+def build_table(table):
+    col_size = [0] * len(table[0])
+    for row in table:
+        for i, col in enumerate(row):
+            col_size[i] = max(col_size[i], len(col))
+
+    col_size = [width + 2 for width in col_size]
+
+    yield "  ".join("=" * width for width in col_size)
+    yield "  ".join(
+        header + (" " * (width - len(header)))
+        for header, width in zip(table[0], col_size)
+    )
+    yield "  ".join("=" * width for width in col_size)
+    for row in table[1:]:
+        yield "  ".join(
+            col + (" " * (width - len(col)))
+            for col, width in zip(row, col_size)
+        )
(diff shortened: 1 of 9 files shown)


docs/exts/eliot_metrics.py
```diff
@@ -0,0 +1,109 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
willkg, line 1: This is kind of nuts. I'm inclined to fold this proof-of-concept into Markus, but I want to bounce it off some people, first.

The Markus issue is https://github.com/willkg/markus/issues/15 .

docs/exts/eliot_metrics.py
```diff
@@ -0,0 +1,109 @@
...
+
+        for line in build_table(table):
+            self.add_line(line, sourcename)
+
+        self.add_line("", sourcename)
+        self.add_line("Metrics details:", sourcename)
+        self.add_line("", sourcename)
+
+        for key, metric in metrics.items():
+            self.add_line(".. py:data:: %s" % key, sourcename)
```
willkg, line 75: Metrics aren't really "py:data", but I didn't want to write another directive and (ab)using "py:data" gives us what we want: something we can reference, looks ok in HTML, and generates an index entry.

eliot-service/eliot/libmarkus.py
```diff
@@ -40,3 +185,28 @@ def setup_metrics(app_config, logger=None):
...
+        metric = self.metrics.get(record.key)
+        if metric is None:
+            LOGGER.warning("metrics key %r is unknown", record.key)
+
+        elif record.stat_type != metric.stat_type:
+            LOGGER.warning(
+                "metrics key %r has wrong type; got %s expecting %s",
+                record.key,
+                record.stat_type,
+                metric.stat_type,
```
willkg, line 206: This doesn't validate the tags. I thought about that, but it seems too burdensome to do all the time.

Another thought I had was to turn the filter on and off depending on whether you're in a server environment so it's not doing extra work validating something that should get caught in development or testing.

eliot-service/eliot/symbolicate_resource.py
```diff
@@ -420,10 +441,12 @@ def frame_to_function(frame):
 
 
 class SymbolicateV5(SymbolicateBase):
+    @METRICS.timer_decorator("eliot.symbolicate.api", tags=["version:v4"])
```
willkg, line 444: Oops--this should be v5.

PR #2269, opened 2020-11-30
bug 1674406: change gauge to histogram

willkg:
This changes the gauge metrics for Eliot to histograms.

When doing that, I hit some issues and decided that logging a warning in
the local dev environment when metrics were wrong wasn't good enough--we
really want them to throw exceptions. So I tweaked that code so that
the filter only runs in local dev environment and tests and raises
exceptions.


willkg:
self-r+

PR #2277, opened 2020-12-09
Update pip-tools to 5.4.0

willkg:
This fixes the issue where pip-compile was using something in pip that
isn't there anymore and throwing a TypeError.


willkg:
self-r+

PR #2289, opened 2021-01-25
bug 1688623: add docs for code_id and code_filename

willkg:
The download API takes two additional querystring params that weren't
documented: `code_id` and `code_filename`.

These are in the HTTP requests from minidump-stackwalk in Socorro's
processor. I don't know what else uses these parameters.


willkg:
https://github.com/mozilla-services/minidump-stackwalk/blob/e8f983e6fc2c463072e0967e6cebf082680478b4/minidump-stackwalk/http_symbol_supplier.cc#L240-L252

PR #2319, opened 2021-03-15
bug 1696585: switch to honcho for eliot

willkg:
This switches Eliot to use honcho for process management. We're going to
standardize on honcho for cases where we need to run multiple processes
in the Docker container.


willkg:
After this lands and deploys to stage, we should check the logs to make sure logging is still working.

PR #2328, opened 2021-03-30
bug 1699546: return correct value for "module" in symbolication

willkg:
For Windows SYM files, the debug_filename is something.pdb, but we
really want to return something.dll which is the PE filename. SYM files
encode this in the INFO line.

symbolic parses that line and captures that information, but doesn't
make it available in the symcache, so we had to add some code to peek in
the SYM file and pull out the PE filename if it's there. But then since
we're caching symcache files on disk, we needed to redo disk caching to
cache items that consist of multiple key/val pairs.

This stores the key/val data as a single file using msgpack. That way
the disk cache LRU can think about files as units and we don't have to
deal with the complexities of other possibilities.

### DIFF
--- a/eliot-service/eliot/cache.py
+++ b/eliot-service/eliot/cache.py
@@ -12,6 +12,9 @@ import re
 import tempfile
 import time
 
+import msgpack
+import msgpack.exceptions
+
 from eliot.libmarkus import METRICS
 
 
@@ -21,8 +24,12 @@ LOGGER = logging.getLogger(__name__)
 NO_DEFAULT = object()
 
 
+class CacheReadError(Exception):
+    """Exception for errors hit when reading the cache from disk"""
+
+
 class DiskCache:
-    """Disk cache of symcache files.
+    """Disk cache of msgpack data files
 
     NOTE(willkg): This sets and checks the cache--it doesn't do any cleanup. The Disk
     Cache Manager watches the disk and enforces a max size by evicting least recently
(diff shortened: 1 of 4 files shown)


eliot-service/eliot/cache.py
```diff
@@ -75,14 +82,65 @@ def __contains__(self, key):
...
+    def read_from_file(self, filepath):
+        """Reads data from a file
+
+        This reads from a file, unpacks it and returns the data.
+
+        :arg Path filepath: the file to read from
+
+        :returns: data as dict
+
+        """
```
willkg, line 94: This raises an exception and that should be in the docstring.

eliot-service/eliot/cache.py
```diff
@@ -75,14 +82,65 @@ def __contains__(self, key):
...
+        """Write data to a file
+
+        This converts the data to msgpack then saves it to disk. It tries to account for
+        race conditions between reads and writes by writing to a temporary file and then
+        renaming it.
+
+        :arg Path filepath: the file to write to
+        :arg dict data: the data to write
+
+        """
```
willkg, line 115: This returns True/False depending on success.

eliot-service/eliot/cache.py
```diff
@@ -119,29 +177,16 @@ def set(self, key, data):
...
-
-            filepath.parent.mkdir(parents=True, exist_ok=True)
-            Path(temp_fp.name).rename(filepath)
-            result = "success"
-
-        except OSError:
-            LOGGER.exception("Exception when writing to disk cache")
-            result = "fail"
+        ret = self.write_to_file(filepath, data)
+        result = "success" if ret else "fail"
```
willkg, line 189: I kind of wish Markus grew a "structured metrics" system so I could store these tags in a context from wherever and pull them all out at once to generate the metric. Then I wouldn't have to do this awkward carrying through of values.

eliot-service/eliot/symbolicate_resource.py
```diff
@@ -15,7 +15,7 @@
 
    For example, ``11FB836EE6723C07BFF775900077457B0``.
 
-``filename``
```
willkg, line 18: "filename" is super ambiguous, so I tried to clarify it.

eliot-service/eliot/symbolicate_resource.py
```diff
@@ -149,6 +155,58 @@ def validate_stacks(stacks, modules):
                 )
 
 
+def bytes_split_generator(item, sep):
```
willkg, line 158: SYM files are _LARGE_, so I needed something that would parse just the part I need to look at.

eliot-service/eliot/symbolicate_resource.py
```diff
@@ -149,6 +155,58 @@ def validate_stacks(stacks, modules):
...
+    :arg bytes sym_file: the sym file as bytes
+    :arg str debug_filename: the debug filename
+
+    :returns: the module filename
+
+    """
+    # Iterate through the first few lines of the file until we hit FILE in which
+    # case there's no INFO for some reason or we hit the first INFO.
+    for line in bytes_split_generator(sym_file, b"\n"):
+        if line.startswith(b"INFO"):
```
willkg, line 197: I'm basing this structure on what dump_syms outputs where INFO is close to the top of the file.

eliot-service/eliot/symbolicate_resource.py
```diff
@@ -149,6 +155,58 @@ def validate_stacks(stacks, modules):
...
+            parts = line.split(b" ")
+            if len(parts) == 4:
+                return parts[-1].decode("utf-8").strip()
+            else:
+                break
+
+        elif line.startswith(b"FILE"):
```
willkg, line 204: We might want to add PUBLIC and FUNC here, too. I think all three of those indicate we went too far in the file and we're done looking.

eliot-service/eliot/symbolicate_resource.py
```diff
@@ -245,46 +303,58 @@ def parse_sym_file(self, debug_filename, debug_id, data):
...
 
-        This uses the cachemanager and downloader to get the symcache.
+        This uses the cachemanager and downloader to get the symcache. It modifies
+        the Module in place.
 
-        :arg debug_filename: the debug filename
-        :arg debug_id: the debug id
-
-        :returns: symcache or None
+        :arg module: the Module data class
```
willkg, line 312: I should rename Module to ModuleInfo or something like that.

PR #2330, opened 2021-03-30
bug 1674406: measure disk cache usage

willkg:
This fixes an issue in make-stacks.py that I hit while working on the disk cache evictions issue.

This also adds an `eliot.diskcache.usage` gauge so we can see how the disk cache grows over time.


willkg:
One of the things this will help with is that I haven't seen an `eliot.diskcache.evict` go by, yet. `eliot.diskcache.usage` should get emitted more regularly, so we should see whether the cache manager is emitting anything at all.

PR #2382, opened 2021-07-27
bug 1722602: update to everett 2.0.0

willkg:
This updates Eliot to Everett 2.0.0 and reworks component configuration.
It also fixes `setup_metrics` and `setup_logging` so they're explicit about
what argument data is being used.

### DIFF
--- a/eliot-service/eliot/app.py
+++ b/eliot-service/eliot/app.py
@@ -10,8 +10,13 @@ import logging
 import logging.config
 from pathlib import Path
 
-from everett.manager import ConfigManager, ConfigOSEnv, ListOf
-from everett.component import ConfigOptions, RequiredConfigMixin
+from everett.manager import (
+    ConfigManager,
+    ConfigOSEnv,
+    get_config_for_class,
+    ListOf,
+    Option,
+)
 import falcon
 
 from eliot.cache import DiskCache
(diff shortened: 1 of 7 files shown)


eliot-service/eliot/app.py
```diff
@@ -56,67 +61,55 @@ def build_config_manager():
...
+        )
+        symbols_cache_dir = Option(
+            default="/tmp/cache",
+            doc="Location for caching symcache files.",
+        )
+        symbols_urls = Option(
+            default="https://symbols.mozilla.org/try/",
+            doc="Comma-separated list of urls to pull symbols files from.",
+            parser=ListOf(str),
+        )
```
willkg, line 112: Note that the configuration is the same, so we don't have to make any changes to infra.

eliot-service/eliot/app.py
```diff
@@ -132,7 +125,7 @@ def verify_configuration(self):
         This will raise a configuration error if something isn't right.
 
         """
-        for key, opt in self.required_config.options.items():
+        for key, val in get_config_for_class(self.__class__).items():
             self.config(key)
```
willkg, line 129: Eliot doesn't have any namespaced components, so this works fine. If it did have namespaced components, I'm pretty sure this wouldn't work.

eliot-service/eliot/cache_manager.py
```diff
@@ -57,71 +57,59 @@ def handle_exception(exctype, value, tb):
...
+            doc="Location for caching symcache files.",
+        )
+        symbols_cache_max_size = Option(
+            default=str(1024 * 1024 * 1024),
+            parser=int,
+            doc=(
+                "Max size (bytes) of symbols cache. You can use _ to group digits for "
+                "legibility."
+            ),
+        )
```
willkg, line 112: Marvel at how much boilerplate we removed!

PR #2393, opened 2021-08-20
bug 1726652: add ADR scaffolding and symbolication microservice ADR

willkg:
This adds the scaffolding for creating and maintaining ADRs. It adds an
ADR log to readthedocs.

It adds an ADR for breaking out the symbolication API as a microservice.


willkg:
We're in a code freeze, but these changes are entirely to docs/ and nothing else.

PR #2397, opened 2021-08-26
bug 1673887: rewrite disk cache manager

willkg:
This rewrites the disk cache manager to support nested directories. It cleans the code up a bit and makes it easier to follow as well.

This rewrite allows us to re-use it for Socorro where the thing populating the disk cache is minidump-stackwalk.


docker/config/local_dev.env
```diff
@@ -36,3 +36,4 @@ ELIOT_LOGGING_LEVEL=INFO
 ELIOT_STATSD_HOST=statsd
 ELIOT_STATSD_NAMESPACE=mcboatface
 ELIOT_STATSD_PORT=8125
+ELIOT_SYMBOLS_CACHE_MAX_SIZE=400000000
```
willkg, line 39: Setting the cache max size to 400,000,000 forces the symbols cache to update regularly. It can hold about 40 files in it.

eliot-service/eliot/cache.py
```diff
@@ -38,8 +38,7 @@ class DiskCache:
 
     """
 
-    # NOTE(willkg): this puts all files in the same directory--no directory trees
-    BAD_CHARS = re.compile(r"[^A-Za-z0-9._-]")
+    BAD_CHARS = re.compile(r"[^A-Za-z0-9._/-]")
```
willkg, line 41: I had to add `/` here. I waffled on whether to use `os.sep` so the system didn't have one more reason it couldn't work on Windows, but that gets tricky for a bunch of reasons. I ended up going with this because we run Eliot in a Docker container anyhow and that's the only runtime environment I'm supporting currently.

eliot-service/eliot/cache_manager.py
```diff
@@ -26,10 +26,10 @@
 import sys
 import traceback
 
+from boltons.dictutils import OneToOne
```
willkg, line 29: boltons has a `OneToOne` map which simplifies keeping track of each direction of the map. There are some other things in boltons that I think I want to use, so I decided it's worth pulling it in as a dependency now.

eliot-service/eliot/cache_manager.py
```diff
@@ -157,80 +165,57 @@ def verify_configuration(self):
...
-        if lru_key in self.lru and path.exists():
-            self.lru.move_to_end(lru_key, last=True)
-            return
+    def remove_watch(self, path):
+        if path in self.watches:
+            del self.watches[path]
+        else:
+            print(
+                f"something is wrong--{path} not in watches ({list(self.watches.keys())})"
+            )
```
willkg, line 178: TODO: Nix the print statement.

eliot-service/eliot/cache_manager.py
```diff
@@ -157,80 +165,57 @@ def verify_configuration(self):
...
+    def make_room(self, size):
+        total_size = self.total_size + size
+        removed = 0
+
+        while self.lru and total_size > self.max_size:
+            rm_path, rm_size = self.lru.popitem(last=False)
+            total_size -= rm_size
+            removed += rm_size
+            os.remove(rm_path)
+            print(f"evicted {rm_path} {rm_size:,d}")
```
willkg, line 215: TODO: convert to LOGGER.debug.

eliot-service/eliot/cache_manager.py
```diff
@@ -241,50 +226,113 @@ def _event_generator(self, nonblocking=False):
...
-            | flags.MODIFY
-            | flags.MOVED_FROM
-            | flags.MOVED_TO
-        )
-        watch_descriptor = inotify.add_watch(str(self.cachedir), watch_flags)
+        self.inotify = INotify(nonblocking=nonblocking)
+
+        # Set up watches and LRU with what exists already
+        self.watches = OneToOne()
+        self.lru = OrderedDict()
```
willkg, line 233: Maybe it's worth subclassing OrderedDict to an LRU with a clearer API. Something like:

```
# Add or update an item in the LRU
self.lru[path] = size

# Update an item in the LRU
self.lru.touch(path)  # or something like that

# Delete an item from the LRU
del lru[path]
```

Right now we're using `move_to_end` which is an implementation detail and in one case when we update the size, we have to update the size and call `move_to_end`, but updating the size should move it to the end.

I'll think about doing this.

eliot-service/eliot/cache_manager.py
```diff
@@ -241,50 +226,113 @@ def _event_generator(self, nonblocking=False):
...
+        self.total_size = 0
+        self.inventory_existing()
+
+        LOGGER.info(f"Found {len(self.lru)} files ({self.total_size:,d} bytes)")
```
willkg, line 237: TODO: go through and lowercase all the log statements

eliot-service/eliot/cache_manager.py
```diff
@@ -241,50 +226,113 @@ def _event_generator(self, nonblocking=False):
...
-                    LOGGER.debug(event)
+                for event in self.inotify.read(timeout=timeout):
+                    processed_events = True
                     event_flags = flags.from_mask(event.mask)
-                    # Ignore changes to directories
-                    if flags.ISDIR in event_flags:
+
+                    LOGGER.debug(f"EVENT: {event}")
+                    for flag in event_flags:
+                        LOGGER.debug(f"    {str(flag)}")
```
willkg, line 250: TODO: change this to a single `LOGGER.debug` call.

eliot-service/eliot/cache_manager.py
```diff
@@ -241,50 +226,113 @@ def _event_generator(self, nonblocking=False):
...
+                                self.remove_watch(path)
+
                     else:
-                        LOGGER.debug(f"ignored {path} {event}")
+                        # Handle file events which update our LRU cache
+                        if flags.CREATE in event_flags:
+                            if path in self.lru:
+                                print(
+                                    f"something is wrong--create on file that exists {path}"
+                                )
```
willkg, line 273: TODO: fix this print statement

eliot-service/tests/test_symbolicate_resource.py
```diff
@@ -294,7 +294,8 @@ def test_get_symcache_in_cache(self, tmpcachedir, tmpdir):
         symcache.dump_into(data)
         data = {"filename": debug_filename, "symcache": data.getvalue()}
 
-        cache_key = "%s___%s.symc" % (debug_filename, debug_id.upper())
+        cache_key = "%s/%s.symc" % (debug_filename, debug_id.upper())
+        print(cache_key)
```
willkg, line 298: TODO: remove print statement

willkg:
Merging into the eliot-dev branch.

PR #2398, opened 2021-08-30
bug 1728156: add tecken overview

willkg:
This adds an overview to the Tecken docs to give a little more context
as to what Tecken is and how it's structured.


docs/Makefile
```diff
@@ -18,3 +18,10 @@ help:
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+.PHONY: graphviz
+graphviz:
+	# Compile graphviz dot diagrams to PNG images
+	#
+	# Note: This requires graphviz be installed on your machine.
+	dot diagrams/arch.gv -Tpng -o images/arch.png
```
willkg, line 27: With Socorro, I didn't want to install graphviz in the production container, so I did a docs container. But that adds a bunch of stuff to maintain. For example, every month I get a PR to update the Python image it's based on.

This time around, I'm assuming the developer has access to graphviz and can run it on their host. It's less self-contained, but it's less maintenance, too. This diagram is probably not going to be edited any time soon, so this seems like a better option for now.

PR #2400, opened 2021-08-31
bug 1728210: remove celery tasks

willkg:
This is the first step in removing celery usage. This removes all the
tasks and folds the important bits into the code that kicks off those
tasks.

* for tecken.download, if there's a problem saving the missing symbol to
  the db, we let it go instead of spinning that off into a task
* for tecken.upload, we fold the uploads_created bookkeeping into the
  upload task; prod dashboard shows this takes a few seconds at most and
  usually it's under half a second
* don't need a sample_task or the infrastructure to test that celery is
  working anymore
* for tecken.upload, we nix invalidating the symbolicate cache--this
  code isn't long for this world and socorro doesn't invalidate its
  cache on uploads, to I think it's ok to degrade the service here


tecken/tests/conftest.py
```diff
@@ -170,42 +161,6 @@ def fakeuser():
...
-
-    fake_task = FakeTask()
-
-    _mock_function = "tecken.upload.views.invalidate_symbolicate_cache_task"
-    with mock.patch(_mock_function, new=fake_task):
-        yield fake_task
-
-
 @pytest.fixture
 def upload_mock_update_uploads_created_task():
```
willkg, line 165: TODO: remove this, too.

PR #2416, opened 2021-09-16
bug 1730950: fix /__broken__ so it sends errors to sentry

willkg:
The purpose of `/__broken__` is to help us test Sentry connectivity and also test uncaught error handling. However, Falcon changed how uncaught error handling works in 3.0.0 and we needed to update Eliot to work with that. This fixes the handling so that it works with Falcon 3.

### DIFF
--- a/eliot-service/eliot/app.py
+++ b/eliot-service/eliot/app.py
@@ -18,6 +18,7 @@ from everett.manager import (
     Option,
 )
 import falcon
+from falcon.errors import HTTPInternalServerError
 
 from eliot.cache import DiskCache
 from eliot.downloader import SymbolFileDownloader
@@ -143,6 +144,9 @@ class EliotApp(falcon.App):
         # Set up Sentry exception logger if we're so configured
         setup_sentry_logging()
 
+        # Set up uncaught error handler
+        self.add_error_handler(Exception, self.uncaught_error_handler)
+
         # Log application configuration
         log_config(LOGGER, self.config_manager, self)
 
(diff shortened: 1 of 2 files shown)


eliot-service/eliot/app.py
```diff
@@ -143,6 +144,9 @@ def setup(self):
         # Set up Sentry exception logger if we're so configured
         setup_sentry_logging()
 
+        # Set up uncaught error handler
```
willkg, line 147: This probably doesn't need a comment but it seemed aesthetically nicer given the lines around it all have comments.

eliot-service/eliot/libsentry.py
```diff
@@ -47,7 +45,9 @@ def setup_sentry_logging():
     if _SENTRY_CLIENT:
         handler = SentryHandler(_SENTRY_CLIENT)
         handler.setLevel(logging.ERROR)
-        setup_logging(handler)
```
willkg, line 50: I read through the Raven code for `setup_logging` and all it does is add the handler to the root logger. Raven is deprecated and not getting any more dev time, so I figured I'd save future me some effort and swap that line out for manually adding the handler to the logger.

eliot-service/eliot/libsentry.py
```diff
@@ -91,17 +91,17 @@ def __call__(self, environ, start_response):
                 [("content-type", "application/json; charset=utf-8")],
                 exc_info,
             )
-            return [b'{"msg": "COUGH! Internal Server Error"}']
+            return [b'{"msg": "Internal Server Error"}']
```
willkg, line 94: Coughing isn't funny anymore.

PR #2420, opened 2021-09-21
bug 1699776: add exception notification to loop and shutdown

willkg:
This adds exception notification to the event handling loop. If there
are unhandled exceptions, they'll get sent to Sentry. It's possible the
unhandled exception is not ephemeral, so this keeps track of how many
errors of this kind it's hit and terminates if there are too many. I
don't know if 10 is a magic number, but seems like a good starting point
for a disk cache manager.

This also adds exception handling to shutdown when it's removing inotify
watches.

### DIFF
--- a/eliot-service/eliot/cache_manager.py
+++ b/eliot-service/eliot/cache_manager.py
@@ -45,6 +45,7 @@ else:
 
 LOGGER = logging.getLogger(MODULE_NAME)
 REPOROOT_DIR = str(pathlib.Path(__file__).parent.parent.parent)
+MAX_ERRORS = 10
 
 
 class LastUpdatedOrderedDict(OrderedDict):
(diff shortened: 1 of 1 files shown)


eliot-service/eliot/cache_manager.py
```diff
@@ -252,77 +253,90 @@ def _event_generator(self, nonblocking=False):
...
-                                self.lru[path] = size
-                                self.total_size += size
-
-                        elif flags.MOVED_FROM in event_flags:
-                            if path in self.lru:
-                                # If it was moved out of this directory, then treat it
-                                # like a DELETE
-                                size = self.lru.pop(path)
-                                self.total_size -= size
+                try:
```
willkg, line 259: This is all indented in a `try` block.

eliot-service/eliot/cache_manager.py
```diff
@@ -252,77 +253,90 @@ def _event_generator(self, nonblocking=False):
...
+                                if path in self.lru:
+                                    # If it was moved out of this directory, then treat
+                                    # it like a DELETE
+                                    size = self.lru.pop(path)
+                                    self.total_size -= size
+
+                            else:
+                                LOGGER.debug(f"ignored {path} {event}")
+
+                except Exception:
```
willkg, line 332: This is the new part.

PR #2421, opened 2021-09-21
bug 1731412: downgrade to node 14

willkg:
This downgrades Tecken to using node 14. Tecken only uses node to build static assets, so as long as they're building fine, it doesn't really matter what version of node we're using.

This allows people using macs with m1 processors to build and run Tecken. Previous to this fix, they would hit this issue:

https://github.com/docker/for-mac/issues/5831

Just to clarify, this isn't a great fix, but it kicks the can down the road a bit until when we have more time to fix it more thoughtfully.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,8 @@
 # NOTE(willkg): Make sure to update frontend/Dockerfile when you update this
-FROM node:16.9.0-slim@sha256:fca208cb6a90179412cc78ac590a81c3b802b2d3a66b202e31586a9b702e101c as frontend
+#
+# NOTE(willkg): We're using node 14 because of this issue:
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # NOTE(willkg): Make sure to update frontend/Dockerfile when you update this
-FROM node:16.9.0-slim@sha256:fca208cb6a90179412cc78ac590a81c3b802b2d3a66b202e31586a9b702e101c as frontend
+FROM node:14.17.6-slim@sha256:56e3ae9f6981acb5525427c37eacc73c8ad400f0fef155c4064447f220816cbc as frontend
```
willkg, line 2: Oops--I both missed this note and didn't update the other docker file and also forgot to add a note about why we're using node 14.

PR #2431, opened 2021-10-07
bug 1733636: redo Eliot index page

willkg:
This adds scaffolding for styling and JS for the Eliot index page. It's
not the most optimized page size-wise, but it requires minimal
maintenance and can just sit around.

We can improve it going forward and we'll probably want to change how
it's built to match Tecken when we overhaul the Tecken frontend ui.


eliot-service/eliot/app.py
```diff
@@ -62,6 +62,19 @@ def build_config_manager():
...
 
 
+class IndexResource:
+    def __init__(self, staticroot_dir):
+        self.staticroot_dir = staticroot_dir
+        self.data = (Path(self.staticroot_dir) / "index.html").read_text()
+
+    def on_get(self, req, resp):
+        # FIXME
+        self.data = (Path(self.staticroot_dir) / "index.html").read_text()
```
willkg, line 72: Oops.

PR #2471, opened 2022-01-10
bug 1749309: rework cleanse_upload to skip SELECTs

willkg:
Because FileUpload has `on_delete=SET_NULL`, when we delete Upload from the db, it does a SELECT for all the Upload so it can update all the FileUpload. That's a ridiculous amount of data and when cleaning our stage/prod databases, it makes it impossible to run the command.

This (ab)uses a private method `_raw_delete()` to delete the data without doing the SELECTs. I think this should work going forward--Django 4.0 still has it.

This also prints the queries that run so we can more easily use/debug this going forward.

### DIFF
--- a/tecken/upload/management/commands/cleanse_upload.py
+++ b/tecken/upload/management/commands/cleanse_upload.py
@@ -5,6 +5,7 @@
 import datetime
 
 from django.core.management.base import BaseCommand
+from django.db import connection, reset_queries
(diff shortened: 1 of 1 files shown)


willkg:
This works locally and also the SQL queries that get printed out suggests that it works. After landing it, we'll test it in stage, but I'm not sure there's enough data in stage for us to know much.

PR #2476, opened 2022-01-13
bug 1750048: fix readthedocs builds

willkg:
It's hard to tell why the docs aren't building.

This reworks the readthedocs configuration file to match version 2 of their configuration spec.


willkg:
This might fix it? I have no idea. We'll have to see.

PR #2497, opened 2022-02-01
bug 1703253: update django to 3.2.12 and add migration status to site settings admin page

willkg:


willkg:
Once we land this, I can see the migration status of stage and then we can proceed with redoing the auto field thing.

PR #2505, opened 2022-02-17
Update dockerflow to 2022.1.0

willkg:
This updates dockerflow to 2022.1.0 which fixes one of the warnings we
were ignoring in the pytest configuration.


tecken/pytest.ini
```diff
@@ -15,6 +15,3 @@ filterwarnings =
     # dockerflow defines default_app_config in dockerflow.django.__init__ and
     # shouldn't anymore
     ignore:.*default_app_config:django.utils.deprecation.RemovedInDjango41Warning
```
willkg, line 17: This is still there. I wrote up an issue for it.

PR #2531, opened 2022-03-31
bug 1668371: Async fetch Files and Aggregates

edugfilho:
Related to bug 1668371 (doesn't fix it)

This change addresses the slowness on the File Uploads page by splitting that page into two different components that will load asynchronously (note the UI doesn't change, only how the components load): 
* "Files", which populates a table with a list of files and 
* "FilesAggregates", which provides total count, avg size and avg upload time. 

Local tests have indicated usability will increase since users will be able to visualize the list of files quicker while the Aggregates part, slower to fetch, loads in parallel.

This change was also done to make further work on optimizing performance of that page easier, by allowing us to tweak the distinct components individually.

### DIFF
--- a/frontend/src/Files.js
+++ b/frontend/src/Files.js
@@ -59,7 +59,7 @@ class Files extends React.PureComponent {
     }
   }
 
-  _fetch = (url, callback, updateHistory = true) => {
+  _fetch = (url, callback, errorCallback, updateHistory = true) => {
     const qs = filterToQueryString(this.state.filter);
 
     if (qs) {
@@ -85,6 +85,9 @@ class Files extends React.PureComponent {
         return r.json().then((response) => callback(response));
       } else {
         store.fetchError = r;
+        if (errorCallback) {
+          errorCallback();
+        }
       }
     });
   };
@@ -103,8 +106,16 @@ class Files extends React.PureComponent {
       });
       this.setState({ loadingFiles: false });
     };
+    var errorCallback = () => {
+      this.setState({ loadingFiles: false });
+    };
     this.setState({ loadingFiles: true });
-    this._fetch("/api/uploads/files/content", callback, updateHistory);
+    this._fetch(
+      "/api/uploads/files/content",
+      callback,
+      errorCallback,
+      updateHistory
+    );
   };
 
   _fetchAggregates = () => {
@@ -115,8 +126,11 @@ class Files extends React.PureComponent {
       });
       this.setState({ loadingAggregates: false });
     };
+    var errorCallback = () => {
+      this.setState({ loadingAggregates: false });
+    };
     this.setState({ loadingAggregates: true });
-    this._fetch("/api/uploads/files/aggregates", callback);
+    this._fetch("/api/uploads/files/aggregates", callback, errorCallback);
   };
 
   filterOnAll = (event) => {
@@ -161,12 +175,14 @@ class Files extends React.PureComponent {
         {this.state.loadingFiles ? (
           <Loading />
         ) : (
-          <TableSubTitle
-            total={this.state.total}
-            page={this.state.filter.page}
-            batchSize={this.state.batchSize}
-            calculating={this.state.loadingAggregates}
-          />
+          this.state.files && (
+            <TableSubTitle
+              total={this.state.total}
+              page={this.state.filter.page}
+              batchSize={this.state.batchSize}
+              calculating={this.state.loadingAggregates}
+            />
+          )
         )}
 
         {!this.state.loadingFiles && this.state.files && (
@@ -185,7 +201,7 @@ class Files extends React.PureComponent {
           <Loading />
         )}
 
-        {!this.state.loadingAggregates && (
+        {!this.state.loadingAggregates && this.state.files && (
           <DisplayFilesAggregates
             aggregates={this.state.aggregates}
             total={this.state.total}
--- a/frontend/src/Files.js
+++ b/frontend/src/Files.js
@@ -59,13 +59,11 @@ class Files extends React.PureComponent {
     }
   }
 
-  _fetch = (selector, callback, updateHistory = true) => {
+  _fetch = (url, callback, updateHistory = true) => {
     const qs = filterToQueryString(this.state.filter);
-    const qSelector = selector ? "selector=" + selector : "";
 
-    var url = "/api/uploads/files?" + qSelector;
     if (qs) {
-      url += "&" + qs;
+      url += "?" + qs;
     }
     this.props.history.push({ search: qs });
 
@@ -106,7 +104,7 @@ class Files extends React.PureComponent {
       this.setState({ loadingFiles: false });
     };
     this.setState({ loadingFiles: true });
-    this._fetch("files", callback, updateHistory);
+    this._fetch("/api/uploads/files/content", callback, updateHistory);
   };
 
   _fetchAggregates = () => {
@@ -118,7 +116,7 @@ class Files extends React.PureComponent {
       this.setState({ loadingAggregates: false });
     };
     this.setState({ loadingAggregates: true });
-    this._fetch("aggregates", callback);
+    this._fetch("/api/uploads/files/aggregates", callback);
   };
 
   filterOnAll = (event) => {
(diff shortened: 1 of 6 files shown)


tecken/api/views.py
```diff
@@ -614,9 +622,21 @@ def upload_files(request):
...
+    start = (page - 1) * batch_size
+    end = start + batch_size + 1
+    has_next = False
 
     upload_ids = set()
-    for file_upload in qs.order_by("-created_at")[start:end]:
+    file_uploads = qs.order_by("-created_at")[start:end]
+    if len(file_uploads) == batch_size + 1:
+        has_next = True
+    for file_upload in file_uploads[:batch_size]:
```
willkg, line 639: I think line 637 triggers a sql query and then this line triggers a sql query. If we're trying to optimize sql queries out of this, then maybe there's a better way to determine has_next? Like maybe you could do something like:

```
for i, file_upload in enumerate(file_uploads):
    ...
    if i > batch_size:
        has_next = True
```

tecken/api/views.py
```diff
@@ -590,11 +584,25 @@ def upload_files(request):
...
+        pagination_form = PaginationForm(request.GET)
+        if not pagination_form.is_valid():
+            return http.JsonResponse({"errors": pagination_form.errors}, status=400)
+        page = pagination_form.cleaned_data["page"]
+        context = _upload_files_content(page, qs)
+
+    if "aggregates" in selector:
+        aggregates = _upload_files_aggregates(qs)
+        context["aggregates"] = aggregates
+        context["total"] = aggregates["files"]["count"]
```
willkg, line 600: I'm not wild about using the same API endpoint in two selector modes because it doesn't move us a step towards a better and more usable API. It's fine for now, but if we decide that this solution path helps significantly and we're planning to stick with this until we do the rewrite, we might want to revisit this and break this into two separate API endpoints.

In the meeting, we talked about this a bit and how it's not clear the aggregates are helpful in the first place. Maybe we should ditch them altogether?

willkg (commented):
I can't really review the React side of this--I haven't done any React in years and I don't remember enough to context switch into it. I'll make sure to do some local testing when we're ready for that and hopefully that's good enough.

Let me know what you think about the other comments. They're more wonderings than requests for changes, so if you want to stick with this as it stands (barring the lint issues), that's cool.

tecken/api/views.py
edugfilho: right! Fixing that

tecken/api/views.py
edugfilho: I had different endpoints before, then turned it into this thinking it'd cleaner. But you're right, the current design is limiting. I'll split them up again

edugfilho:
> I can't really review the React side of this--I haven't done any React in years and I don't remember enough to context switch into it. I'll make sure to do some local testing when we're ready for that and hopefully that's good enough.
> 
> Let me know what you think about the other comments. They're more wonderings than requests for changes, so if you want to stick with this as it stands (barring the lint issues), that's cool.

the more feedback I get the better, especially in these early days since I'm calibrating.

tecken/api/views.py
edugfilho: About ditching the aggregates, less code is better, so if you say it's not so useful I'd be in favor of ditching it, yes. But would you mind doing that in a future step? 
Btw, do we also want to get rid of that total count on the top-left corner of the table?
Screen Shot 2022-04-01 at 12 22 50 PM

willkg: I don't mind pushing removing the aggregates off to a future step.

Knowing the total count is generally helpful. I think I've used that when filtering.

edugfilho:
@willkg I addressed your points: split the api endpoints and adapted the front-end + tests. It's ready for another pass when you're good

tecken/api/urls.py
```diff
@@ -36,6 +36,14 @@
         name="uploads_created_backfilled",
     ),
     path("uploads/files/", views.upload_files, name="upload_files"),
+    path(
+        "uploads/files/content", views.upload_files_content, name="upload_files_content"
```
willkg, line 40: Nit: This path should end in a `/`.

tecken/api/urls.py
```diff
@@ -36,6 +36,14 @@
...
+    ),
+    path(
+        "uploads/files/aggregates",
```
willkg, line 43: Nit: This path should end in a `/`.

tecken/api/views.py
```diff
@@ -589,34 +580,28 @@ def upload_files(request):
             include_bucket_names.append(bucket_name)
     if include_bucket_names:
         qs = qs.filter(bucket_name__in=include_bucket_names)
+    return qs
+
+
+def _upload_files_content(request, qs):
+    pagination_form = PaginationForm(request.GET)
+    if not pagination_form.is_valid():
+        raise BadRequest(pagination_form.errors)
```
willkg, line 589: Using `BadRequest` changes the response. With this change, my local dev environment returns this:

```
HTTP 400
Content-Type: application/json

{
  "errors": "<ul class=\"errorlist\"><li>page<ul class=\"errorlist\"><li>Not a number &#x27;abc&#x27;</li></ul></li></ul>"
}
```

Whereas the equivalent request on stage is this:

```
HTTP 400
Content-Type: application/json

{
  "errors": {
    "page": [
      "Not a number 'abc'"
    ]
  }
}
```

I think you want:

```
        return http.JsonResponse({"errors": pagination_form.errors}, status=400)
```

willkg (commented):
I had a couple of issues that should get looked at, but otherwise this look good.

After we land this, what's the plan? What do we want to measure in prod before we push this out? Do we need to put anything in place before we land this?

edugfilho:
Thanks a lot for testing and catching all of this. I thought I had the nav buttons fixed, but some of my most recent changes must have brought that bug that displays "Next page" on the last page back. I'll fix it

> I had a couple of issues that should get looked at, but otherwise this look good.
> 
> After we land this, what's the plan? What do we want to measure in prod before we push this out? Do we need to put anything in place before we land this?

My plan is to at least get some loading times from the current Upload Files page so we can compare it to the File List component of that page after we deploy. Something simple, can be manual or like running a jmeter script, without DoS-ing it.

After the deployment I'd like to measure the different api calls and act on the bottlenecks. My suspicion is that /aggregates will be the slowest, then we can decide to e.g. get rid of it or make optimizations. In case /content is still too slow I think we should prioritize it and look into cursor pagination and ask Jason for help with db performance.

How does that sound?

willkg:
This code is really hard in that there's no tests for it, so everything we do to QA it is manual. That's not a great situation to be in, but it is what it is for now and not something we can fix without spending some serious effort.

I think that plan sounds fine. :+1:

edugfilho:
> This code is really hard in that there's no tests for it, so everything we do to QA it is manual. That's not a great situation to be in, but it is what it is for now and not something we can fix without spending some serious effort.
> 
> I think that plan sounds fine. 👍

maybe you mean other types of tests but I spotted some some unit tests, that actually helped me during development. I even included a couple into test_api.py btw

willkg:
There are tests for Python code, however the frontend is predominantly JS and there's no tests for that.

tecken/api/views.py
edugfilho: I'll fix it and make sure to add a test for it too.
Main reason for raising BadRequest there is that if I return a JsonResponse it's not json serializable (thanks unit tests that caught that!). I can return `{"errors": pagination_form.errors}` and check for `"errors"` in the returned object outside this function. 
How's that sound? lmk if you got another idea

tecken/api/urls.py
edugfilho: fixed

tecken/api/urls.py
edugfilho: fixed

tecken/api/views.py
edugfilho: fixed

edugfilho:
@willkg I fixed the previous issues. You can continue the review

frontend/src/Files.js
```diff
@@ -80,25 +78,61 @@ class Files extends React.PureComponent {
...
+        batchSize: response.batch_size,
+      });
+      this.setState({ loadingFiles: false });
+    };
+    var errorCallback = () => {
+      this.setState({ loadingFiles: false });
+    };
+    this.setState({ loadingFiles: true });
+    this._fetch(
+      "/api/uploads/files/content",
```
willkg, line 114: This needs a `/` at the end. Otherwise, Django returns an HTTP 301 and this follows the redirect and does a second HTTP request.

edugfilho: fixed

willkg (approved):
Looks good!

PR #2538, opened 2022-04-01
Update dependencies

edugfilho:
Update dependencies. This covers:

* Bump symbolic from 8.5.0 to 8.7.0 (from PR #2537)
* Bump sentry-sdk from 1.5.6 to 1.5.8 (from PR #2536)
* Bump sphinx from 4.4.0 to 4.5.0 (from PR #2535)
* Bump pytest from 6.2.5 to 7.1.1 (from PR #2534)
* Bump Python image from 3.9.10-slim to 3.9.12-slim (manually)


willkg (approved):
Tecken's stage environment isn't used very much so unless we're testing on stage, we won't see any issues. I don't see anything in here that gives me pause.

Checks pass so it's probably fine. :)

willkg:
Oh, one other thing--can you update the Python image used in the Dockerfile? It's currently using 3.9.10-slim, but we should update to 3.9.12-slim. Dependabot won't do that because it really wants us to update to 3.10.something and we can't do that because of that bug.

willkg (approved):

PR #2539, opened 2022-04-05
Update release.py

willkg:
```
ae2b6c4 Add bug checking
d5d4946 Update black dev requirement
25c0900 Default Priority to P2 (#18)
16d1a9a Add tag summary line (#19)
```

I tested this with Antenna and Socorro already and the fixes worked great.

### DIFF
--- a/bin/release.py
+++ b/bin/release.py
@@ -13,7 +13,7 @@ both.
 This requires Python 3 to run.
 
 repo: https://github.com/willkg/socorro-release/
-sha: 522573cab99d7d7de2106853f3283d30bf88aaa9
+sha: a3c6e973dd90ecd8a805747c5a5346bf45cb88d4
 
 """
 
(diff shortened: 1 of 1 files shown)


willkg:
@edugfilho I used this version of `bin/release.py` to deploy antenna and socorro just now, so it works. Do the changes look ok to you?

The `release.py` script comes from https://github.com/willkg/socorro-release/ and I do testing in https://github.com/willkg/socorro-release-test/ .

bin/release.py
```diff
@@ -129,11 +130,19 @@ def get_remote_name(github_user):
...
     """Tags a release."""
-    message = "\n".join(commits_since_tag)
-
     if bug_number:
-        # Add bug number to tag
-        message = message + f"\n\nDeploy bug #{bug_number}"
+        resp = fetch(BZ_BUG_JSON_URL + bug_number, is_json=True)
+        bug_summary = resp["bugs"][0]["summary"]
+
+        input(f">>> Using bug {bug_number}: {bug_summary}. Correct? Ctrl-c to cancel")
```
edugfilho, line 137: When I read a similar message while deploying this morning I admit I paused for a second and hit enter, not 100% sure what I was supposed to do in case I *didn't* want to cancel.
Doesn't prevent me from approving, though

edugfilho (approved):

bin/release.py
willkg: I'm game for different prompting. If you have ideas on what would feel more intuitive (or anything else), write up an issue in the socorro-release issue tracker.

willkg:
Thank you!

PR #2541, opened 2022-04-11
bug 1764069: update django to 3.2.13

edugfilho:


willkg:
Did you update the django version requirement in `requirements.in` and then run `make rebuildreqs` or something different? This picked up a lot of secondary dependency updates and I don't think it should have.

edugfilho:
oh, I ran `make updatereqs` instead. Sorry, I'll fix it.

PR #2542, opened 2022-04-13
update docs

willkg:
This is a pass at cleaning up documentation that's in the repository and was spread out between `README.rst`, `CONTRIBUTING.rst`, and `docs/index.rst`. It also unifies workflow and conventions documentation between Tecken and Socorro/Antenna.


docs/dev.rst
```diff
@@ -37,6 +97,147 @@ The Symbols Service webapp is at: http://localhost:3000
...
+
+
+Code reviews
+------------
+
+Pull requests should be reviewed before merging.
+
+Style nits should be covered by linting as much as possible.
+
+Code reviews should review the changes in the context of the rest of the system.
```
edugfilho, line 162: I get the general idea here but I feel this was added with something more specific in mind. If that's the case, can we have an example?

docs/dev.rst
```diff
@@ -37,6 +97,147 @@ The Symbols Service webapp is at: http://localhost:3000
...
+We use "Rebase and merge" in GitHub.
+
+
+Conventions
+===========
+
+Python code conventions
+-----------------------
+
+All Python code files should have an MPL v2 header at the top::
```
edugfilho, line 184: I wonder if we can offload that to black. I'll see if that's possible and if so, create a bug for it. WDYT?

edugfilho (approved):
Looks great. Thank you for updating this!

docs/dev.rst
edugfilho: ah, same for the JS one

willkg: I think we have a license checker script/tool somewhere, but it's definitely not active or hooked up anywhere.

docs/dev.rst
willkg: If you have ideas for an example, we can add it.

edugfilho: Nevermind: I see this text was present before in the docs and was moved. I thought you had just written it based on a recent situation.

docs/dev.rst
edugfilho: I'll see if I can find it

willkg: I hunted for a bit and then wrote up this bug with findings and next step:

https://bugzilla.mozilla.org/show_bug.cgi?id=1764622

willkg:
I'll wait to land this until I get back.

willkg:
Thank you!

PR #2543, opened 2022-04-14
bug 1764622: license check

willkg:
This adds a `bin/license_check.py` script which checks code files for the license header. Currently, it only checks Python files, but there's some scaffolding for adding support for other languages.

It has a `--file-only` mode which only shows file names. I looked at other commands that have a similar behavior and the argument they use to trigger it. ripgrep and grep use `--files-with-matches`, but that feels awkward here. `git grep` supports both `--files-with-matches` and an alias for that `--name-only`. The latter fits better but "name" felt too general. I thought about `--filename-only` and `--file-only` and then went with the latter since it was shorter and probably good enough.

To test:

1. run `python bin/license_check.py` and it'll go through all the files in source control -- this is the one we'll use in CI
2. run `python bin/license_check.py bin/license_check.py` and it'll check `bin/license_check.py` only
3. run `python bin/license_check.py tecken` and it'll check all the files in the `tecken/` directory tree

The wry twist is that it fails on `bin/release.py`. I need to fix that in another repository and copy it over. I'll do that in a separate PR which we can land and then rebase this one to pick up the fix. Once nice thing is that that'll show it fails in CI.


willkg:
Rrr... CI fails because the lint job runs inside the docker image but we're not mounting the volume because we can't do that in CI so it's not running in the git repository.

I'm going to toss it in its own lint step since that seems easiest and it doesn't use any requirements.

willkg:
That shows what a license check fail looks like. Yay!

```
#!/bin/bash -eo pipefail
python ./bin/license_check.py

File bin/release.py does not have license header.
Files with missing headers: 1

Add this:

This Source Code Form is subject to the terms of the Mozilla Public
License, v. 2.0. If a copy of the MPL was not distributed with this
file, You can obtain one at https://mozilla.org/MPL/2.0/.

Exited with code exit status 1
```

edugfilho (approved):
lgtm, this is really nice!

willkg:
The socorro-release fix is here: https://github.com/willkg/socorro-release/pull/23

willkg:
@edugfilho Thank you!

PR #2557, opened 2022-05-26
bug 1771319: switch from minio to localstack

willkg:
This switches the local dev environment from minio to localstack for emulated AWS S3.

For some background on why, see https://bugzilla.mozilla.org/show_bug.cgi?id=1771193 . The gist of it is that minio doesn't make buckets publicly viewable via HTTP by default and to fix that, we need to do minio-specific things and at that point, it feels like it strays too far from our need for an AWS S3 emulator. On top of that, we use localstack for Antenna and Socorro and switching Tecken to localstack removes yet another difference between the things we maintain.

One interesting thing this shows is how tangled the local dev environment is in production code. That's terrible. There have to be better ways to do things such that the local dev environment details aren't littered across the codebase with a bunch of edge cases. This PR doesn't fix that. I think we need to rework the symbol source code.

To test:

1. run `make build` and `make setup`
2. run systemtests


edugfilho (approved):
looks great. Thanks for working on this! It improves local development and aligns tecken with socorro and antenna on this front

willkg:
Thank you!

PR #2558, opened 2022-05-26
bug 1758566: fix cache invalidation for try symbols

willkg:
When try symbols get uploaded, the "does this exist in s3?" cache wasn't
getting invalidated, so symbol download could continue to kick up 404s.

This fixes that specific issue by adding the try symbols source to the
downloader hard-coded in the upload code. There's another bug to rewrite
how all this works and we can fix it in a more resilient way then.

To test:

1. log in and create an API token with "upload try symbols files" permission
2. create a zip file to upload
3. `curl --head http://localhost:8000/try/<debugfilename>/<debugid>/<filename>.sym`

   This should return a 404 because the symbol doesn't exist.
5. upload the zip file and make sure the file got uploaded
6. `curl --head http://localhost:8000/try/<debugfilename>/<debugid>/<filename>.sym`

   This should return a 200 now because the symbol does exist and the cache was invalidated.

If you don't have a symbol zip file, you can create a zip to upload using the systemtests script `systemtests/bin/make-symbols-zip.py`. When generating the file, it'll spit out output like this:

```
app@9ad507c4c348:/app/systemtests$ ./bin/make-symbols-zip.py --auth-token 23c38b3b02c948ed86264350200a79d1 --max-size=5000000 data/zip-files/
Generating ZIP file data/zip-files/symbols_20220526_180657.zip ...
Adding v1/libxul.so/8D4A6289D19D5094AE3A169FFE398C6C0/libxul.so.sym (22408377) ...
size: 22745935, max_size: 5000000
Completed data/zip-files/symbols_20220526_180657.zip (22745935)!
app@9ad507c4c348:/app/systemtests$
```

The debugfilename for that symbol is "libxul.so" and the debug id is "8D4A6289D19D5094AE3A169FFE398C6C0" so the curl line would be:

```
curl --head http://localhost:8000/try/libxul.so/8D4A6289D19D5094AE3A169FFE398C6C0/libxul.so.sym
```
```

### DIFF
--- a/docker/config/local_dev.env
+++ b/docker/config/local_dev.env
@@ -17,6 +17,8 @@ SECRET_KEY=DontusethisinproductionbutitneedsbelongforCI1234567890
 
 DATABASE_URL=postgresql://postgres:postgres@db/tecken
 
+REDIS_URL=redis://redis-cache:6379/0
+
 SENTRY_DSN=http://public@fakesentry:8090/1
 
 # NOTE(willkg): See docker-compose.yml on how localstack is set up.
--- a/tecken/base/symboldownloader.py
+++ b/tecken/base/symboldownloader.py
@@ -108,7 +108,10 @@ def exists_in_source(source, key):
 @metrics.timer_decorator("symboldownloader_public_exists")
 def check_url_head(url):
     session = session_with_retries()
-    return session.head(url).status_code == 200
+    resp = session.head(url)
+    if resp.status_code not in (200, 404):
+        logger.error(f"check_url_head: {url} status code is {resp.status_code}")
+    return resp.status_code == 200
 
 
 class SymbolDownloader:
--- a/tecken/settings.py
+++ b/tecken/settings.py
@@ -68,6 +68,7 @@ if TOOL_ENV:
         ("OIDC_OP_TOKEN_ENDPOINT", "http://example.com/"),
         ("OIDC_OP_USER_ENDPOINT", "http://example.com/"),
         ("DATABASE_URL", "postgresql://postgres:postgres@db/tecken"),
+        ("REDIS_URL", "redis://redis-cache:6379/0"),
         ("SYMBOL_URLS", "https://example.com/"),
         ("UPLOAD_DEFAULT_URL", "https://example.com/"),
         ("UPLOAD_TRY_SYMBOLS_URL", "https://example.com/try/"),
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: docker/config/local_dev.env [hunk 1]
```diff
@@ -17,6 +17,8 @@ SECRET_KEY=DontusethisinproductionbutitneedsbelongforCI1234567890
 
 DATABASE_URL=postgresql://postgres:postgres@db/tecken
 
+REDIS_URL=redis://redis-cache:6379/0
```
willkg, line 20: I forgot this got into this PR. I can probably extract it out into a separate PR, but what's going on here is that the local dev environment redis url was being used as a default value in the configuration and that's not how we want to set things up. The defaults should be good for a server environment.

Given that, I pulled this value out into this `local_dev.env` file and removed the default. Not having it set is a server configuration error and should show up as such.

tecken/base/symboldownloader.py
```diff
@@ -108,7 +108,10 @@ def exists_in_source(source, key):
 @metrics.timer_decorator("symboldownloader_public_exists")
 def check_url_head(url):
     session = session_with_retries()
-    return session.head(url).status_code == 200
+    resp = session.head(url)
+    if resp.status_code not in (200, 404):
+        logger.error(f"check_url_head: {url} status code is {resp.status_code}")
+    return resp.status_code == 200
```
willkg, line 114: This was a bug and made debugging with minio really hard because minio would return a 403 which is not 200. Previously, a 403 would just cause this to return False, but 403 is an error scenario (500, too) and we should log an error.

tecken/settings.py
```diff
@@ -68,6 +68,7 @@ def dict_parser(val):
         ("OIDC_OP_TOKEN_ENDPOINT", "http://example.com/"),
         ("OIDC_OP_USER_ENDPOINT", "http://example.com/"),
         ("DATABASE_URL", "postgresql://postgres:postgres@db/tecken"),
+        ("REDIS_URL", "redis://redis-cache:6379/0"),
```
willkg, line 71: This set of settings is gross, but it's the easiest way to make sure `./manage.py collectstatic` works when building the container in a context where there is no configuration.

Because I removed the default for `REDIS_URL`, we need to add this line here.

tecken/upload/utils.py
```diff
@@ -25,7 +25,9 @@
 logger = logging.getLogger("tecken")
 metrics = markus.get_metrics("tecken")
 
-downloader = SymbolDownloader(settings.SYMBOL_URLS)
+# FIXME(willkg): This downloader needs to point to all the things that can be downloaded
+# from so it can correctly invalidate the cache
+downloader = SymbolDownloader(settings.SYMBOL_URLS + [settings.UPLOAD_TRY_SYMBOLS_URL])
```
willkg, line 30: This is lousy because we're defining multiple SymbolDownloaders that can get out of sync which was the crux of the problem here.

When we rewrite the storage interface ([1564452](https://bugzilla.mozilla.org/show_bug.cgi?id=1564452)), we will fix this.

edugfilho (approved):
lgtm. Thanks for covering the different parts of this implementation, made it a lot easier to review!

willkg:
Thank you for reviewing!

PR #2568, opened 2022-06-09
bug 1764569: implement sentry scrubbing

willkg:
sentry_sdk doesn't scrub sentry events before sending them to the Sentry server. This adds libsentry which has a scrubbing system as well as a `set_up_sentry` more tailored to our uses which we'll share with other socorro-eng projects.

This sets up the Boto3, Redis, and Django integrations as well as the default ones sentry_sdk sets up.

This makes some initial decisions on what to scrub. Generally, it tries to scrub anything related to auth tokens, the user's session id, and user oauth state.

To test:

1. run the server `make run`
2. go to `http://localhost:8000/__broken__` to trigger an exception
3. view the exception in `http://localhost:8090/`
4. run the tests (`make test`)

This is a draft for now. I want to do another review pass and also mull a bit on the `scrub_keys` argument. I'll implement scrubbing for Eliot and it's disk cache manager process in a follow-up PR.

### DIFF
--- a/tecken/apps.py
+++ b/tecken/apps.py
@@ -17,9 +17,9 @@ from django.apps import AppConfig
 from tecken.libdockerflow import get_release_name
 from tecken.libsentry import (
     build_scrub_query_string,
-    scrub,
     Scrubber,
-    SCRUB_KEYS_DEFAULT,
+    SCRUB_RULES_DEFAULT,
+    ScrubRule,
     set_up_sentry,
 )
 
@@ -27,19 +27,31 @@ from tecken.libsentry import (
 logger = logging.getLogger("django")
 
 
-SCRUB_KEYS_TECKEN = [
+SCRUB_RULES_TECKEN = [
     # HTTP request bits
-    (
-        "request.headers",
-        ("Auth-Token", "Cookie", "X-Forwarded-For", "X-Real-Ip"),
-        scrub,
+    ScrubRule(
+        key_path="request.headers",
+        keys=["Auth-Token", "Cookie", "X-Forwarded-For", "X-Real-Ip"],
+        scrub_function="scrub",
     ),
-    ("request.data", ("csrfmiddlewaretoken", "client_secret"), scrub),
-    ("request", ("query_string",), build_scrub_query_string(params=["code", "state"])),
-    ("request", ("cookies",), scrub),
+    ScrubRule(
+        key_path="request.data",
+        keys=["csrfmiddlewaretoken", "client_secret"],
+        scrub_function="scrub",
+    ),
+    ScrubRule(
+        key_path="request",
+        keys=["query_string"],
+        scrub_function=build_scrub_query_string(params=["code", "state"]),
+    ),
+    ScrubRule(key_path="request", keys=["cookies"], scrub_function="scrub"),
     # "request" shows up in exceptions as a repr which in Django includes the
     # query_string, so best to scrub it
-    ("exception.values.[].stacktrace.frames.[].vars", ("request",), scrub),
+    ScrubRule(
+        key_path="exception.values.[].stacktrace.frames.[].vars",
+        keys=["request"],
+        scrub_function="scrub",
+    ),
 ]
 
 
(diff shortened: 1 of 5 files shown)


tecken/apps.py
```diff
@@ -6,17 +6,43 @@
...
+    set_up_sentry,
+)
+
 
 logger = logging.getLogger("django")
 
 
+SCRUB_KEYS_TECKEN = [
+    # HTTP request bits
+    (
```
willkg, line 32: I'm not thrilled about how this is defined. I tossed around some other possibilities that weren't parens heavy, but they seemed overly verbose.

The dataclass approach would be something like this:

```
SCRUB_KEYS_TECKEN = [
    ScrubRule(
        parent="request.headers",
        keys=("Auth-Token", "Cookie", "X-Forwarded-For", "X-Real-Ip"),
        actions=scrub
    )
    ...
]
```

That's much clearer. We really want clarity here, so maybe that's the right thing to do.

tecken/libsentry.py
```diff
@@ -0,0 +1,400 @@
...
+SENTRY_MODULE_NAME = __name__
+metrics = markus.get_metrics(SENTRY_MODULE_NAME)
+logger = logging.getLogger(SENTRY_MODULE_NAME)
+
+
+MASK_TEXT = "[Scrubbed]"
+
+
+ALL_COOKIE_KEYS = object()
+ALL_QUERY_STRING_KEYS = object()
```
willkg, line 23: This is a nice way to do constants that you can compare cheaply using Python `is` because there's only one of them and `is` does equality by memory address.

tecken/libsentry.py
```diff
@@ -0,0 +1,400 @@
...
+SCRUB_KEYS_DEFAULT = [
+    # Hide stacktrace variables
+    ("exception.values.[].stacktrace.frames.[].vars", ("username", "password"), scrub),
+]
+
+
+def get_target_dicts(event, key_path):
+    """Given a key_path, yields the target dicts.
+
+    Uses a dotted path of keys. To traverse arrays, use `[]`.
```
willkg, line 190: This isn't correct.

tecken/libsentry.py
```diff
@@ -0,0 +1,400 @@
...
+                            filtered_val = scrub_fun(val)
+                        except Exception:
+                            logger.exception(f"LIBSENTRYERROR: Error in {scrub_fun}")
+                            metrics.incr("scrub_fun_error")
+                            filtered_val = "ERROR WHEN SCRUBBING"
+
+                        parent[key] = filtered_val
+            except Exception:
+                logger.exception("LIBSENTRYERROR: Error in get_target_dicts")
+                metrics.incr("get_target_dicts_error")
```
willkg, line 288: If we wanted to make this into a library we published on PyPI, we'd remove markus usage here (not everyone uses markus) and replace it with a hook for reporting an error with the scrubber.

tecken/apps.py
```diff
@@ -6,17 +6,55 @@
...
+    ScrubRule(
+        key_path="request",
+        keys=["query_string"],
+        scrub_function=build_scrub_query_string(params=["code", "state"]),
+    ),
+    ScrubRule(key_path="request", keys=["cookies"], scrub_function="scrub"),
+    # "request" shows up in exceptions as a repr which in Django includes the
+    # query_string, so best to scrub it
+    ScrubRule(
+        key_path="exception.values.[].stacktrace.frames.[].vars",
```
willkg, line 51: These strings can get long and there's nothing that ensures they're correct. I'm not really sure what to do about that. I think I'm going to leave it for now.

_[author begins to lame lack of schema and then on a lark, does a web search ...]_

Turns out there's a schema that's used to verify the Python sentry sdk:

https://github.com/getsentry/sentry-data-schemas/blob/main/relay/event.schema.json

We can use that to verify these in a test.

willkg: I thought about this some more. I think it's a doable idea, but I'm not going to do it in this PR.

willkg:
@edugfilho This is still a draft because I'm waffling about whether to turn the scrubbing portion into a library. If I did that, then this PR would change significantly because the bulk of it would be a library. I'll decide one way or the other next week.

In the meantime, what do you think of this?

edugfilho:
In general I think turning it into a library makes sense. If I were to do it myself, because I don't have as much experience as you, I'd probably want to mature this in socorro before. But I think that doesn't apply to you as much because you're able to tell whether something is good to become a lib. 

- What are you're considering in terms of pros an cons before deciding?
- Is it a concern that we'll left something behind?
   - If so, would it be possible to scrub a whole path and, instead of specifying keys to scrub, declare keys to include? That may make it safer if paths have a high correlation with data we want to scrub

willkg:
I'm going to close this out. I'm extracting the relevant code into a library and we can talk about library API bits there. Then we can redo this PR and it'll be a lot shorter since it won't include library bits.

PR #2571, opened 2022-06-14
bug 1774203: document uploading to stage

willkg:
This adds some documentation to the upload API encouraging people to use
the stage environment for testing symbol uploads and walks through how
to do that.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -96,7 +96,7 @@ release = "1.0"
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = "en"
 
 # There are two options for replacing |today|: either, you set today to some
 # non-false value, then it is used:
--- a/docs/upload.rst
+++ b/docs/upload.rst
@@ -13,9 +13,9 @@ Uploading symbols
 Basics
 ======
 
-Tecken lets you upload symbols. It stores these symbols files allowing others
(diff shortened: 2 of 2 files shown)


willkg:
@edugfilho Can you skim this and see if it makes sense to you? The gist of it is that we have two environments and we want to document enough of the stage environment and how to use it that people who are testing uploads and testing scripts have enough information here to understand what we support and do what they need to do.

docs/conf.py
```diff
@@ -96,7 +96,7 @@
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = "en"
```
willkg, line 99: Sphinx 5 raises a warning with our previous value.

docs/upload.rst
```diff
@@ -73,17 +74,6 @@ For example, here's the contents of a symbols ZIP file from a Firefox build::
...
 
-Permissions and auth token
---------------------------
-
-Uploading requires two things:
-
-1. an account with upload permissions
-2. an auth token for that account with upload permissions
-
-The auth token is used in the HTTP POST to authenticate the upload.
```
willkg, line 84: This part was in this document twice, so I nixed it from here figuring the other section is more fleshed out.

edugfilho (approved):
looks great!

willkg:
Thank you!

PR #2572, opened 2022-06-23
bug 1764569: implement sentry scrubbing

willkg:
This adds Fillmore to the dependency list and implements Sentry event scrubbing for Tecken webapp, Eliot webapp, and Eliot disk cache manager.

This adds integration tests for all three of those things so we know if anything changes allowing us to update our scrubbing configuration.

This makes some initial decisions on what to scrub. Generally, it tries to scrub anything related to auth tokens, the user's session id, and user oauth state.

Fillmore: https://github.com/willkg/fillmore


eliot-service/eliot/app.py
```diff
@@ -39,6 +42,15 @@
 STATICROOT_DIR = str(Path(__file__).parent / "static")
 
 
+SCRUB_RULES_ELIOT = [
+    Rule(
+        path="request.headers",
+        keys=["X-Forwarded-For", "X-Real-Ip"],
+        scrub="scrub",
+    ),
+]
```
willkg, line 51: Eliot has no user accounts, so the only sensitive data it's got is infrastructure-related.

eliot-service/eliot/app.py
```diff
@@ -254,10 +266,13 @@ def get_app(config_manager=None):
...
     )
-    setup_sentry(
-        basedir=str(REPOROOT_DIR),
-        host_id=app_config("host_id"),
+
+    set_up_sentry(
         sentry_dsn=app_config("secret_sentry_dsn"),
+        release=get_release_name(REPOROOT_DIR),
+        host_id=app_config("host_id"),
+        integrations=[FalconIntegration()],
```
willkg, line 274: There are some issues in the sentry_sdk issue tracker about how the `FalconIntegration` is broken. However, we can see with our integration test that it's definitely doing something, so it seems fine.

eliot-service/eliot/liblogging.py
```diff
@@ -109,4 +109,4 @@ def log_config(logger, config_manager, component):
             value = "*****"
 
         full_key = generate_uppercase_key(key, ns).upper()
-        logger.info(f"{full_key}={value}")
+        logger.debug(f"{full_key}={value}")
```
willkg, line 112: This changes the settings logging to be at the DEBUG level. We don't need to log the settings at INFO level every time things reboot.

eliot-service/eliot/libsentry.py
```diff
@@ -1,75 +0,0 @@
...
-        LOGGER.warning("no sentry_dsn set")
-        return
-
-    release = get_release(basedir)
-
-    sentry_logging = LoggingIntegration(level=logging.INFO, event_level=logging.ERROR)
-
-    sentry_sdk.init(
-        dsn=sentry_dsn,
-        integrations=[sentry_logging],
```
willkg, line 70: It's interesting to point out that the default integrations include the `LoggingIntegration`, so I'm not entirely sure what this did. Does it include the `LoggingIntegration` twice?

eliot-service/tests/test_cache_manager.py
```diff
@@ -2,7 +2,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+import json
 import pathlib
+from unittest.mock import ANY
```
willkg, line 7: `ANY` is a magic amazing thing. Prior to discovering it existed, I wrote a `WHATEVER` that was kind of the same thing, but funnier to read.

tecken/apps.py
```diff
@@ -6,17 +6,55 @@
...
+        path="request.data",
+        keys=["csrfmiddlewaretoken", "client_secret"],
+        scrub="scrub",
+    ),
+    Rule(
+        path="request",
+        keys=["query_string"],
+        scrub=build_scrub_query_string(params=["code", "state"]),
+    ),
+    Rule(path="request", keys=["cookies"], scrub="scrub"),
```
willkg, line 47: Because we're setting `send_default_pii=False`, there isn't any cookie information. So we probably don't really need this. Further, the cookie information is repeated `request.headers.Cookie` and this scrubbing doesn't account for that which is fine because it's not there anyhow on account of setting `send_default_pii=False`.

I'll remove this line.

tecken/tests/test_sentry.py
```diff
@@ -0,0 +1,229 @@
...
+
+    * update sentry_sdk to a new version
+    * update Django to a new version that somehow adjusts the callstack for an
+      exception happening in view code
+
+    In those cases, we should copy the new event, read through it for new problems, and
+    redact the parts that will change using ANY so it passes tests.
+
+    """
+    client = Client(application)
```
willkg, line 198: Here we use the Werkzeug test client instead of the Django one. The Django one is very very fake and exceptions don't trigger Sentry event generation. Using the Werkzeug one treats the whole thing like a regular WSGI application and the "faking" happens as part of the WSGI API so the Sentry events to get generated and everything works splendid.

I spent some time trying to get this working with LiveServerTestCase. That works by spinning off a thread and running a live Django application there that you can connect to using HTTP clients. It does allow Sentry to generate events, but I could only get it to work sometimes for reasons I'm completely unclear on. I suspect there's something about threads and sentry client initialization and the order in which some things happen.

willkg:
@edugfilho I didn't implement the include/exclude keys thing, but I did toss that into an issue to think about at some point. The integration tests help us know if other things get added to the event that we weren't aware of, so not having an include-keys seems ok for now.

The code for fillmore is here: https://github.com/willkg/fillmore

What do you think of the whole thing?

willkg:
Oh, whoops--I'm missing one thing. I need to add an error_handler that emits metrics via markus. I'll do that tomorrow.

willkg:
^^^ That adds the missing emit a metric on scrub error bit.

eliot-service/eliot/liblogging.py
edugfilho: the way this is phrased implies DEBUG has less information than INFO, but if I'm not mistaken DEBUG contains INFO, so I'm confused.

eliot-service/tests/test_cache_manager.py
```diff
@@ -403,3 +410,166 @@ def test_nested_directories(cm_client, tmpdir):
 
     cm.run_once()
     assert cm.lru == {str(file2): 4, str(file3): 2}
+
+
+# NOTE(willkg): If this changes, we should update it and look for new things that should
+# be scrubbed. Use ANY for things that change between tests.
+BROKEN_EVENT = {
```
edugfilho, line 417: I wonder if in the future we can write something to make testing more straightforward, like being able to declare a template for this to be used in the different tests across the app.

edugfilho (approved):
Interesting to see how to add Fillmore to a project!

eliot-service/eliot/liblogging.py
willkg: I don't understand what you mean by that.

With Python logging, you can log at different levels. Then the application sets the level below which logging calls are ignored. Our server set up is at INFO, but local dev environment is set up at DEBUG. So by changing this from INFO to DEBUG, we will continue to see these lines (which spit out all the configuration) in the logs in a local dev environment, but not in a server environment.

edugfilho: yes it does. My bad, it was my misinterpretation.

willkg:
I want to land this today so we have a bunch of days to let it run. We rarely get errors in the stage environment, so I'll have to spend some time breaking things so as to see how the data ends up in Sentry. Having a few days to make changes makes that easier.

willkg:
Thank you!

PR #2598, opened 2022-08-20
bug 1784290: accept dashes in the ID, and compress and serve files named "debuginfo" and "executable"

mstange:
These are the changes for https://bugzilla.mozilla.org/show_bug.cgi?id=1784290.

At the moment these patches fail tests, and I haven't had a chance to debug those failures.
I'm pushing this branch now in case someone else wants to take a look while I'm on PTO.


biancadanforth:
@mstange This is almost four years old. Are you planning to pick this up again any time soon? Else can it be closed?

mstange:
I'm happy to close this for now until I or somebody else find time to pick this back up. Once we have a samply-based symbolication server, we'll be able to serve assembly from uploaded binaries for the profiler's assembly view, and this will make this a higher priority because then we'll be just missing assembly for Mac/Linux/Android builds.

PR #2621, opened 2022-10-19
bug 1796264: change upload_upload.size field to bigint

willkg:
We hit the maximum value this field can store, so we need to change the field to a bigint.

The sql this is going to run is this:

```
BEGIN;
--
-- Alter field size on upload
--
ALTER TABLE "upload_upload" ALTER COLUMN "size" TYPE bigint USING "size"::bigint;
COMMIT;
```

That table has < 300k rows in production. I'm pretty sure running this will cause an outage, but I have no idea how long it'll take. We should test this before landing.


willkg:
I talked this over with @haroldwoo and our collective gut feeling is we should go with this migration. CI tests that the db table change works with the application. I did some testing on my local machine.

I'm going to upload a bunch of symbols to stage to add additional rows to its table (currently 173).

Then when we land this PR, it'll run the migration on stage, we can see how long it took, and we can verify that the db table change works with the site.

haroldwoo (approved):

PR #2628, opened 2022-10-28
bug 1793984: clean up symbolication code

willkg:
This moves some symbolication code around to reduce duplication between the v4 and v5 views and also to change the work of the `symbolicate` method so it's operating on a list of jobs rather than a single job. This sets us up to rewrite `symbolicate` for more efficient processing.

We switched stackwalkers at the end of 2021 and the new one emits nulls rather than omitting the field altogether. This updates `make-stacks.py` to account for nulls better.

### DIFF
--- a/eliot-service/eliot/symbolicate_resource.py
+++ b/eliot-service/eliot/symbolicate_resource.py
@@ -380,137 +380,143 @@ class SymbolicateBase:
 
         module_info.has_symcache = module_info.symcache is not None
 
-    def symbolicate(self, stacks, modules, debug_stats):
-        """Takes stacks and modules and returns symbolicated stacks.
+    def symbolicate(self, jobs, debug_stats):
+        """Takes jobs and returns symbolicated results.
 
-        :arg stacks: list of stacks each of which is a list of
-            (module index, module offset)
-        :arg modules: list of (debug_filename, debug_id)
+        :arg jobs: list of jobs containing stack and module information
         :arg debug_stats: DebugStats instance for keeping track of timings and other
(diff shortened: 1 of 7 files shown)


eliot-service/eliot/symbolicate_resource.py
```diff
@@ -380,137 +380,143 @@ def get_symcache(self, module_info, debug_stats):
...
+            module_records = [
+                ModuleInfo(
+                    filename=debug_filename,
+                    debug_filename=debug_filename,
+                    debug_id=debug_id,
+                    has_symcache=None,
+                    symcache=None,
+                )
+                for debug_filename, debug_id in modules
+            ]
```
willkg, line 411: To reduce work acquiring symcache files, we should bump this outside of the `for` loop. However, right now we're having memory errors so having fewer modules in memory at the same time is better. However, however, we're not seeing any multi-job symbolication requests right now so it doesn't matter.

mstange: > However, however, we're not seeing any multi-job symbolication requests right now so it doesn't matter.

The profiler uses multi-job requests, but it only requests a single module per job.
https://github.com/firefox-devtools/profiler/blob/b5f7ff53dba06e4918eafec88a2d17bb41239bfc/src/profile-logic/mozilla-symbolication-api.js#L276-L287

willkg: Huh. My metrics show no indication of multiple-jobs-per-request. I'll have to look into that.

PR #2637, opened 2022-11-02
bug 1793984: add max requests and jitter

willkg:
This changes the gunicorn configuration to use max requests at 10,000 and jitter at 1,000 so we have fresh workers.


bin/run_eliot_web.sh
```diff
@@ -9,6 +9,8 @@
 : "${ELIOT_GUNICORN_WORKERS:=1}"
 : "${ELIOT_GUNICORN_TIMEOUT:=300}"
 : "${ELIOT_GUNICORN_CMD_PREFIX:=}"
+: "${ELIOT_GUNICORN_MAX_REQUESTS:=10000}"
+: "${ELIOT_GUNICORN_MAX_REQUESTS_JITTER:=1000}"
```
willkg, line 13: These are the same values we use in the Socorro collector (antenna). We can tune it going forward.

PR #2647, opened 2022-11-17
bug 1793984: reorder frames to symbolicate

willkg:
This reworks symbolication to reorder all the frames to symbolicate by module so that as we're symbolicating frames, we only need one module in memory at a time.

This also removes the need for using a temporary directory to store a symbol file that was just downloaded in order to parse it with symbolic. Now we parse it straight from bytes.

I tested this in a few ways:

1. updated unit tests
2. run system tests
3. pull down crash reports to build stacks to symbolicate and then compare the results from prod with the results of my local dev environment
4. run a "memory test" where I measure the memory usage of the container running the disk cache manager, gunicorn, and a single eliot gunicorn worker symbolicating multiple different requests

It looks ok.

### DIFF
--- a/eliot-service/eliot/libsymbolic.py
+++ b/eliot-service/eliot/libsymbolic.py
@@ -8,8 +8,6 @@ Utilities for using symbolic library.
 
 from io import BytesIO
 import logging
-import os
-import tempfile
 
 import symbolic
 
@@ -103,13 +101,12 @@ def convert_debug_id(debug_id):
         raise BadDebugIDError("invalid_identifier")
 
 
-def parse_sym_file(debug_filename, debug_id, data, tmpdir):
+def parse_sym_file(debug_filename, debug_id, data):
     """Convert sym file to symcache file
 
     :arg debug_filename: the debug filename
     :arg debug_id: the debug id
     :arg data: bytes
-    :arg tmpdir: the temp directory to use
 
     :returns: symcache or None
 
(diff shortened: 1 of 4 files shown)


eliot-service/eliot/libsymbolic.py
```diff
@@ -103,13 +101,12 @@ def convert_debug_id(debug_id):
         raise BadDebugIDError("invalid_identifier")
 
 
-def parse_sym_file(debug_filename, debug_id, data, tmpdir):
```
willkg, line 106: When Eliot downloads a sym file from symbols.mozilla.org, it has the file in memory and now parses it from memory. Previously, it would have it in memory, save it to disk (which caused other problems), parse it from disk, and then delete the file from disk. But, it's doing this while the file is still in memory, so there's no memory savings to saving it to disk and saving it and loading it from disk probably slowed things down (I never measured this--it would have been really hard to get good measurements).

Anyhow, this removes the whole save it to disk / parse it from disk / remove the temporary file thing.

eliot-service/eliot/symbolicate_resource.py
```diff
@@ -295,25 +280,26 @@ def parse_sym_file(self, debug_filename, debug_id, data):
                 tags=[f"reason:{psfe.reason_code}"],
             )
 
-    def get_symcache(self, module_info, debug_stats):
+    def get_symcache(self, debug_filename, debug_id, debug_stats):
         """Gets the symcache for a given module.
 
         This uses the cachemanager and downloader to get the symcache. It modifies
         the Module in place.
```
willkg, line 287: Oops--this should be updated.

eliot-service/eliot/symbolicate_resource.py
```diff
@@ -295,25 +280,26 @@ def parse_sym_file(self, debug_filename, debug_id, data):
...
-        :arg module_info: the ModuleInfo data class
+        :arg debug_filename: the debug filename
+        :arg debug_id: the debug id
         :arg debug_stats: DebugStats instance for keeping track of timings and other
             useful things
 
-        """
-        debug_filename = module_info.debug_filename
-        debug_id = module_info.debug_id
+        :returns: symcache, filename or None
```
willkg, line 294: I hate this return, but I didn't have a better way to deal with this well without creating data structures.

eliot-service/eliot/symbolicate_resource.py
```diff
@@ -392,114 +376,148 @@ def symbolicate(self, jobs, debug_stats):
...
                     module_index, module_offset = frame
-                    module_info = None
-                    data = {
-                        "frame": frame_index,
-                        "module": "<unknown>",
+
+                    if 0 <= module_index < len(memorymap):
+                        debug_filename, debug_id = memorymap[module_index]
+                    else:
+                        debug_filename, debug_id = None, None
```
willkg, line 406: I should switch this to use empty strings because those are falsey and the type of these variables is consistent.

PR #2700, opened 2023-03-08
bug 1821063: fix temporary directory usage

willkg:
This drops `UPLOAD_TEMPDIR_PREFIX` for `UPLOAD_TEMPDIR` which is an absolute path to a directory to use for building temporary workspaces for handling symbol uploads.

While doing this, I moved the `make_tempdir` decorator closer to the only place it's used in the codebase.

This requires a configuration change in the infrastructure code before landing.


willkg:
The infrastructure changes landed, so I'm going to land this and verify it on stage.

PR #2701, opened 2023-03-10
bug 1817902: add orphaned files removal

willkg:
When a gunicorn worker is handling a symbol upload request and it takes too long, gunicorn will kill the worker off orphaning the workspace the worker was using. This causes the disk to fill up and then the node to fail.

This fixes that by adding a "disk manager" process that wakes up periodically to remove files that have been around too long and have likely been orphaned.

### DIFF
--- a/tecken/settings.py
+++ b/tecken/settings.py
@@ -542,6 +542,15 @@ UPLOAD_TEMPDIR = _config(
     default="/tmp/uploads",
     doc="The directory to use as a workspace for handling symbol uploads.",
 )
+UPLOAD_TEMPDIR_ORPHANS_CUTOFF = _config(
+    "UPLOAD_TEMPDIR_ORPHANS_CUTOFF",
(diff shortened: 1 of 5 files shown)


tecken/settings.py
```diff
@@ -542,6 +542,15 @@ def dict_parser(val):
     default="/tmp/uploads",
     doc="The directory to use as a workspace for handling symbol uploads.",
 )
+UPLOAD_TEMPDIR_ORPHANS_CUTOFF = _config(
+    "UPLOAD_TEMPDIR_ORPHANS_CUTOFF",
+    default="60",
```
willkg, line 547: It might be interesting to add parsers to Everett for handling things like "60m" or "5m30s" where it converts a time value to seconds. It could do the validation and parsing. It'd improve the readability of the configuration value. I'll write up an Everett feature for that.

PR #2716, opened 2023-04-18
bug 1828542: fix debug timings for symbolication

willkg:
This fixes the debug timings for symbolication to split up "download", "parse", and "save" into separate timings. This will make it easier to discover performance issues.

### DIFF
--- a/eliot-service/eliot/symbolicate_resource.py
+++ b/eliot-service/eliot/symbolicate_resource.py
@@ -290,10 +290,12 @@ class SymbolicateBase:
         :arg debug_stats: DebugStats instance for keeping track of timings and other
             useful things
 
-        :returns: symcache, filename or None
+        :returns: ``(symcache, filename)`` or ``None``
 
         """
         if not debug_filename or not debug_id:
+            # If we don't have a debug_filename or debug_id, there's nothing we can
+            # download, so return None
             return
 
         symcache = None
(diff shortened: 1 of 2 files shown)


eliot-service/eliot/symbolicate_resource.py
```diff
@@ -320,49 +322,96 @@ def get_symcache(self, debug_filename, debug_id, debug_stats):
         end_time = time.perf_counter()
         debug_stats.incr("cache_lookups.time", end_time - start_time)
 
-        # We didn't find it in the cache, so try to download it
```
willkg, line 323: This is one big difference section, but what's going on is that I added some additional timing blocks and I redid the code structure so it returns early when possible with the correct return value. That dedents a lot of this and makes it easier to maintain going forward.

eliot-service/eliot/symbolicate_resource.py
```diff
@@ -320,49 +322,96 @@ def get_symcache(self, debug_filename, debug_id, debug_stats):
...
+
+        debug_stats.incr("downloads.count", 1)
+
+        if sym_file is None:
+            # Nothing was downloaded, so capture the timing as a download fail and
+            # return None
+            debug_stats.incr(
+                [
+                    "downloads",
+                    "fail_time_per_module",
```
willkg, line 341: I waffled about how to represent this. We want to know timings for success _and_ failure--maybe the failure case took a really long time or something like that. I wanted to keep things in sections by action. I ended up with this.

These are debug stats, so we can always change them later.

PR #2749, opened 2023-06-21
bug 1728859: remove UploadsCreated model and related code

willkg:
This removes the UploadsCreate model, the code that populates that data (which was in the upload view, so this should speed that up a bit), the code that uses that data (which was in the front page, which for logged in users who have view symbols permissions, might slow the front page down), and the undocumented APIs that exposed the data.


tecken/api/views.py
```diff
@@ -711,65 +594,33 @@ def stats(request):
     start_yesterday = start_today - datetime.timedelta(days=1)
     last_30_days = today - datetime.timedelta(days=30)
 
-    if not all_uploads:
```
willkg, line 714: If you had the `can_view_all` permission (which only a handful of us do), then you get the summarized data from the UploadsCreated model. Everyone else who's logged in gets the raw data from the tables. The data from the UploadsCreated model is probably faster to retrieve because it's summarized, but it's only available to a handful of people.

Instead, I'm nixing it and forcing the code to calculate the stats from the raw data for everyone. Counts are problematic, so if this slows down the front page for everyone, then I'll write up a new bug and look at either removing the stats information altogether (who even cares about it?) or cache the answers or something.

tecken/api/views.py
```diff
@@ -836,39 +687,6 @@ def count_missing(start, end, use_cache=True):
     return http.JsonResponse(context)
 
 
-@api_login_required
-def stats_uploads(request):
```
willkg, line 840: This undocumented api view wasn't used anywhere.

tecken/tests/test_api.py
```diff
@@ -1093,19 +957,59 @@ def test_stats(client):
     response = client.get(url)
     assert response.status_code == 200
     data = response.json()
-    assert data["stats"]["uploads"]
-    assert not data["stats"]["uploads"]["all_uploads"]
-    assert "users" not in data["stats"]
-    assert data["stats"]["tokens"]
+    assert data["stats"] == {
```
willkg, line 960: I adjusted this test so it's clearer wtf `data["stats"]` has in it in the different modes.

PR #2753, opened 2023-06-28
bug 1814436: default uploads and files pages to last 30 days

willkg:
This changes the uploads and files pages to default to looking at only the last 30 days. This radically reduces how much time it takes to render the pages and gets us around the "omg--it's so slow" problem.

### DIFF
--- a/tecken/api/views.py
+++ b/tecken/api/views.py
@@ -522,6 +522,8 @@ def _upload_files_content(request, qs):
 @api_permission_required("upload.view_all_uploads")
 def upload_files(request):
     qs = _upload_files_build_qs(request)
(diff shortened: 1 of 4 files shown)


tecken/api/views.py
```diff
@@ -522,6 +522,8 @@ def hydrate_upload(upload_id):
 @api_permission_required("upload.view_all_uploads")
 def upload_files(request):
     qs = _upload_files_build_qs(request)
+    if isinstance(qs, http.JsonResponse):
+        return qs
```
willkg, line 526: `_upload_files_build_qs` can return a `JsonResponse` if the field values are invalid. This fixes handling for that case.

PR #2754, opened 2023-06-29
bug 1840655: reduce home page stats

willkg:
This does a few things to reduce the work done by the stats view which is called by the home page:

1. removes the "missing symbols" stats--we're removing missing symbols bookkeeping, so we can remove this now
2. removes the uploaded files counts--I'm not sure this really means anything and comparing two uploaded files counts numbers doesn't really tell us anything
3. removes some other things that were generated by the stats view but weren't used in the frontend--this was extra work being done with no value

### DIFF
--- a/frontend/src/Home.js
+++ b/frontend/src/Home.js
@@ -71,27 +71,21 @@ class SignedInTiles extends React.PureComponent {
     return (
       <div>
         <div className="tile is-ancestor">
-          <div
-            className={user.is_superuser ? "tile is-parent" : "tile is-parent"}
-          >
+          <div className="tile is-parent is-6">
             {store.hasPermission("upload.upload_symbols") ? (
               <UploadsStatsTile loading={loading} stats={stats} />
             ) : (
               <AboutUploadsPermissionTile />
             )}
           </div>
-          <div className="tile is-parent">
-            <DownloadsStatsTile loading={loading} stats={stats} />
+          <div className="tile is-parent is-6">
+            <LinksTile />
           </div>
         </div>
 
         <div className="tile is-ancestor">
-          <div className="tile is-parent is-6">
-            <YouTile user={user} />
-          </div>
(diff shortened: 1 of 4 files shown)


frontend/src/Home.js
```diff
@@ -71,27 +71,21 @@ class SignedInTiles extends React.PureComponent {
     return (
       <div>
         <div className="tile is-ancestor">
-          <div
-            className={user.is_superuser ? "tile is-parent" : "tile is-parent"}
```
willkg, line 75: This isn't doing anything now, so I nixed the conditional.

tecken/api/views.py
```diff
@@ -598,92 +597,31 @@ def stats(request):
...
-            "downloads"
-        ]["missing"]["today"]["count"]
-
-    with metrics.timer("api_stats", tags=["section:your_tokens"]):
-        # Gather some numbers about tokens
-        tokens_qs = Token.objects.filter(user=request.user)
-        numbers["tokens"] = {
-            "total": tokens_qs.count(),
-            "expired": tokens_qs.filter(expires_at__lt=today).count(),
-        }
```
willkg, line 675: This isn't in the frontend anywhere. I didn't look at the logs to figure out when we took it out.

tecken/api/views.py
```diff
@@ -598,92 +597,31 @@ def stats(request):
...
+            if isinstance(value, dict):
+                nones_to_zero(value)
+            elif value is None:
+                obj[key] = 0
 
-    # Gather some numbers about users
-    if request.user.is_superuser:
-        with metrics.timer("api_stats", tags=["section:all_users"]):
-            users_qs = User.objects.all()
-            numbers["users"] = {
```
willkg, line 681: This isn't in the frontend anywhere. I didn't look at the logs to figure out when we took it out.

PR #2758, opened 2023-07-20
Update localstack to 2.2.0

willkg:


docker-compose.yml
```diff
@@ -102,14 +102,11 @@ services:
     image: redis:6.2.6
 
   # https://hub.docker.com/r/localstack/localstack/
-  # localstack running a fake AWS S3 and SQS
+  # localstack running a fake AWS S3
   localstack:
-    image: localstack/localstack:1.1.0
+    image: localstack/localstack:2.2.0
     environment:
-      - SERVICES=s3
```
willkg, line 109: This is deprecated. Services are now started on first use.

docker-compose.yml
```diff
@@ -102,14 +102,11 @@ services:
     image: redis:6.2.6
 
   # https://hub.docker.com/r/localstack/localstack/
-  # localstack running a fake AWS S3 and SQS
```
willkg, line 105: We never used SQS in Tecken. I think I copied this over from Socorro. Oops.

docker-compose.yml
```diff
@@ -102,14 +102,11 @@ services:
...
+  # localstack running a fake AWS S3
   localstack:
-    image: localstack/localstack:1.1.0
+    image: localstack/localstack:2.2.0
     environment:
-      - SERVICES=s3
-      - DEFAULT_REGION=us-east-1
```
willkg, line 110: This is deprecated. It's now multi-region or something like that.

docker-compose.yml
```diff
@@ -102,14 +102,11 @@ services:
...
-      - HOSTNAME=localstack
-      - HOSTNAME_EXTERNAL=localstack
```
willkg, line 112: These are both deprecated.

PR #2759, opened 2023-07-22
bug 1843356, 1834626: remove support for private buckets

willkg:
This is a big PR because it covers several tangled things:

1. removes all the `UPLOAD_URL_EXCEPTIONS` stuff which tied specific accounts to specific private buckets via configuration
2. removes all the private bucket support
3. reworks the tests removing the bulk of the mocking and instead relying on manipulating the localstack fake s3 service the tests run with (bug 1834626)

### DIFF
--- a/docker/config/local_dev.env
+++ b/docker/config/local_dev.env
@@ -46,20 +46,3 @@ OIDC_OP_USER_ENDPOINT=http://oidcprovider:8080/openid/userinfo
 OIDC_VERIFY_SSL=false
 # Disable NotBlockedInAuth0Middleware
 ENABLE_AUTH0_BLOCKED_CHECK=false
-
-
-# Eliot
-# -----
-
-# Local development flag
-ELIOT_LOCAL_DEV_ENV=true
-
-# Logging
-ELIOT_LOGGING_LEVEL=INFO
-
-# Statsd things
-ELIOT_STATSD_HOST=statsd
-ELIOT_STATSD_NAMESPACE=mcboatface
-ELIOT_STATSD_PORT=8125
-ELIOT_SYMBOLS_CACHE_MAX_SIZE=400000000
-ELIOT_SECRET_SENTRY_DSN=http://public@fakesentry:8090/1
--- a/docker/config/local_dev.env
+++ b/docker/config/local_dev.env
@@ -28,9 +28,9 @@ AWS_ENDPOINT_URL=http://localstack:4566/
 
 DEBUG=true
 LOCAL_DEV_ENV=true
-SYMBOL_URLS=http://localstack:4566/publicbucket/?access=public
-UPLOAD_DEFAULT_URL=http://localstack:4566/publicbucket/?access=public
-UPLOAD_TRY_SYMBOLS_URL=http://localstack:4566/publicbucket/try/?access=public
+SYMBOL_URLS=http://localstack:4566/publicbucket/
+UPLOAD_DEFAULT_URL=http://localstack:4566/publicbucket/
+UPLOAD_TRY_SYMBOLS_URL=http://localstack:4566/publicbucket/try/
 
 # Default to the test oidcprovider container for Open ID Connect
 #
--- a/docker/config/test.env
+++ b/docker/config/test.env
@@ -33,14 +33,3 @@ OIDC_OP_USER_ENDPOINT=https://auth.example.com/authorize
 SYMBOL_FILE_PREFIX=v0
 
 UPLOAD_TEMPDIR=/tmp/test/uploads
-
-
-# Eliot settings
-# --------------
-
-ELIOT_HOST_ID=testnode
-
-ELIOT_SYMBOLS_URLS=http://symbols.example.com/
-ELIOT_TMP_DIR=/tmp/test/junk/
-ELIOT_SYMBOLS_CACHE_DIR=/tmp/test/cache/
-ELIOT_SECRET_SENTRY_DSN=http://public@fakesentry:8090/1
--- a/docker/config/test.env
+++ b/docker/config/test.env
@@ -30,6 +30,4 @@ OIDC_RP_CLIENT_SECRET=abcdef
 # This makes sure this is never a real valid URL.
 OIDC_OP_USER_ENDPOINT=https://auth.example.com/authorize
 
-SYMBOL_FILE_PREFIX=v0
-
 UPLOAD_TEMPDIR=/tmp/test/uploads
--- a/tecken/base/symboldownloader.py
+++ b/tecken/base/symboldownloader.py
@@ -18,8 +18,6 @@ from tecken.storage import StorageBucket
 logger = logging.getLogger("tecken")
 metrics = markus.get_metrics("tecken")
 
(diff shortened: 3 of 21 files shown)


docker/config/local_dev.env
```diff
@@ -46,20 +46,3 @@ OIDC_OP_USER_ENDPOINT=http://oidcprovider:8080/openid/userinfo
 OIDC_VERIFY_SSL=false
 # Disable NotBlockedInAuth0Middleware
 ENABLE_AUTH0_BLOCKED_CHECK=false
-
-
-# Eliot
```
willkg, line 51: These are vestigates of when Eliot was in this repo that should have been removed ages ago.

docker/config/test.env
```diff
@@ -30,17 +30,4 @@ OIDC_RP_CLIENT_SECRET=abcdef
 # This makes sure this is never a real valid URL.
 OIDC_OP_USER_ENDPOINT=https://auth.example.com/authorize
 
-SYMBOL_FILE_PREFIX=v0
```
willkg, line 33: The symbol file prefix matches a file structure. It's wrong for the codebase to be using v1, but the tests are using v0. I removed this.

tecken/base/symboldownloader.py
```diff
@@ -31,33 +29,6 @@ def __init__(self, status_code, url):
         self.url = url
 
 
-def iter_lines(stream, chunk_size=ITER_CHUNK_SIZE):
```
willkg, line 34: This was something used by the symbolication code ages ago and is no longer used, so I nixed it.

tecken/tests/test_api.py
```diff
@@ -1177,35 +1177,3 @@ def test_file_upload_try_upload(client):
     assert response.status_code == 200
     data = response.json()["file"]
     assert data["url"] == "/foo.pdb/deadbeaf123/foo.sym?try"
-
-
-@pytest.mark.django_db
-def test_possible_upload_urls(client, settings):
```
willkg, line 1183: We still have this function, but it's a shade of it's former self and will likely get removed when I rewrite the storage layer. Ergo, I nixed this test.

tecken/tests/test_download.py
```diff
@@ -37,164 +37,143 @@ def reload_downloaders(urls, try_downloader=None):
         views.try_downloader = SymbolDownloader([try_downloader])
 
 
-def test_client_happy_path(client, botomock):
-    reload_downloaders("https://s3.example.com/private/prefix/")
+def test_client_happy_path(client, db, s3_helper):
```
willkg, line 40: A lot of these tests relied on the private bucket support which was botomocked out. In order to remove private bucket support, I had to rewrite the download tests to not use botomock.

tecken/tests/test_storage.py
```diff
@@ -165,39 +144,3 @@ def test_storageerror_msg():
         " operation: Forbidden"
     )
     assert str(error) == expected
-
-
-def test_StorageBucket_client():
```
willkg, line 170: I don't understand what this is testing other than the specifics of how it's implemented. I don't think we get a lot of value out of testing implementation details like this, so I nixed it.

tecken/tests/test_upload.py
```diff
@@ -51,21 +46,11 @@ def _join(x):
 DUPLICATED_DIFFERENT_SIZE_ZIP_FILE = _join("duplicated-different-size.zip")
 
 
-class FakeUser:
```
willkg, line 54: This is a really weird class. I dropped this in favor of fixtures that create users in the db with the required properties.

tecken/tests/test_upload.py
```diff
@@ -469,70 +254,55 @@ def mock_api_call(self, operation_name, api_params):
...
-            # Successfully uploaded
-            return {}
-
-        raise NotImplementedError((operation_name, api_params))
+    s3_helper.create_bucket("publicbucket")
+
+    # Upload flag.jpeg from the zip file into the bucket so it's already there
+    # and gets ignored when it's uploaded
+    rootdir = str(tmp_path)
+    with open(ZIP_FILE, "rb") as fp:
```
willkg, line 272: I had to pull the file out of the ZIP_FILE to upload it so that it was already in S3. This is weird. Instead of having pre-built zip files which are opaque and harder to reason about in the tests ("wtf is in _this_ zip file? i have no idea when reading this test."), we should build the zip files with the contents we need them to have. Then we wouldn't have to do this weird thing where we extract the file from the zip file to upload it.

Also, we'd be able to more easily create well-formed files--the debug id in one or more of the zip files is invalid.

tecken/tests/test_upload.py
```diff
@@ -542,314 +312,155 @@ def mock_api_call(self, operation_name, api_params):
...
-        key_existing.invalidate(client, "mybucket", "filename")
-        size, metadata = key_existing(client, "mybucket", "filename")
-        assert size == 6789
-        assert metadata is None
-        assert len(lookups) == 2
+    client = s3_helper.conn
 
+    size, metadata = key_existing(client, "publicbucket", "somefile.txt")
+    assert size == 6
+    assert metadata == {}
```
willkg, line 327: The asserted metadata values for this test were wrong because we were testing against the mirror dimension.

tecken/tests/test_upload.py
```diff
@@ -542,314 +312,155 @@ def mock_api_call(self, operation_name, api_params):
...
 
-    # Upload the same file again. This time some of the S3 HeadObject
-    # operations should benefit from a cache.
-    with botomock(mock_api_call), open(ZIP_FILE, "rb") as f:
-        response = client.post(url, {"file.zip": f}, HTTP_AUTH_TOKEN=token.key)
+    # Upload the same file again. This time some of the S3 HeadObject operations should
+    # benefit from a cache.
+    #
+    # FIXME(willkg): we're not testing whether some of the lookups were from the cache
+    # or not
```
willkg, line 389: The way caching is implemented makes it really hard to test against. When we rewrite the storage layer, we should architect it in a way that makes it possible (easy?) to test whether something was pulled from cache.

tecken/tests/test_upload.py
```diff
@@ -1268,119 +754,39 @@ def test_upload_duplicate_files_in_zip_different_name(fakeuser, client):
...
 
     settings.UPLOAD_DEFAULT_URL = "http://s3.example.com/buck/prefix"
-    bucket_info = get_bucket_info(user)
+    bucket_info = get_bucket_info(uploaderuser)
     assert bucket_info.name == "buck"
     assert bucket_info.endpoint_url == "http://s3.example.com"
     assert bucket_info.region is None
 
 
-def test_get_bucket_info_try_symbols(settings):
```
willkg, line 1309: This test makes no sense to me. Users are put into groups and there's a "Uploaders" group which has both permissions. So this situation isn't possible and therefore we shouldn't be testing it.

willkg:
I need to do a Tecken prod push to land the Redis updates before I land this.

willkg:
I did a prod push and I ran the system tests against stage so I have a baseline. I'm ready to land this and see how it goes.

PR #2775, opened 2023-08-09
bug 1847930: add UBD column

willkg:
This adds an "upload-by-download" column to the uploads view so it's easy to see which were upload-by-download and which weren't.


frontend/src/Uploads.js
```diff
@@ -469,6 +471,7 @@ class DisplayUploads extends React.PureComponent {
                   </td>
                   <td>{upload.user.email}</td>
                   <td>{formatFileSize(upload.size)}</td>
+                  <td>{upload.download_url === null ? "" : "yes"}</td>
```
willkg, line 474: `download_url` is either a url or it's null.

PR #2777, opened 2023-08-14
bug 1848567: fix download API for symbols with special characters

willkg:
This fixes the download API to url quote paths before doing an HTTP HEAD. This correctly handles characters like `+` where it needs to be quoted. Otherwise we get back an HTTP 404 from AWS S3 even if the file is there.

### DIFF
--- a/tecken/base/symboldownloader.py
+++ b/tecken/base/symboldownloader.py
@@ -2,8 +2,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-import time
 from functools import wraps
+import time
+from urllib.parse import quote
 
 from cache_memoize import cache_memoize
 import logging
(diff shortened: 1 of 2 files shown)


tecken/base/symboldownloader.py
```diff
@@ -97,21 +98,26 @@ def invalidate_cache(self, symbol, debugid, filename):
...
+        :arg symbol:
+        :arg debugid:
+        :arg filename:
+
+        :returns: url quoted relative path to be joined with a base url
+
+        """
+        # The are some legacy use case where the debug ID might not already be
+        # uppercased. If so, we override it. Every debug ID is always in uppercase.
+        return quote(f"{prefix}/{symbol}/{debugid.upper()}/{filename}")
```
willkg, line 120: This is weird, but it'll get rewritten when we redo the storage layer api. Having the docstring comment will help me remember to url quote paths when I do an HTTP storage backend.

PR #2787, opened 2023-09-06
bug 1851809: implement createtoken command

willkg:
This will aid development since it makes it possible to create API tokens from the command line.

```
$ docker compose run --rm web bash python manage.py createtoken EMAIL
```

```
$ docker compose run --rm web bash python manage.py createtoken EMAIL TOKENKEY
```

### DIFF
--- a/tecken/tokens/management/commands/createtoken.py
+++ b/tecken/tokens/management/commands/createtoken.py
@@ -12,15 +12,11 @@ class Command(BaseCommand):
     help = "Create an API token."
 
     def add_arguments(self, parser):
-        parser.add_argument("email", default=None, nargs="?")
+        parser.add_argument("email")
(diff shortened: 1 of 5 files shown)


tecken/tokens/management/commands/createtoken.py
```diff
@@ -0,0 +1,48 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument("email", default=None, nargs="?")
+        parser.add_argument("token_key", default=None, nargs="?")
+
+    def handle(self, *args, **options):
+        email = options["email"]
+        if not email:
+            email = input("Email: ").strip()
+        if " " in email or email.count("@") != 1:
+            raise CommandError(f"Invalid email {email!r}")
```
willkg, line 23: I shouldn't make email optional. I'll fix that.

PR #2789, opened 2023-09-07
bug 1746940: fix FileUpload admin; add fake data generator

willkg:
This fixes the `FileUpload` admin which wasn't usable in prod before because it'd create a drop-down for upload which had ... 300k items in it. We'll never want to change that field value, so now it's readonly.

This adds a fake data generator that's kind of terrible, but good enough for now and something we can improve over time as needs arise.

### DIFF
--- a/tecken/upload/admin.py
+++ b/tecken/upload/admin.py
@@ -32,6 +32,9 @@ class UploadAdmin(admin.ModelAdmin):
 
 @admin.register(FileUpload)
 class FileUploadAdmin(admin.ModelAdmin):
+    readonly_fields = [
+        "upload",
+    ]
     date_hierarchy = "created_at"
     search_fields = ["key"]
     list_display = ["id", "upload_id", "bucket_name", "key", "size", "created_at"]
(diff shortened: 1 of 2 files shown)


tecken/upload/admin.py
```diff
@@ -32,6 +32,9 @@ def download_by_upload(self, obj):
 
 @admin.register(FileUpload)
 class FileUploadAdmin(admin.ModelAdmin):
+    readonly_fields = [
+        "upload",
+    ]
```
willkg, line 37: We never want to change this field, so let's make it readonly. That saves us from Django creating a dropdown with every upload entry ... which is 246k currently and makes loading the page really really rough.

tecken/upload/management/commands/generate_fake_data.py
```diff
@@ -0,0 +1,99 @@
...
+                    objs.append(
+                        FileUpload(
+                            upload=upload,
+                            bucket_name="publicbucket",
+                            key=f"/{debug_info['debug_filename']}/{debug_info['debug_id']}/{sym_file}",
+                            size=random.randint(1000, 100000),
+                            # debug_filename=debug_info["debug_filename"],
+                            # debug_id=debug_info["debug_id"],
+                            # code_file=debug_info["code_file"],
+                            # code_id=debug_info["code_id"],
```
willkg, line 96: I commented these out. I'll uncomment them in a PR very soon when we add the sym debug information.

PR #2793, opened 2023-09-07
bug 1746940: rework download API to support /code_file/code_id/sym_file

willkg:
This reworks the download API to support `/code_file/code_id/sym_file` urls. If it sees this, it'll look up `code_file` / `code_id` in the fileupload records. If it finds one, it'll return an HTTP 302 with the correct `debug_filename` and `debug_id`.

This works for both HEAD and GET requests.

This updates the download API documentation as well.

### DIFF
--- a/tecken/download/urls.py
+++ b/tecken/download/urls.py
@@ -26,12 +26,12 @@ app_name = "download"
 
 urlpatterns = [
     path(
-        "try/<str:symbol>/<hex:debugid>/<str:filename>",
+        "try/<str:debugfilename>/<hex:debugid>/<str:filename>",
         views.download_symbol_try,
         name="download_symbol_try",
     ),
     path(
-        "<str:symbol>/<hex:debugid>/<str:filename>",
+        "<str:debugfilename>/<hex:debugid>/<str:filename>",
         views.download_symbol,
         name="download_symbol",
     ),
--- a/tecken/download/views.py
+++ b/tecken/download/views.py
@@ -8,6 +8,7 @@ import markus
 
 from django import http
 from django.conf import settings
+from django.urls import reverse
 
 from tecken.base.decorators import (
     set_request_debug,
@@ -36,10 +37,13 @@ try_downloader = SymbolDownloader(
 )
 
 
-def _ignore_symbol(symbol, debugid, filename):
+def _ignore_symbol(debugfilename, debugid, filename):
     # The MS debugger will always try to look up these files. We
     # never have them in our symbol stores. So it can be safely ignored.
     if filename == "file.ptr":
+        return True
+
+    if debugid == BOGUS_DEBUG_ID:
         return True
 
     if not filename.endswith(tuple(settings.DOWNLOAD_FILE_EXTENSIONS_ALLOWED)):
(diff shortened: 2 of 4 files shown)


tecken/download/urls.py
```diff
@@ -26,12 +26,12 @@ class MixedCaseHexConverter(_Converter):
 
 urlpatterns = [
     path(
-        "try/<str:symbol>/<hex:debugid>/<str:filename>",
+        "try/<str:debugfilename>/<hex:debugid>/<str:filename>",
```
willkg, line 29: This fixes naming to be more unified across documentation and specs.

tecken/download/views.py
```diff
@@ -36,64 +37,73 @@
 )
 
 
-def _ignore_symbol(symbol, debugid, filename):
+def _ignore_symbol(debugfilename, debugid, filename):
```
willkg, line 40: This fixes the naming to be more unified across documentation and specs.

tecken/download/views.py
```diff
@@ -36,64 +37,73 @@
...
     # The MS debugger will always try to look up these files. We
     # never have them in our symbol stores. So it can be safely ignored.
     if filename == "file.ptr":
         return True
 
+    if debugid == BOGUS_DEBUG_ID:
+        return True
```
willkg, line 47: In the previous iteration of this change, we were using the bogus debug id as an indication that we should do a lookup using the `code_file` and `code_id` for the real debug id. Because of that, we needed to not ignore those requests. However, in the case where the debug id is bogus, the debug filename is also missing and the stackwalker won't attempt to use the download API to get the symbol, so that iteration of the change would never work.

Now that we're changing how we're supporting `code_file` / `code_id` lookups, I re-added this.

tecken/download/views.py
```diff
@@ -102,30 +112,17 @@ def download_symbol(request, symbol, debugid, filename, try_symbols=False):
         return response
 
     refresh_cache = "_refresh" in request.GET
+    if refresh_cache:
+        logger.debug("refreshing cache")
```
willkg, line 116: I can probably nix this.

tecken/tests/test_download.py
```diff
@@ -34,19 +34,19 @@ def test_client_happy_path(client, db, s3_helper):
         try_downloader=os.environ["UPLOAD_TRY_SYMBOLS_URL"],
     )
 
-    module = "xul.pdb"
+    debugfilename = "xul.pdb"
     debugid = "44E4EC8C2F41492B9369D6B9A059577C2"
-    debugfn = "xul.sym"
+    symfile = "xul.sym"
```
willkg, line 39: I fixed the naming of things in these tests so they use the right names.

tecken/tests/test_download.py
```diff
@@ -358,20 +360,30 @@ def test_get_code_id_lookup(client, db, s3_helper):
...
+    assert response.status_code == 302
+    parsed = urlparse(response["location"])
+    assert parsed.path == f"/{debug_filename}/{debug_id}/{sym_file}"
 
     # Try with querystring params
     response = client.get(url, {"code_file": code_file, "code_id": code_id})
     assert response.status_code == 302
     parsed = urlparse(response["location"])
     assert parsed.path == f"/{debug_filename}/{debug_id}/{sym_file}"
+    assert parsed.query == f"code_file={code_file}&code_id={code_id}"
```
willkg, line 386: Mmm... I don't know if this might occasionally fail or not. It would depend on whether the transformations are stable. I should probably assume it's not stable and add the other possibility.

PR #2798, opened 2023-09-15
bug 1746940: move info lookup to syminfo API endpoint

willkg:
This moves the codefile/codeid lookup to its own API endpoint and also adds support for looking up by debugfile/debugid. This is generally useful since it provides additional symbol file information. Further, it makes it easier to test the database query and improve it before we put it back in the heavily used download API.

After uploading `libEGL.pdb/E4F12533BFDE65F64C4C44205044422E1/libEGL.sym`, the following work:

Lookup by code file and code id:

```
curl -v 'http://localhost:8000/api/syminfo/libEGL.dll/64F9FA47F000'
```

Lookup by debug file and debug id:

```
curl -v 'http://localhost:8000/api/syminfo/libEGL.pdb/E4F12533BFDE65F64C4C44205044422E1'
```

You get back an answer like this:

```json
{
  "debug_id": "E4F12533BFDE65F64C4C44205044422E1",
  "debug_filename": "libEGL.pdb",
  "code_file": "libEGL.dll",
  "code_id": "64F9FA47F000",
  "generator": "mozilla/dump_syms 2.2.0",
  "url": "http://localhost:8000/libEGL.pdb/E4F12533BFDE65F64C4C44205044422E1/libEGL.sym",
  "time": 0.005093574523925781
}
```


willkg:
I should add caching for this lookup.

I also need to see what happens if there are duplicate fileupload records with the same pairs. I think that's not possible to have happen because a duplicate record would be incomplete and not have this data, but that's not enforced by the db.

tecken/api/views.py
```diff
@@ -628,3 +631,69 @@ def nones_to_zero(obj):
 
     context = {"stats": numbers}
     return http.JsonResponse(context)
+
+
+# Store a result for 10 minutes
+SYMINFO_RESULT_TIMEOUT = 600
```
willkg, line 637: I should add CACHE to this name.

tecken/api/views.py
```diff
@@ -628,3 +631,69 @@ def nones_to_zero(obj):
...
+
+# Indicates there's nothing in the cache
+NO_VALUE_IN_CACHE = object()
+
+
+@metrics.timer_decorator("lookup_by_syminfo.timing")
```
willkg, line 643: We should adjust `timer_decorator` to call transform function `on_return` or `on_error` so they can adjust the timing metric with tags. I'll toss an item in markus for that.

tecken/api/views.py
```diff
@@ -628,3 +631,69 @@ def nones_to_zero(obj):
...
+    This value is cached.
+
+    :arg somefile: a string that's either a debug_file or a code_file
+    :arg someid: a string that's either a debug_id or a code_id
+    :arg refresh_cache: force a cache refresh
+
+    :returns: dict with keys from lookup_by_syminfo
+
+    """
+    key = f"{somefile}//{someid}"
```
willkg, line 656: This key at least needs a prefix. It'd be good if it could use a version, too, so the cache is versioned across deploys. We could use get_version's commit.

tecken/api/views.py
```diff
@@ -628,3 +631,69 @@ def nones_to_zero(obj):
...
+
+        return http.JsonResponse(
+            {
+                "debug_id": ret["debug_id"],
+                "debug_filename": ret["debug_filename"],
+                "code_file": ret["code_file"],
+                "code_id": ret["code_id"],
+                "generator": ret["generator"],
+                "url": new_url,
+                "time": end_time - start_time,
```
willkg, line 695: This is goofy--I should ditch this. If I'm doing timings locally, I can wrap the script in a timer and that's good enough. This is in a timer_decorator so we'll get timings in grafana.

tecken/download/urls.py
```diff
@@ -2,26 +2,11 @@
...
-
-    def to_url(self, value):
-        return value
-
-
-class MixedCaseHexConverter(_Converter):
-    regex = "[0-9A-Fa-f]+"
-
-
-register_converter(MixedCaseHexConverter, "hex")
```
willkg, line 22: I moved this to `tecken/urls.py` because we want to use this converter in a couple of places.

PR #2800, opened 2023-09-18
bug 1746940: thread refresh_cache through api params

willkg:


tecken/api/views.py
```diff
@@ -668,7 +668,10 @@ def cached_lookup_by_syminfo(somefile, someid, refresh_cache=False):
 @api_require_http_methods(["GET"])
 @set_cors_headers(origin="*", methods="GET")
 def syminfo(request, some_file, some_id):
-    ret = cached_lookup_by_syminfo(somefile=some_file, someid=some_id)
+    refresh_cache = "_refresh" in request.GET
```
willkg, line 671: This is the same thing we do with the download API.

PR #2802, opened 2023-09-20
bug 1854219: fix unboundlocalerror in extract_sym_header_data

willkg:
This fixes the unboundlocalerror when creating the SymParseError. It also fixes the memory usage issue by switching to readline instead of readlines.

### DIFF
--- a/tecken/upload/utils.py
+++ b/tecken/upload/utils.py
@@ -72,8 +72,9 @@ def extract_sym_header_data(file_path):
         "generator": "",
     }
     with open(file_path, "r") as fp:
+        line = "no line yet"
         try:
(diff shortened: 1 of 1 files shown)


tecken/upload/utils.py
```diff
@@ -72,8 +72,9 @@ def extract_sym_header_data(file_path):
         "generator": "",
     }
     with open(file_path, "r") as fp:
+        line = "no line yet"
         try:
-            for line in fp.readlines():
+            for line in fp:
```
willkg, line 77: This switches to using `readline` rather than `readlines` which reads the whole thing into memory.

PR #2805, opened 2023-09-22
bug 1851909: remove unneeded upload_id method

willkg:
upload_id already works and we don't need the additional upload_id method which forces the Django admin to additionally pull each upload record one-by-one.


willkg:
We should hold off on landing this until bug 1853981 is done so we can test the migration.

PR #2812, opened 2023-10-04
bug 1746940: implement codeinfo lookup in download API

willkg:
This re-implements the codeinfo lookup in the download API which kicks in if a symbol file wasn't found and if the debugfilename/debugid look like a possible codefile/codeid.

### DIFF
--- a/tecken/api/views.py
+++ b/tecken/api/views.py
@@ -9,7 +9,6 @@ import markus
 
 from django.conf import settings
 from django.contrib.auth.models import Permission
-from django.core.cache import cache
 from django.core.exceptions import PermissionDenied, BadRequest
 from django.db.models import Aggregate, Count, Q, Sum
 from django import http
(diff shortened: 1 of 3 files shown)


tecken/api/views.py
```diff
@@ -632,43 +632,6 @@ def nones_to_zero(obj):
...
 
-# Store a result for 10 minutes
-SYMINFO_RESULT_CACHE_TIMEOUT = 600
-
-# Indicates there's nothing in the cache
-NO_VALUE_IN_CACHE = object()
-
-
-@metrics.timer_decorator("syminfo.lookup.timing")
-def cached_lookup_by_syminfo(somefile, someid, refresh_cache=False):
```
willkg, line 643: I moved this to the download view code. It's a little awkward there, too, but it seemed a better place than here.

PR #2825, opened 2023-10-23
bug 1860576: Make Docker Hub upload script more robust.

smarnach:
This is the same change as https://github.com/mozilla-services/antenna/pull/947, but for Tecken.


smarnach:
CI here is failing because of the bug fixed by #2824 and vice versa. We can either merge one of these PRs with CI failing, or merge them into a single PR.

willkg (approved):
Nice! :rocket:

I think we can merge this PR into the other one (or vice versa)--whichever is more convenient for you.

PR #2826, opened 2023-10-23
bug 1860574: Make "make clean" recreate .gitkeep file.

smarnach:


Makefile
```diff
@@ -64,6 +64,8 @@ clean: .env stop  ## | Stop and remove docker containers and artifacts.
 	${DC} rm -f
 	rm -fr .docker-build
 	rm -rf frontend/build/
+	mkdir -p frontend/build/static
+	touch frontend/build/static/.gitkeep
```
willkg, line 68: What if instead of `touch` we did `git checkout frontend/build/static/.gitkeep` so it's "the same file"? Does that create problems with git workflow stuff?

smarnach: The question whether one empty file is "the same" as some other file is a bit philosophical, but I like this better anyway. I went with restoring the whole directory.

willkg: Theseus' respository...

willkg (approved):
I never knew about `git restore`--that's neat!

PR #2834, opened 2023-11-20
bug-1809727: upgrade django to 4.2.7 LTS

biancadanforth:
@smarnach went through the release notes for Django comparing to the previously installed version and reviewed all the major changes and the extent to which we were using those Django features. He didn't find any changes strictly necessary, though he did find that we could upgrade our Postgres DB library `psycopg2` to version 3.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -44,5 +44,5 @@ importlib-metadata==6.8.0
 typing-extensions==4.8.0
 zipp==3.17.0
 
-# NOTE(willkg): We stick with LTS releases and the next one is 4.2 (2023).
-Django==3.2.23
+# NOTE(willkg): We stick with LTS releases.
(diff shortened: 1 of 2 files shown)


smarnach (approved):
Django 3.2 does not support Python 3.11, so we needed to upgrade Django before Python.

We were unable to find anything that needs to be changed for the Djanog upgrade. The CI tests still pass. I also ran the system tests, but it looks like [the list of symbols files to be downloaded](https://github.com/mozilla-services/tecken/blob/main/systemtests/data/sym_files_to_download.csv) needs to be updated first; other than the 404s resulting from the out-of-date list, the tests passed.

I filed bug-1865780 for the psycopg upgrade.

I can't think of anything else to do, so this looks good.

PR #2835, opened 2023-11-20
 bug-1743835: upgrade Python from 3.9 to 3.11 

biancadanforth:
No source code changes were necessary, just some find-and-replace for the Python version in a readthedocs config file and in one Sentry test for paths to Python site packages.

This PR is based on #2834 , and it shouldn't be merged until that one is merged. This is because Sven noted that [Django 3.2 doesn't support Python 3.11](https://docs.djangoproject.com/en/4.2/releases/3.2/#python-compatibility), so we have to upgrade Django first before upgrading Python.


smarnach (approved):
Looks good to me.

I also tread the "Porting to Python 3.10" and "Porting to Python 3.11" sections in the release notes, and didn't find anything we'd need to do, so I think we are done.

PR #2837, opened 2023-11-27
install procps and net-tools into container for troubleshooting

haroldwoo:


willkg:
Based on what you've said in Slack so far, the problem isn't in the container but rather the host. Making these changes won't help debugging issues with the host.

haroldwoo:
will still help in the future

willkg:
We worked out the issue in PR #2838 , so I think we're good here. I don't really want to install these tools into the container unless we need them. We can install them then.

PR #2840, opened 2023-11-28
bug-1853962: improve disk manager resilience

willkg:
- bug-1853962: improve disk cache manager error handling
- bug-1853962: change UPLOAD_TEMPDIR does not exist into non-error

This does a few things to make the disk manager more resilient:
    
1. removes the daemon functionality: the script runs in a loop already, so we don't need the additional daemon loop in the command
2. wrap the `remove_orphaned_files` command in `sentry-wrap`: this will make sure we get errors sent to sentry
3. add a timeout for the command: we probably don't need a timeout, but this adds one anyway so if it takes longer than 60 seconds to run, `sentry-wrap` will kill it and send an error to Sentry
4. adds a `--verbose` to `sentry-wrap`: this makes it so that `sentry-wrap` doesn't spit out non-essential output unless asked to
5. fix output: this removes non-error output which gets logged and moves errors to stderr

A while back, we had a problem where the db disappeared and then the disk manager process died and then the instances all went sideways after some time as they accumulated orphaned files. Lots of problems here, but this reduces the lack of resilience with the disk manager so it doesn't die when the db isn't available because it doesn't need the db.

This also improves the likelihood we get some signal that things are bad in Sentry with details we can use to debug the issue.

### DIFF
--- a/bin/sentry-wrap.py
+++ b/bin/sentry-wrap.py
@@ -51,13 +51,18 @@ def test_sentry(ctx):
     default=300,
     help="Timeout in seconds to wait for process before giving up.",
 )
+@click.option(
+    "--verbose/--no-verbose",
+    default=False,
+    help="Whether to print verbsoe output.",
+)
 @click.argument("cmd", nargs=-1)
 @click.pass_context
-def wrap_process(ctx, timeout, cmd):
+def wrap_process(ctx, timeout, verbose, cmd):
     sentry_dsn = os.environ.get("SENTRY_DSN")
 
     if not sentry_dsn:
-        click.echo("SENTRY_DSN is not defined. Exiting.")
+        click.echo("SENTRY_DSN is not defined. Exiting.", err=True)
         sys.exit(1)
 
     if not cmd:
@@ -69,7 +74,8 @@ def wrap_process(ctx, timeout, cmd):
 
     cmd = " ".join(cmd)
     cmd_args = shlex.split(cmd)
-    click.echo(f"Running: {cmd_args}")
+    if verbose:
+        click.echo(f"Running: {cmd_args}")
 
     try:
         ret = subprocess.run(cmd_args, capture_output=True, timeout=timeout)
@@ -86,13 +92,14 @@ def wrap_process(ctx, timeout, cmd):
             click.echo(ret.stdout.decode("utf-8"))
             click.echo(ret.stderr.decode("utf-8"))
             time_delta = (time.time() - start_time) / 1000
-            click.echo(f"Fail. {time_delta:.2f}s")
+            click.echo(f"Command failed. time: {time_delta:.2f}s", err=True)
             ctx.exit(1)
 
         else:
-            click.echo(ret.stdout.decode("utf-8"))
+            click.echo(ret.stdout.decode("utf-8"), nl=False)
             time_delta = (time.time() - start_time) / 1000
-            click.echo(f"Success! {time_delta:.2f}s")
+            if verbose:
+                click.echo(f"Success. time: {time_delta:.2f}s")
 
     except click.exceptions.Exit:
         raise
--- a/bin/sentry-wrap.py
+++ b/bin/sentry-wrap.py
@@ -78,19 +78,24 @@ def wrap_process(ctx, timeout, verbose, cmd):
         click.echo(f"Running: {cmd_args}")
 
     try:
-        ret = subprocess.run(cmd_args, capture_output=True, timeout=timeout)
+        ret = subprocess.run(
+            cmd_args,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            timeout=timeout,
+        )
         if ret.returncode != 0:
             sentry_sdk.set_context(
                 "status",
                 {
                     "exit_code": ret.returncode,
-                    "stdout": ret.stdout.decode("utf-8"),
-                    "stderr": ret.stderr.decode("utf-8"),
+                    # Truncate stdout/stderr to the last 5,000 characters in case the
+                    # output is monstrously large.
+                    "stdout": ret.stdout.decode("utf-8")[-5000:],
                 },
             )
             capture_message(f"Command {cmd!r} failed.")
             click.echo(ret.stdout.decode("utf-8"))
-            click.echo(ret.stderr.decode("utf-8"))
             time_delta = (time.time() - start_time) / 1000
             click.echo(f"Command failed. time: {time_delta:.2f}s", err=True)
             ctx.exit(1)
--- a/tecken/tests/test_upload.py
+++ b/tecken/tests/test_upload.py
@@ -933,6 +933,28 @@ def test_clearuploads_records_dry_run(db, fakeuser):
 
 
 class Test_remove_orphaned_files:
+    def test_no_watchdir(self, db, settings, caplog):
+        caplog.set_level(logging.INFO)
+
+        tempdir = "/does/not/exist"
+        settings.UPLOAD_TEMPDIR = tempdir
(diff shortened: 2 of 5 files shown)


willkg:
I messed up--this isn't a _disk cache_ manager, it's just a _disk_ manager. So I need to fix the naming.

biancadanforth (commented):
Well, this was a whirlwind adventure. Thanks for walking me through the patch and the context around the patch.

I'm going to try to test this locally based on your instructions.

The only thing we talked about was adding a couple of tests around certain kinds of errors/non-errors:
* add a new test for no upload directory
* update the `test_errors` test to ensure it's using the Sentry mock

willkg:
^^^ That adds a test for when the watchdir doesn't exist, improves the errors test to assert that a Sentry event is emitted, fixes "disk cache manager" to "disk manager" in descriptive text, and changes `SLEEP_SECONDS` to 60.

That should be ready for re-review.

willkg:
^^^ That fixes sentry-wrap where it was ignoring stderr in the case where the subprocess completed successfully.

bin/sentry-wrap.py
```diff
@@ -69,30 +74,37 @@ def wrap_process(ctx, timeout, cmd):
...
-    click.echo(f"Running: {cmd_args}")
+    if verbose:
+        click.echo(f"Running: {cmd_args}")
 
     try:
-        ret = subprocess.run(cmd_args, capture_output=True, timeout=timeout)
+        ret = subprocess.run(
+            cmd_args,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
```
willkg, line 84: I switched this from `capture_output=True` which captures stdout and stderr separately to this which captures them both in the stdout pipe, but multiplexed so the lines are in the same order that you'd see them in a terminal. That's a lot easier to understand when debugging things.

bin/sentry-wrap.py
```diff
@@ -69,30 +74,37 @@ def wrap_process(ctx, timeout, cmd):
...
         if ret.returncode != 0:
             sentry_sdk.set_context(
                 "status",
                 {
                     "exit_code": ret.returncode,
-                    "stdout": ret.stdout.decode("utf-8"),
-                    "stderr": ret.stderr.decode("utf-8"),
+                    # Truncate stdout/stderr to the last 5,000 characters in case the
+                    # output is monstrously large.
+                    "stdout": ret.stdout.decode("utf-8")[-5000:],
```
willkg, line 94: It's possible for the subprocess to generate a _lot_ of output. In practice, I don't think we use sentry-wrap in a way that this is likely to happen except for when we run migrations. It's possible a migration we wrote could have a lot of output.

It his truncates the output to the last 5000 characters before sending it to Sentry.

tecken/upload/utils.py
```diff
@@ -153,6 +153,9 @@ def __init__(self, path, name):
     def size(self):
         return os.stat(self.path).st_size
 
+    def __repr__(self):
```
willkg, line 156: This is a function that runs whenever you print the repr of something. It's better than the default which in this case is something like `<FileMember object at 0x7fa4879902e0>`. Now it'll print something like `<FileMember /tmp/uploads/ou812/something.sym something.sym>`.

bin/sentry-wrap.py
biancadanforth: I've never seen output truncated by number of character before -- only number of lines (like `--tail 10` returns the last 10 lines of a log file in `pm2`, a process manager I've used before).

Is there a possibility that a single line could be 5000 characters (or otherwise thousands of characters) long, such that this might not contain enough of the output for debugging?

tecken/tests/test_upload.py
```diff
@@ -933,6 +933,28 @@ def test_clearuploads_records_dry_run(db, fakeuser):
...
+        caplog.set_level(logging.INFO)
+
+        tempdir = "/does/not/exist"
+        settings.UPLOAD_TEMPDIR = tempdir
+        settings.UPLOAD_TEMPDIR_ORPHANS_CUTOFF = 5
+
+        call_command("remove_orphaned_files", verbose=True)
+
+        # Make sure we've got the right expires value
+        assert "expires: 5 (minutes)" in caplog.text
```
biancadanforth, line 946: Huh, I'm surprised that the tests in this file are making assertions based on what's logged rather than what functions are/are not called or what the function under test returns. Why is it set up this way?

It seems like the log line/branching logic could stay the same, but some substantive part of the function body could change (say a function is no longer called that previously was), and these tests would still pass. 🤔

tecken/tests/test_upload.py
```diff
@@ -1081,15 +1105,21 @@ def adjusted_getmtime(fn):
...
-        )
-        assert len(incr_records) == 1
+            # Assert a Sentry event was emitted
+            event = sentry_client.events[0]
+            assert event["logentry"]["message"] == "error getting size: %s"
+            assert event["logger"] == "tecken.remove_orphaned_files"
+
+            # Assert metric is emitted
+            incr_records = metricsmock.filter_records(
+                "incr", stat="tecken.remove_orphaned_files.delete_file_error"
```
biancadanforth, line 1123: Ah yeah, this is more what I'm used to seeing in tests. I mean sometimes we do make assertions about what's logged, but those are typically secondary to what functions are/aren't being called, whether an error is thrown, etc.

tecken/upload/utils.py
biancadanforth: TIL: https://docs.python.org/3/library/functions.html#repr

bin/sentry-wrap.py
biancadanforth: I can confirm I'm now seeing stderr and stdout in the `make run` output now locally. I can also now see the disk manager logs:

```
tecken-web-1         | [2023-11-29 21:40:37 +0000] [29] [ERROR] Worker (pid:42) exited with code 1
tecken-web-1         | [2023-11-29 21:40:37 +0000] [29] [ERROR] Worker (pid:42) exited with code 1.
tecken-web-1         | [2023-11-29 21:40:37 +0000] [24812] [INFO] Booting worker with pid: 24812
tecken-web-1         | WARNING 2023-11-29 21:40:41,662 - webapp - tecken Auth0 blocked check disabled
tecken-web-1         | INFO 2023-11-29 21:41:17,783 - disk_manager - tecken.remove_orphaned_files expires: 15 (minutes)
tecken-web-1         | INFO 2023-11-29 21:41:17,788 - disk_manager - tecken.remove_orphaned_files watchdir: '/tmp/uploads'
```

biancadanforth (approved):
Another whirlwind adventure with Shrek and Donkey! (I'm Donkey)

I spent a lot of time trying to get an orphaned file in my local stack, so that the disk manager could clean it up via `remove_orphaned_files`, but the failure mode I was using (Gunicorn worker terminated by Gunicorn manager due to request timeout) does not create an orphaned file. I am not sure what kinds of failure modes do, so I will just leave it at that for now.

I have one or two non-blocking questions in-line, but other than that I will let the matter rest! Thanks Willkg for working with me on some STR, dev setup and helper scripts.

bin/sentry-wrap.py
willkg: It is possible--everything is possible in a sense. I don't think we'll hit this issue with what we've got currently. But when/if we do, we'll see it in Sentry and we can figure out a better way of truncating. Maybe we truncate characters in a line and truncate lines, too.

tecken/upload/utils.py
willkg: repr is really handy sometimes. It's what gets called when you do things like:

```
logger.info("watchdir: %r", watchdir)

print(f"watchdir: {watchdir!r}")
```

willkg:
Thank you for reviewing this! It improved this PR a lot.

PR #2842, opened 2023-11-30
bug-1867587: improve dev docs; add slick.sh script

willkg:
- bug-1867587: Improve dev docs pulling how to information into one place
- bug-1867587: add slick.sh-dist

This improves the dev documentation and adds a "how to" section that walks through using the local dev environment.

This fixes some minor ergonomic issues with the `Makefile` so that the dev documentation works even in some edge cases. It makes it a little more maintainable.

This also adds my `slick.sh` script. In order to add that, I generalized a `bin/cp-dist-script.sh` script and reworked how `.env` is created to use that, too.

After this, doing things like:

```
make .env
```
and:
```
make slick.sh
```
should work correctly and _not stomp on an existing file_.

Additionally these should continue to work:

```
make build
make run
make test
```

Further, dev documentation should be clearer and include more information on how to manipulate your local dev environment for debugging, testing, and other engineering work. You can see it by pulling down these changes:

```
gh pr checkout 2842
make docs
```
And then opening `docs/_build/html/dev.html` in your browser.

There was a section on generating a metrics file--whatever that was, it's busted. I wrote up [bug 1867570](https://bugzilla.mozilla.org/show_bug.cgi?id=1867570) to look into that.


Makefile
```diff
@@ -16,6 +16,8 @@ USE_GID ?= 10001
 DOCKER := $(shell which docker)
 DC=${DOCKER} compose
 
+SERVICES=db fakesentry redis-cache localstack statsd oidcprovider
```
willkg, line 19: This list of services showed up in a couple of places in the file. In order to make sure we don't miss one in one of the places, it's better to centralize the list here.

Makefile
```diff
@@ -81,12 +87,12 @@ psql: .env .docker-build  ## | Open psql cli.
 
 .PHONY: test
 test: .env .docker-build  ## | Run Python unit test suite.
-	${DC} up -d db redis-cache localstack statsd oidcprovider
+	${DC} up --detach ${SERVICES}
```
willkg, line 90: I changed the flag from `-d` to the more descriptive and human-friendly `--detach`.

docker/config/slick.sh-dist
```diff
@@ -0,0 +1,43 @@
...
+# Run setup
+make setup
+
+# Set up user
+docker compose exec oidcprovider /code/manage.py createuser "${FAKEUSERNAME}" "${FAKEPASSWORD}" "${FAKEEMAIL}"
+docker compose run --rm web bash python manage.py superuser "${FAKEEMAIL}"
+docker compose run --rm web bash python manage.py createtoken "${FAKEEMAIL}" "${FAKETOKEN}"
+#
+# Reset fakesentry
+if command -v curl &> /dev/null
```
willkg, line 37: I wrapped this section in a "if they have curl installed" thing, but this is half-hearted. It doesn't tell tell the user to install curl. It doesn't tell you it didn't reset fakesentry. I felt like I spent a lot of time on this PR and then didn't want to keep spending time on it, so I left this bit in this state.

docs/dev.rst
```diff
@@ -88,6 +88,195 @@ To run the webapp service, do:
...
+============  ====  =============================================
+frontend      3000  Javascript proxy for webapp--use with browser
+web           8000  Django webapp--use with APIs
+localstack    4566  S3 emulation service
+db            5432  Postgres database
+redis         6379  Redis service
+fakesentry    8090  Sentry emulation service
+oidcprovider  8080  SSO emulation service
+statsd        8081  Grafana / statsd
+============  ====  =============================================
```
willkg, line 108: Having this table is realllly helpful. I wish I had created it ages ago. I can never remember which fake service is on which port.

docs/dev.rst
```diff
@@ -88,6 +88,195 @@ To run the webapp service, do:
...
+   Commands:
+     create        Creates a bucket
+     delete        Deletes a bucket
+     list_buckets  List S3 buckets
+     list_objects  List contents of a bucket
+
+
+How to access the database
+--------------------------
+
```
willkg, line 275: Oops--I forgot to write this.

docs/dev.rst
```diff
@@ -428,107 +552,43 @@ this code and it has to be tested manually. You can do something like this:
...
-
-Additionally, in a local development environment, you can create a
-corresponding user in the oidcprovider service like this:
-
-.. code-block:: shell
-
-   $ docker compose exec oidcprovider /code/manage.py createuser yourname yourpassword yourname@example.com
-
-
-Giving users permission to upload symbols
```
willkg, line 474: This section is now a runbook in Confluence, so I removed it from here.

docs/dev.rst
```diff
@@ -428,107 +552,43 @@ this code and it has to be tested manually. You can do something like this:
...
-   permission.
-5. Resolve the bug.
+* local: local dev environment
+* stage: the stage server environment
+* prod: the prod server environment--will not run destructive tests
 
+System tests can help verify that upload API and download API work. They
+periodically need to be updated as symbols files expire out of the systems.
 
-Viewing all metrics keys
```
willkg, line 495: This section is busted. The `LogAllMetricsKeys` thing is written for an older version of markus and needs to be updated. I wrote up bug 1867570 to rework this.

docs/dev.rst
```diff
@@ -428,107 +552,43 @@ this code and it has to be tested manually. You can do something like this:
...
-there that you know you don't need or care about in Datadog, then delete it
-from the code.
+They're generally unmaintained, but can be a good starting point for a new load
+testing effort.
 
-The file ``all-metrics-keys.json`` can be deleted any time and it will be
-recreated again.
 
 
-Localstack (S3 mock server)
```
willkg, line 518: We used to use minio and I think this worked with minio, but it doesn't work with localstack, so I removed this section.

docs/dev.rst
```diff
@@ -428,107 +552,43 @@ this code and it has to be tested manually. You can do something like this:
...
-When started with docker, it starts a web server on ``:4566`` that you can
-use to browse uploaded files. Go to ``http://localhost:4566``.
-
-.. _localstack: https://github.com/localstack/localstack
+How to do local Upload by Download URL
+======================================
 
+.. Note::
 
-How to do local Upload by Download URL
```
willkg, line 530: I left this section here. It probably needs to be updated, but I don't want to spend the time on it right now. The systemtests have a download-by-upload test that looks at symbols.zip files from recent Taskcluster jobs and that's probably sufficient, so maybe we should just delete this section altogether.

https://github.com/mozilla-services/tecken/blob/ec4b677c8cbfe4705aeb0974d96958c677df64ae/systemtests/test_env.sh#L66-L69

Which would you prefer?

docs/dev.rst
```diff
@@ -562,11 +622,15 @@ Now, you can send these in with ``tecken-loadtest`` like this:
    $ python upload-symbol-zips.py http://localhost:8000 -t 160 --download-url=http://localhost:9090/symbols.zip
 
 This way you'll have 3 terminals. 2 bash terminals inside the container
-and one outside in the ``tecke-loadtests`` directory on your host.
+and one outside in the ``tecken-loadtests`` directory on your host.
 
 
 Debugging a "broken" Redis
```
willkg, line 628: I have no idea if this section is valid anymore. It pre-dates me and I've never used it, so maybe we should just delete it.

docs/dev.rst
```diff
@@ -594,18 +658,22 @@ verify the session cookie value on each and every request.
 
 
 Auth debugging
```
willkg, line 660: I have no idea if this section is valid anymore. It also pre-dates me and I've never used it.

smarnach (approved):
Looks good to me. I had a few comments.

docker/config/slick.sh-dist
```diff
@@ -0,0 +1,43 @@
...
+FAKEEMAIL=""
+
+# 32-character hex token
+FAKETOKEN="c7c1f8cab79545b6a06bc4122f0eb3cb"
+
+# Build Tecken
+make build
+
+# Start services
+docker compose up -d --remove-orphans db redis-cache statsd localstack oidcprovider fakesentry
```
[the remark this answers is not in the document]
willkg: I have this line in my `slick.sh` script because I've had a lot of problems with race conditions and other things, so having this line makes it more likely that all the backing services have fully started. We could add `waitfor` lines to `make setup` so it starts services and then waits to make sure they're started up. That would probably alleviate some of the problems I had, but it's probably easier to just do this.

Plus it has the `--remove-orphans` flag which removes orphaned services.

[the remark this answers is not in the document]
willkg: We can take it out and do any other improvements in follow-up PRs.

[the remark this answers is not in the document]
smarnach: In that case, I think the right solution is to bake this into the Docker Compose config, using health checks and dependencies to make sure the services are ready before a web container is started. Maybe at some point in the future, doesn't seem urgent.

PR #2843, opened 2023-11-30
bug-1847235: emit tecken.gunicorn_worker_abort incr on timeout

willkg:
This uses gunicorn server hooks to emit a tecken.gunicorn_worker_abort incr when the gunicorn manager terminates a gunicorn worker because it's exceeded the timeout. This happens when it's taking too long to process an upload API request.

The test for this is contrived at best and only verifies that the metric we expect is emitted.

It's difficult to do a real test because that requires Tecken to be running with gunicorn and for a gunicorn worker to exceed the timeout such that the gunicorn manager kills the process. We can manually test it using the steps in this blog post:

https://bluesock.org/~willkg/blog/mozilla/tecken_worker_exit.html

When the Gunicorn manager sends a SIGABRT to the Gunicorn worker, we'll see the metric emitted:

```
tecken-web-1           | INFO 2023-11-30 20:26:08,248 - webapp - markus METRICS|2023-11-30 20:26:08|timing|tecken.upload_file_exists|6.280567002249882|
tecken-web-1           | DEBUG 2023-11-30 20:26:08,249 - webapp - tecken key_existing cache miss on publicbucket:v1/libglib-2.0.so.0/441450BE1030C69946EC77871CA83C5D0/libglib-2.0.so.0
tecken-web-1           | [2023-11-30 20:27:08 +0000] [16] [CRITICAL] WORKER TIMEOUT (pid:22)
tecken-web-1           | INFO 2023-11-30 20:27:08,968 - webapp - markus METRICS|2023-11-30 20:27:08|incr|tecken.gunicorn_worker_abort|1|
tecken-web-1           | [2023-11-30 20:27:09 +0000] [16] [ERROR] Worker (pid:22) was sent SIGKILL! Perhaps out of memory?
tecken-web-1           | [2023-11-30 20:27:09 +0000] [34] [INFO] Booting worker with pid: 34
tecken-web-1           | WARNING 2023-11-30 20:27:10,365 - webapp - tecken Auth0 blocked check disabled
```

### DIFF
--- a/docker/config/local_dev.env
+++ b/docker/config/local_dev.env
@@ -7,6 +7,9 @@
 # Tecken
 # ------
 
+# Gunicorn things
+GUNICORN_TIMEOUT=60
+
 # Statsd things
 STATSD_HOST=statsd
 STATSD_PORT=8125
--- a/tecken/gunicornhooks.py
+++ b/tecken/gunicornhooks.py
@@ -6,19 +6,23 @@
 Gunicorn server hooks.
 
 See https://docs.gunicorn.org/en/stable/settings.html#server-hooks
+
+Note: Don't import anything that involves Django machinery here.
 """
 
 import markus
-
-from tecken import settings
 
 
 metrics = markus.get_metrics("tecken")
 
 
-def configure_markus():
-    markus.configure(settings.MARKUS_BACKENDS)
-
-
 def worker_abort(worker):
+    """Emit metric when a Gunicorn worker is terminated from timeout
+
+    .. Note::
+
+       This gets called by the Gunicorn worker handle_abort() function so it'll use the
+       markus and logging configuration of the Django webapp.
+
+    """
     metrics.incr("gunicorn_worker_abort")
--- a/tecken/gunicornhooks.py
+++ b/tecken/gunicornhooks.py
@@ -1,0 +1,24 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+"""
+Gunicorn server hooks.
+
+See https://docs.gunicorn.org/en/stable/settings.html#server-hooks
+"""
+
+import markus
+
+from tecken import settings
+
+
+metrics = markus.get_metrics("tecken")
+
+
+def configure_markus():
+    markus.configure(settings.MARKUS_BACKENDS)
+
+
+def worker_abort(worker):
+    metrics.incr("gunicorn_worker_abort")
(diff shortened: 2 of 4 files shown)


docker/config/local_dev.env
```diff
@@ -7,6 +7,9 @@
 # Tecken
 # ------
 
+# Gunicorn things
+GUNICORN_TIMEOUT=60
```
willkg, line 11: This drops the `GUNICORN_TIMEOUT` in the local dev environment to 60 seconds. It reduces the number of steps it takes to simulate the gunicorn worker timeout scenario and 60 seconds is probably a better value for a local dev environment anyhow.

tecken/gunicornhooks.py
```diff
@@ -0,0 +1,24 @@
...
+
+import markus
+
+from tecken import settings
+
+
+metrics = markus.get_metrics("tecken")
+
+
+def configure_markus():
```
smarnach, line 19: What is calling this function? It doesn't seem to be one of the predefined Gunicorn hooks, so I don't understand how it gets called. Maybe this should be called `post_worker_init()`, or whatever the appropriate hook is?

tecken/gunicornhooks.py
```diff
@@ -0,0 +1,24 @@
...
+
+"""
+Gunicorn server hooks.
+
+See https://docs.gunicorn.org/en/stable/settings.html#server-hooks
+"""
+
+import markus
+
+from tecken import settings
```
smarnach, line 13: This doesn't pull in any actual Django machinery, right? I'm thinking about whether this might be to heavy-weight to pull into gunicorn, but I don't think it is.

smarnach (approved):
Overall this looks reasonable and simple. I don't understand how the initialization works, but I'm approving the PR so we can deploy it if you decide to go ahead.

tecken/gunicornhooks.py
willkg: That's correct. This doesn't pull in any Django machinery. I'll a note about that at the top of the file.

tecken/gunicornhooks.py
willkg: Whoops--it doesn't get called. But it works when I manually test things plus the logging output is what we configure in the Django webapp and not what Gunicorn is emitting, so I wonder if `worker_abort` gets called in the gunicorn worker in the SIGABRT handling and not the Gunicorn master process.

I checked the code and that's what's going on. So we don't need the `configure_markus` and some of that other stuff at all. I'll remove it and add a note.

https://github.com/benoitc/gunicorn/blob/ca9162d9cd0f2403a4b65a54ddb996eb855e8b58/gunicorn/workers/base.py#L200-L203

willkg:
Thank you!

PR #2852, opened 2023-12-04
bug-1847235: move markus usage to hook function

willkg:
We don't want the gunicorn manager process to import Tecken webapp stuff, so this moves importing markus and creating a markus client to the point where it's used.


smarnach (approved):

willkg:
Thank you!

PR #2853, opened 2023-12-05
bug-1868132: add CODEOWNERS file

willkg:
This adds a CODEOWNERS file which adds a default rule covering the
entire repository with the obs-team.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+# Default owner for everything in this repository
+*   @mozilla-services/obs-team


willkg:
@biancadanforth Does this look right for the `CODEOWNERS` file structure?

.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+# Default owner for everything in this repository
+*   @mozilla-services/obs-team
```
biancadanforth, line 2: This looks fine to me; FWICT the extra whitespace doesn't matter (file is valid; I see similar extra whitespace in the [example file](https://github.com/mozilla/fxa/blob/ed0de1b49ac1c548c8dd3b2d299c62ab1b6d2897/.github/CODEOWNERS#L1) in the docs).

willkg:
I think we do want to add that branch protection. I'll make that change when we land this.

willkg:
Thank you!

PR #2854, opened 2023-12-06
bug-1835888: initialize CLOUD_SERVICE_PROVIDER env var

biancadanforth:
Because:
* There is application-level branching logic required[1], depending on which cloud service provider we are using.

This commit:
* Creates a new env var, CLOUD_SERVICE_PROVIDER, defaulting to "AWS"
* Adds this env var to the Settings section in the Django /admin/sitestatus page.

[1]: For example, markus needs to add an additional tag in GCP that is already tagged by Telegraf in AWS. Another example is that we will need different storage backend setup code depending on if we're using AWS S3 or GCP GCS.

Tecken config docs:
Screenshot 2023-12-06 at 1 14 19 PM

Django admin settings page:
Screenshot 2023-12-06 at 1 14 07 PM

I have separately added a line item for the [Tecken GCP migration plan](https://docs.google.com/document/d/1tLoGhMJQrP1m6A5H3D5zElNbTHQ9_vzYbOE9QnrxLUc/edit?usp=sharing) to set this env var to "GCP" as part of setting up the GCP stage and prod environments. Since it is defaulting to "AWS", no further upstream config is needed in our cloudops repo currently.

### DIFF
--- a/tecken/settings.py
+++ b/tecken/settings.py
@@ -421,6 +421,12 @@ else:
         },
     }
 
+CLOUD_SERVICE_PROVIDER = _config(
+    "CLOUD_SERVICE_PROVIDER",
+    default="AWS",
+    doc="The cloud service provider Tecken is using. Either AWS or GCP.",
+)
+
 AWS_ACCESS_KEY_ID = _config("AWS_ACCESS_KEY_ID", default="", doc="AWS access key id.")
 AWS_SECRET_ACCESS_KEY = _config(
     "AWS_SECRET_ACCESS_KEY", default="", doc="AWS secret access key."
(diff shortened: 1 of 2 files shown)


tecken/settings.py
```diff
@@ -421,6 +421,12 @@ def filter(self, record):
         },
     }
 
+CLOUD_SERVICE_PROVIDER = _config(
```
biancadanforth, line 424: I had a few questions here:
* Is there an optimal location within this `settings.py` file to add this new env var?
* Is there a way to enforce that this config is an enum (i.e. one of "AWS" or "GCP")? I realize we may do this in the code downstream.
* I noticed in the Tecken docs that this env var is listed as "Required: No". I don't know how to configure that or if I need to. I only see a small number of env vars listed as "Required: Yes".

biancadanforth (commented):
Just added a few questions inline.

PR #2855, opened 2023-12-07
bug-1868790: Return Unix timestamps from get_last_modified().

smarnach:
The results of `get_last_modified()` are cached in Redis, so we can't use datetime objects here, since they can't be serialized by MessagePack. We did not catch this mistake in the tests because the tests used to use an in-memory cache without serialization. This change also makes the tests use an actual Redis container in the tests.

### DIFF
--- a/docker/config/test.env
+++ b/docker/config/test.env
@@ -31,3 +31,7 @@ OIDC_RP_CLIENT_SECRET=abcdef
 OIDC_OP_USER_ENDPOINT=https://auth.example.com/authorize
 
 UPLOAD_TEMPDIR=/tmp/test/uploads
+
+# Use a different Redis database for the tests, so the tests don't interfere
+# with the dev environment.
+REDIS_URL=redis://redis-cache:6379/1
--- a/tecken/base/symboldownloader.py
+++ b/tecken/base/symboldownloader.py
@@ -2,7 +2,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-import datetime
 from email.utils import parsedate_to_datetime
 from functools import wraps
 import time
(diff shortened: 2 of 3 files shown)


willkg (approved):
This looks good!

The only thing I'm wondering about is that now the local dev environment and the test environment share a Redis service.

The tests wipe the cache between tests:

https://github.com/mozilla-services/tecken/blob/main/tecken/tests/conftest.py#L21-L23

That'll wipe the cache for the local dev environment, too. I looked through django-redis and redis-py to see if they had prefixes we could set when creating the client. Then we could have the test environment use a different prefix than the local dev environment. I don't think either library supports a client-scoped prefix. I'm not sure this is easily fixable.

I think we should go with this for now and maybe think about segregating the test and local dev environment Redis cache somehow if we find it to be a problem down the road.

smarnach:
Good point regarding the tests interfering with the dev environment. I think we can simply use different databases for the dev environment and the tests. I updated the PR to make the tests use database 1, while the dev env uses database 0. Could you take another look?

willkg:
I pulled this down, ran Tecken in one shell, ran the systemtests in another shell, and then checked to see what redis had:

```
$ make redis-cache-cli
redis-cache:6379> info keyspace
# Keyspace
db0:keys=7,expires=7,avg_ttl=21593849
redis-cache:6379> select 0
OK
redis-cache:6379> KEYS *
1) ":1:6eb57350779c084fa11f712931db689f"
2) ":1:238cfebe2bac7e5fd4160c3a6f1dda0e"
3) ":1:f68e23336fc497b4e2fba99e97a14b89"
4) ":1:3cfae37282a0b1a6a451d4daab3bb393"
5) ":1:426605a2b8e40636d771f23dda459192"
6) ":1:acc9a72ca617c373b13741d7205ec4f1"
7) ":1:87ba954803e3624a4dfaa872c8ea4d7e"
```

I then ran the tests in another shell and rechecked db 0 to see if the keys were still there:

```
redis-cache:6379> KEYS *
1) ":1:6eb57350779c084fa11f712931db689f"
2) ":1:238cfebe2bac7e5fd4160c3a6f1dda0e"
3) ":1:f68e23336fc497b4e2fba99e97a14b89"
4) ":1:3cfae37282a0b1a6a451d4daab3bb393"
5) ":1:426605a2b8e40636d771f23dda459192"
6) ":1:acc9a72ca617c373b13741d7205ec4f1"
7) ":1:87ba954803e3624a4dfaa872c8ea4d7e"
```

I think that means it's good. Thank you!

PR #2857, opened 2023-12-18
Add VS Code dev container configuration.

smarnach:
This PR adds a dev container configuration for VS Code. Features:

* All dependencies are installed in the dev container, so IntelliSense fully works, and type checking will also work once we add type annotations.
* Allows running the tests using [VS Code's Python test browser](https://code.visualstudio.com/docs/python/testing), which makes running and debugging tests really convenient. (This is the main reason to use a container instead of a virtualenv in the first place.)
* Sets up Black as the default Python formatter. We can consider adding a setting to automatically format Python code on save.

I'm not really an expert on VS Code, so it's possible there are better ways of setting this up. This configuration works well for me, though, and it might be useful for others as well.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,10 @@
+FROM mcr.microsoft.com/devcontainers/python:3.11-bullseye@sha256:105bf6a63ab025206f019a371a735fec6553db0be520030c7a2fd0e002947232
+
+WORKDIR /app
+
+# Install Python dependencies
+COPY requirements.txt /app/
+RUN pip install -U 'pip>=20' && \
+    pip install --no-cache-dir --no-deps -r requirements.txt && \
+    pip install --no-cache-dir ipython && \
+    pip check --disable-pip-version-check
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,4 +1,5 @@
 {
+  "name": "Tecken",
   "dockerComposeFile": [
     "../docker-compose.yml",
     "docker-compose.yml"
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,26 @@
+{
+  "dockerComposeFile": [
+    "../docker-compose.yml",
+    "docker-compose.yml"
+  ],
+  "service": "devcontainer",
+  "runServices": [
+    "devcontainer"
+  ],
+  "shutdownAction": "none",
+  "workspaceFolder": "/app",
+  "customizations": {
+    "vscode": {
+      "settings": {
+        "python.testing.unittestEnabled": false,
+        "python.testing.pytestEnabled": true,
+        "[python]": {
+          "editor.defaultFormatter": "ms-python.black-formatter"
+        }
+      },
+      "extensions": [
+        "ms-python.black-formatter"
+      ]
+    }
+  }
+}
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -8,6 +8,7 @@ services:
       service: test
     build:
       dockerfile: .devcontainer/Dockerfile
+    image: tecken-devcontainer
     links:
       - db
       - fakesentry
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -1,0 +1,18 @@
+---
+# Dev container https://containers.dev/, e.g. for VS Code
+version: '2.4'
+services:
+  devcontainer:
+    extends:
+      file: "docker-compose.yml"
+      service: test
+    build:
+      dockerfile: .devcontainer/Dockerfile
+    links:
+      - db
+      - fakesentry
+      - redis-cache
+      - localstack
+      - oidcprovider
+      - statsd
+    entrypoint: ["sleep", "inf"]
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,5 @@
+{
+    "recommendations": [
+        "ms-vscode-remote.remote-containers"
+    ]
+}
--- a/docs/dev.rst
+++ b/docs/dev.rst
@@ -101,6 +101,7 @@ should be prompted whether you want to reopen the folder in a container on
 startup. You can also use the "Dev containers: Reopen in container" command
 from the command palette. The container has all Python requirements installed.
 IntelliSense, type checking, code formatting with Black and running the tests
+from the test browser are all set up to work without further configuration.
 
 What services are running in a local dev environment
 ----------------------------------------------------
--- a/docs/dev.rst
+++ b/docs/dev.rst
@@ -91,6 +91,17 @@ The Symbols Service webapp is at: http://localhost:3000
 How to
 ======
 
+How to set up a development container for VS Code
+-------------------------------------------------
+
+The repository contains configuration files to build a
+`development container <https://containers.dev/>`_ in the `.devcontainer`
+directory. If you have the "Dev Containers" extension installed in VS Code, you
+should be prompted whether you want to reopen the folder in a container on
+startup. You can also use the "Dev containers: Reopen in container" command
+from the command palette. The container has all Python requirements installed.
+IntelliSense, type checking, code formatting with Black and running the tests
+
 What services are running in a local dev environment
 ----------------------------------------------------
 


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM mcr.microsoft.com/devcontainers/python:3.11-bullseye@sha256:105bf6a63ab025206f019a371a735fec6553db0be520030c7a2fd0e002947232
+
+WORKDIR /app
+
+# Install Python dependencies
+COPY requirements.txt /app/
+RUN pip install -U 'pip>=20' && \
+    pip install --no-cache-dir --no-deps -r requirements.txt && \
+    pip install --no-cache-dir ipython && \
+    pip check --disable-pip-version-check
```
biancadanforth, line 10: I'm taking this as an opportunity to learn some of the Python stuff as well (hope you don't mind 😅 ).

I understand this is largely based on `docker/Dockerfile`, I think I grok most of the options here except the `--no-deps` one.

IIUC, this means [don't do dependency resolution](https://pip.pypa.io/en/stable/user_guide/#how-to-upgrade-and-migrate) -- because pip might throw some version icompatibility error when in reality the deps work together just fine?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM mcr.microsoft.com/devcontainers/python:3.11-bullseye@sha256:105bf6a63ab025206f019a371a735fec6553db0be520030c7a2fd0e002947232
+
```
biancadanforth, line 2: In an earlier version of this patch, this also had some Python env vars set up and an `EXPOSE`, similar to the main `docker/Dockerfile`.
```
ENV PYTHONUNBUFFERED=1 \
    PYTHONDONTWRITEBYTECODE=1 \
    PYTHONPATH=/app/ \
    PORT=8000

EXPOSE $PORT
```

Since we're basing this container off the `test` container based on `.devcontainer/docker-compose.yml`, is this no longer necessary?

I can confirm that when I open a shell in the dev container, that these env vars are indeed set.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM mcr.microsoft.com/devcontainers/python:3.11-bullseye@sha256:105bf6a63ab025206f019a371a735fec6553db0be520030c7a2fd0e002947232
```
biancadanforth, line 1: These appear to be hosted by Microsoft instead of on Docker Hub. I'm having trouble finding this exact SHA, though I found the list of Dev Container Python image tags: https://mcr.microsoft.com/v2/devcontainers/python/tags/list. Where can I find this?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,26 @@
+{
+  "dockerComposeFile": [
+    "../docker-compose.yml",
+    "docker-compose.yml"
+  ],
+  "service": "devcontainer",
+  "runServices": [
+    "devcontainer"
+  ],
+  "shutdownAction": "none",
```
biancadanforth, line 10: This is what prevents the devcontainer from shutting down when we close out of VS Code IIUC.

.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+    build:
+      dockerfile: .devcontainer/Dockerfile
+    links:
+      - db
+      - fakesentry
+      - redis-cache
+      - localstack
+      - oidcprovider
+      - statsd
+    entrypoint: ["sleep", "inf"]
```
biancadanforth, line 18: IIUC, normally docker containers live until their commands finish running and then shutdown. This is saying to keep the container running indefinitely.

biancadanforth (commented):
Thanks Sven for these improvements! Like you, I am not a VS Code expert, but I agree this is a nice improvement for devs using VS Code.

I do have one request to add a small docs section for VS Code on using the Dev Containers extension. Perhaps we can add a `.vscode/extensions.json` file with a `recommendations` field pointing to the Dev Containers extension as well (referencing [something similar](https://github.com/mozilla/fxa/blob/4c0a43b64acaae917062768faf6488e00fc2d1ab/.vscode/extensions.json#L1) I remember as a dev for FxA)?

.devcontainer/Dockerfile
relud: https://mcr.microsoft.com/en-us/product/devcontainers/python/tags and search for the sha

.devcontainer/Dockerfile
relud: Packages on pypi can be changed in place, which could introduce a new dependency with code that we don't trust. For this reason it is sometimes considered good practice to set `--no-deps` to ensure no unverified dependencies are installed.

If a transitive dependency is missing from requirements.txt and requirements.txt contains hashes, then pip will refuse to install anything, because if one dependency has a hash set all dependencies must have a hash set. The most common reason I've seen for a missing transitive dependency is using different python versions, because packages can set version-specific dependencies, and pip-compile will only include transitive dependencies based on the current python version.

Another concrete reason to use `--no-deps` is that [pip has a bug with version resolution and extras](https://github.com/pypa/pip/issues/9644) that has been haunting me for years. if your requirements file contains hashes, and in your dependency tree you have something like `a` requires `c`, `b` requires `c[d]`, and your requirements file looks like:
```
a==1.0 --hash[...]
b==1.0 --hash[...]
c[d]==1.0 --hash[...]
```
then pip will fail because the requirements file doesn't contain a hash for `c` (without extras). using `--no-deps` fixes the issue. this impacts certain gcp sdk libraries that require something like either `sdk-common` or `sdk-common[protobuf]`.

And last but not least, because we expect requirements.txt to contain all transitive dependencies and pip-compile enforces this behavior, so `--no-deps` speeds things up because we don't need to check for deps.

.devcontainer/Dockerfile
relud: I don't see any of those 4 env vars set in my dev container, and i don't see any exposed ports for it either

.devcontainer/Dockerfile
smarnach: Thanks for the details, relud! I thought the reason for `--no-deps` was mainly the last thing you mentioned – we use pip-compile to resolve dependencies, so there is no reason to have pip resolve them _again_. But I remember that bug you mention, and it's definitely another good reason.

Installing IPython here is kind of a hack. If we ever want more dev-container-specific dependencies, we can figure out a better way of installing them.

.devcontainer/Dockerfile
smarnach: I originally simply copied [the environment variables from the Tecken Dockerfile](https://github.com/mozilla-services/tecken/blob/ddde451cff69026e8f6c479e6242dc051c6578d8/docker/Dockerfile#L17-L20) into this Dockerfile, but when Bianca asked me why I did that I realized there is no good reason, so I removed them again.

.devcontainer/Dockerfile
smarnach: To look this up, I of course had to do the reverse search. I went to https://mcr.microsoft.com/en-us/product/devcontainers/python/tags and searched for the `3.11-bullseye` tag.

.devcontainer/devcontainer.json
smarnach: Yes, because we had issues when it shut the container down.

.devcontainer/docker-compose.yml
smarnach: Something like this, though I'm not sure about the exact definition of the "lifetime" of a container. The container kind of keeps existing after the command finished (unless you use `--rm` or something similar on the command line), but it's not _running_ anymore. VS Code requires the container to be running, and this ensure it keeps running until it's shut down.

.vscode/extensions.json
```diff
@@ -0,0 +1,5 @@
+{
+    "recommendations": [
+        "ms-vscode-remote.remote-containers"
```
relud, line 3: I think this should include the black extension referenced in the configuration:
```suggestion
        "ms-vscode-remote.remote-containers",
        "ms-python.black-formatter"
```

.devcontainer/Dockerfile
biancadanforth: Thank you both for the information. So much behind one ostensibly simple option.

> Packages on pypi can be changed in place
Can you clarify what you mean here? E.g. if we specify `latest` or otherwise have not pinned a dependency at a specific version in `requirements.in`, and a package update is available?

> For this reason it is sometimes considered good practice to set --no-deps to ensure no unverified dependencies are installed.
What does unverified mean in this case?

.vscode/extensions.json
biancadanforth: Well the `black` extension gets installed in the `devcontainer` as [part of its VSCode config](https://github.com/mozilla-services/tecken/pull/2857/files#diff-24ad71c8613ddcf6fd23818cb3bb477a1fb6d83af4550b0bad43099813088686R18), so that should already be covered.

docs/dev.rst
```diff
@@ -91,6 +91,17 @@ The Symbols Service webapp is at: http://localhost:3000
...
+How to set up a development container for VS Code
+-------------------------------------------------
+
+The repository contains configuration files to build a
+`development container <https://containers.dev/>`_ in the `.devcontainer`
+directory. If you have the "Dev Containers" extension installed in VS Code, you
+should be prompted whether you want to reopen the folder in a container on
+startup. You can also use the "Dev containers: Reopen in container" command
+from the command palette. The container has all Python requirements installed.
+IntelliSense, type checking, code formatting with Black and running the tests
```
biancadanforth, line 103: Nit: Maybe the line starting with "IntelliSense, ..." should be part of the previous sentence? Seems like it's an incomplete sentence on its own at the moment.

docs/dev.rst
```diff
@@ -91,6 +91,17 @@ The Symbols Service webapp is at: http://localhost:3000
...
 ======
 
+How to set up a development container for VS Code
+-------------------------------------------------
+
+The repository contains configuration files to build a
+`development container <https://containers.dev/>`_ in the `.devcontainer`
+directory. If you have the "Dev Containers" extension installed in VS Code, you
+should be prompted whether you want to reopen the folder in a container on
+startup. You can also use the "Dev containers: Reopen in container" command
```
biancadanforth, line 101: Might be worth noting that they will likely need to restart VS Code after they install Dev Containers for the first time (as @relud mentioned they had issues opening the terminal for example on the first run until they did that).

Also even before that, might be worth mentioning that Tecken should be running locally first before any of this will work (might be obvious, but I prefer to overcommunicate generally).

biancadanforth (approved):
LGTM! Made a couple of small suggestions on the docs. Thank you.

.devcontainer/Dockerfile
relud: > Packages on pypi can be changed in place

if i have already published version 1.1.1, I can modify the packages hosted on pypi at any time, and the only indication will be different hash values. One reason why we have `requirements.in` and `requirements.txt` is so that we can use `pip-compile --generate-hashes` to add `--hash` arguments for each package in `requirements.txt`, so that if a package is modified we get an error instead of installing potentially malicious code.

>> set --no-deps to ensure no unverified dependencies are installed.
>
> What does unverified mean in this case?

I was using verified to mean "matches a package in `requirements.txt`".

I should have worded this better, because really it's having hashes in `requirements.txt` that provides most of the protection here. `--no-deps` would only be providing any protection if we didn't include hashes in `requirements.txt` and thus could install dependencies not present in `requirements.txt`, and then it would only protect against malicious code in those dependencies and not malicious code added to an existing dependency.

In theory before merging any dependabot PRs or adding any new dependencies directly, we could be checking the content of every package hash added to `requirements.txt` to make sure no malicious code has been included, and ~7 years ago i was advised to do that once by the data org python expert at the time. In practice I have not seen a project do this, and just using hashes is usually considered sufficient protection.

.vscode/extensions.json
relud: ah, i didn't realize the extensions were sandboxed like that, ignore me

docs/dev.rst
relud: > Tecken should be running locally first before any of this will work

in particular I had to run `docker compose -f docker-compose.yml -f .devcontainer/docker-compose.yml build` before this worked for me, but I didn't actually have to have it running, that part worked automatically.

biancadanforth: Ah right, I was thinking about previously where we were connected to the `tecken-web` container that Tecken had to be running locally. In this case, this is a completely separate container that VS Code is starting up from the other Tecken containers, so Tecken doesn't need to be running (unless you want to run some integration or end to end tests that require communicating with other containers).

smarnach: It shouldn't actually be necessary to build the other containers manually first – at least that's my understanding how it is supposed to work, which could very well be wrong. I'll do a few tests before adding any further documentation.

docs/dev.rst
smarnach: This was meant to be a new sentence, but apparently I just forgot mid-sentence that I was writing a sentence.

smarnach:
I found a bug in the configuration. Since the `devcontainer` service inherits from the `test` service, it also inherits the image name `tecken:build` from that service. As a result, building the dev container overrides the tag for the Tecken image, and if you stop and restart the whole stack, the dev container is used for any service inheriting from `base`. I fixed this by using a different image name for the dev conatiner.

PR #2858, opened 2023-12-18
Make ports exposed in docker compose configurable

relud:
see also https://github.com/mozilla-services/eliot/pull/113

per https://github.com/mozilla-services/tecken/pull/2857/files#r1430382049
> ```
>  "shutdownAction": "none",
> ```
> [biancadanforth](https://github.com/biancadanforth) [Dec 18, 2023](https://github.com/mozilla-services/tecken/pull/2857/files#r1430382049):
>  This is what prevents the devcontainer from shutting down when we close out of VS Code IIUC.

with devcontainer and its docker-compose dependencies running even when vscode isn't open, ports conflict between tecken, eliot, and socorro, so this PR makes the exposed ports configurable via environment variables, which can be set in `.env` in each project to avoid overlap.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,7 +20,7 @@ services:
       - db
       - redis-cache
     ports:
-      - "8000:8000"
+      - "${EXPOSE_TECKEN_PORT:-8000}:8000"
     links:
       - db
       - fakesentry
@@ -62,7 +62,7 @@ services:
     extends:
       service: base
     ports:
-      - "8000:8000"
+      - "${EXPOSE_TECKEN_PORT:-8000}:8000"
     links:
       - db
       - redis-cache
@@ -108,7 +108,7 @@ services:
     environment:
       - LOCALSTACK_HOST=localstack
     ports:
-      - "4566:4566"
+      - "${EXPOSE_LOCALSTACK_PORT:-4566}:4566"
 
   # https://github.com/willkg/kent
   fakesentry:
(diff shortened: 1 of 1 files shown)


willkg (approved):
`${EXPOSE_REVIEW_APPROVED}`

docker-compose.yml
```diff
@@ -108,15 +108,15 @@ services:
     environment:
       - LOCALSTACK_HOST=localstack
     ports:
-      - "4566:4566"
+      - "${EXPOSE_LOCALSTACK_PORT:-4566}:4566"
```
smarnach, line 111: How do you want to set the environment variables? If you set them in `.env`, they will only be used by the `Makefile`, and not if VS Code brings the container up. If you set them in your `.bashrc`, then all the variable names should contain `TECKEN`, so they don't conflict with variable names for other services.

relud: docker compose automatically picks up variables from .env https://docs.docker.com/compose/environment-variables/set-environment-variables/#substitute-with-an-env-file

I made sure it works by exiting vscode, running `docker compose -f docker-compose.yml -f .devcontainer/docker-compose.yml down`, opening vscode, waiting for the devcontainer to finish starting, and finally checking the exposed ports via `docker compose -f docker-compose.yml -f .devcontainer/docker-compose.yml ps`

smarnach: Oh, nice! I didn't know. We should make use of that for the build arguments. They are currently passed in explicitly from the `Makefile`, but we should add them to `docker-compose.yml` instead.

PR #2859, opened 2023-12-19
bug-1743835: Upgrade target Python versions for black and ruff to 3.11.

smarnach:
This also upgrades ruff and black to their respective lateste versions. The black upgrade pulls in a few new dependencies, most notably and somewhat surprisingly aiohttp with all its dependencies.


requirements.txt
```diff
@@ -4,6 +4,88 @@
...
+    --hash=sha256:d7481f581251bb5558ba9f635db70908819caa221fc79ee52a7f58392778c636 \
+    --hash=sha256:df9cf74b9bc03d586fc53ba470828d7b77ce51b0582d1d0b5b2fb673c0baa32d \
+    --hash=sha256:e1f80197f8b0b846a8d5cf7b7ec6084493950d0882cc5537fb7b96a69e3c8590 \
+    --hash=sha256:ecca113f19d5e74048c001934045a2b9368d77b0b17691d905af18bd1c21275e \
+    --hash=sha256:ee2527134f95e106cc1653e9ac78846f3a2ec1004cf20ef4e02038035a74544d \
+    --hash=sha256:f27fdaadce22f2ef950fc10dcdf8048407c3b42b73779e48a4e76b3c35bca26c \
+    --hash=sha256:f694dc8a6a3112059258a725a4ebe9acac5fe62f11c77ac4dcf896edfa78ca28 \
+    --hash=sha256:f800164276eec54e0af5c99feb9494c295118fc10a11b997bbb1348ba1a52065 \
+    --hash=sha256:ffcd828e37dc219a72c9012ec44ad2e7e3066bec6ff3aaa19e7d435dbf4032ca
+    # via black
```
relud, line 84: why is aiohttp being included when we aren't including the extra `black[d]`?
https://github.com/psf/black/blob/d9ad09a32b0e0481bb4fef548d35b7a49cc03c5d/pyproject.toml#L78-L81

relud: to be clear, this isn't on us, i'll file an upstream bug if i can't figure it out myself

relud (approved):

requirements.txt
relud: filed https://github.com/psf/black/issues/4120 because it doesn't look intentional on black's part

smarnach:
Thanks for the review, and for digging into why the dependencies got installed. We could go back to 23.11.0 until next month's release, but I don't think the unneeded dependencies cause any problems, so I'll just merge this as is.

requirements.txt
biancadanforth: Nice catch!

PR #2860, opened 2023-12-21
Remove unused code.

smarnach:
This removes some random unused code I encountered while reading the code. This isn't the result of any kind of systematic survey, so there's probably more unused code.

### DIFF
--- a/tecken/api/forms.py
+++ b/tecken/api/forms.py
@@ -8,7 +8,7 @@ import re
 import dateutil
 
 from django import forms
-from django.contrib.auth.models import Permission, Group, User
+from django.contrib.auth.models import Permission, User
 from django.utils import timezone
 
 
@@ -62,24 +62,6 @@ class TokensForm(forms.Form):
         if value and value not in ("expired", "all"):
             raise forms.ValidationError(f"Unrecognized state value {value!r}")
         return value
-
-
-class UserEditForm(forms.ModelForm):
-    groups = forms.CharField(required=False)
-
-    class Meta:
-        model = User
-        fields = ("is_active", "is_superuser")
-
-    def clean_groups(self):
-        value = self.cleaned_data["groups"]
-        groups = []
-        for pk in [x for x in value.split(",") if x.strip()]:
-            try:
-                groups.append(Group.objects.get(id=pk))
-            except ValueError as exc:
-                raise forms.ValidationError("Invalid group ID") from exc
-        return groups
 
 
 class BaseFilteringForm(forms.Form):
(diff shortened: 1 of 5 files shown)


tecken/upload/models.py
```diff
@@ -12,10 +12,6 @@
 logger = logging.getLogger("tecken")
 
 
-class SumCardinality(models.Aggregate):
-    template = "SUM(CARDINALITY(%(expressions)s))"
```
willkg, line 16: This was used by a page I removed a while back as I was paring down the web ui.

tecken/api/views.py
```diff
@@ -39,10 +39,6 @@
 BIG_NUMBER = 1_000_000
 
 
-class SumCardinality(Aggregate):
-    template = "SUM(CARDINALITY(%(expressions)s))"
```
willkg, line 43: This was used by a page I removed a while back as I was paring down the web ui.

tecken/api/forms.py
```diff
@@ -64,24 +64,6 @@ def clean_state(self):
         return value
 
 
-class UserEditForm(forms.ModelForm):
```
willkg, line 67: I dropped the web ui user admin in favor of the Django admin user pages to reduce the maintenance burden.

tecken/base/symboldownloader.py
```diff
@@ -22,16 +22,6 @@
...
 
 
-class SymbolNotFound(Exception):
-    """Happens when you try to download a symbols file that doesn't exist"""
-
-
-class SymbolDownloadError(Exception):
-    def __init__(self, status_code, url):
-        self.status_code = status_code
-        self.url = url
```
willkg, line 32: I split the symbolication API out into a separate service (Eliot). This is vestiges from the old symbolication code.

tecken/storage.py
```diff
@@ -113,19 +113,6 @@ def __repr__(self):
...
-        """Return a backend-specific client, cached on first access.
-
-        TODO(jwhitlock): Build up StorageBucket API so users don't work directly with
-        the backend-specific clients (bug 1564452).
-        """
-        if not getattr(self, "_client", None):
-            self._client = get_storage_client(
-                endpoint_url=self.endpoint_url, region_name=self.region
-            )
-        return self._client
```
willkg, line 127: This was last used in the code that supported private symbols buckets that weren't accessible via HTTP which I removed in July.

willkg (approved):
I went through and figured out how these bits became unused. Everything looks good! Thank you for cleaning this up!

PR #2862, opened 2023-12-21
bug-1870254: ignore FileNotFoundError in remove_orphaned_files

relud:


tecken/tests/test_upload.py
```diff
@@ -1093,25 +1093,20 @@ def test_errors(
...
         assert contents == ["/upload1", "/upload1/file1.sym"]
 
-        # Monkeypatch os.path.getmtime to get the mtime, delete the file, and return the
-        # mtime so as to simulate a race condition between getting the mtime and getting
-        # the size
-        original_getmtime = os.path.getmtime
+        # Monkeypatch os.path.getsize to throw an OSError so as to simulate a race
+        # condition between getting the mtime and getting the size
+        def adjusted_getsize(fn):
+            raise OSError()
```
willkg, line 1099: This is a very different monkeypatch than the one we had originally and changes the semantics of the test. This version tests what happens in the code if `getsize` raises an `OSError`. It doesn't set up the race condition and doesn't test the scenario you're fixing.

I think I would rename this test to `test_filenotfound_racecondition` and change the assertion to what we're hoping for--namely that nothing gets logged. Then I'd write a new test called `test_oserror` that asserts something gets logged if an `OSError` is raised.

willkg (changes requested):
I'd change the tests, but otherwise the fix looks good. Thank you!

willkg (approved):
Looks good!

PR #2867, opened 2024-01-03
bug-1864651: add host tag to all emitted metrics

relud:


tecken/libmarkus.py
```diff
@@ -6,8 +6,26 @@
 import time
 from functools import partialmethod
 
+import markus
 from markus import INCR, GAUGE, HISTOGRAM, TIMING
 from markus.backends import BackendBase
+from markus.filters import AddTagFilter
+
+_IS_MARKUS_SETUP = False
+METRICS = markus.get_metrics("tecken")
```
relud, line 15: this has to be imported and reused everywhere instead of re-calling `markus.get_metrics("tecken")` in order for the AddTagFilter to apply

willkg: You can specify it as part of the backend configuration, too. We do that in Eliot:

https://github.com/mozilla-services/eliot/blob/82bcae05b11400c72c4299709690357788003bc1/eliot/libmarkus.py#L181

relud: I don't see any reference to a host tag in the backend configuration there, except when eliot uses its shared `METRICS` constant the same way this PR does:

https://github.com/mozilla-services/eliot/blob/82bcae05b11400c72c4299709690357788003bc1/eliot/libmarkus.py#L226:
```
        METRICS.filters.append(AddTagFilter(f"host:{hostname}"))
```

relud: per slack [@willkg 15 minutes ago](https://mozilla.slack.com/archives/C060L2ZA0DT/p1704306428334699?thread_ts=1704306007.642629&cid=C060L2ZA0DT)
> [@relud](https://mozilla.slack.com/team/U5R388Y5A) on second thought, keep what you have with the way the filter stuff is implemented. i bet i did it this way in eliot because it was hard to verify that the filter kicks in.

tecken/tests/test_sentry.py
```diff
@@ -221,8 +220,9 @@ def test_sentry_scrubbing(sentry_helper, transactional_db):
         assert differences == []
 
 
-def test_count_sentry_scrub_error():
-    with MetricsMock() as metricsmock:
```
willkg, line 225: Weird. I wonder why I used the contextmanager mode rather than the pytest fixture. Thank you for fixing this!

willkg (changes requested):
I think we missed these files:

```
tecken/upload/utils.py
26:metrics = markus.get_metrics("tecken")

tecken/upload/views.py
43:metrics = markus.get_metrics("tecken")
```

Can you update those files with the METRICS thing?

relud:
updated those locations, now the only files that import markus are:
```
tecken/libmarkus.py
systemtests/bin/upload-symbols-by-download.py
systemtests/bin/upload-symbols.py
systemtests/bin/download-sym-files.py
```
and I believe the systemtests are correctly not using the tecken metrics instance.

willkg (approved):
Looks super, plus I see my banana now:

```
tecken-web-1           | INFO 2024-01-03 21:13:29,326 - webapp - markus METRICS|2024-01-03 21:13:29|timing|tecken.upload_dump_and_extract|1533.210732974112|#host:banana
```

tecken/libmarkus.py
biancadanforth: > this has to be imported and reused everywhere instead of re-calling markus.get_metrics("tecken") in order for the AddTagFilter to apply

Late to the game here, but this is because the tag is tied to that instance of the `MetricsInterface` returned by `markus.get_metrics("tecken")` -- so we have to use the same instance we added the tag to when the app starts up via `_configure_markus`?

relud: that is my understanding

PR #2868, opened 2024-01-03
bug-1827507: implement GCP dev environment

biancadanforth:
This PR:
- [ ] Adds a new `gcs-emulator` service, linking it where `localstack` is also linked
- [ ] Creates a new GCS CLI for local dev with the same commands as the S3 CLI
  - [ ] create
  - [ ] delete
  - [ ] list buckets
  - [ ] list objects 
- [ ] Calls the new CLI in `make setup`
- [ ] Updates docs


bin/gcs_cli.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+@click.group()
+def gcs_group():
+    """Local dev environment GCS manipulation script"""
+
+
+@gcs_group.command("create")
+@click.argument("bucket_name")
+def create_bucket(bucket_name):
```
biancadanforth, line 42: I'm not passing in the `ctx` object here as in `bin/s3_cli.py`, as I don't believe it is needed, though my understand per the `click` docs is fuzzy: https://click.palletsprojects.com/en/8.1.x/complex/#contexts . Is this needed?

bin/gcs_cli.py
```diff
@@ -0,0 +1,106 @@
...
+    endpoint_url = f"{base_url}/storage/v1/b/{bucket_name}"
+
+    client = get_client()
+
+    try:
+        client.get_bucket(bucket_name)
+        # TODO Bucket already exists, but do we have access to it?
+        # s3 client has a head_bucket method which also checks access permissions:
+        # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/head_bucket.html#head-bucket
+        # Address this for all uses of client.get_bucket in this script.
```
biancadanforth, line 58: This method isn't exactly equivalent to the S3 client's `head_bucket` method in that it doesn't check permissions. Given that this is an emulator, I don't think ACLs really apply anyway, so perhaps that's fine? That's what this TODO is about.

bin/setup-services.sh
```diff
@@ -10,6 +10,14 @@ set -eo pipefail
 python bin/s3_cli.py delete "${UPLOAD_DEFAULT_URL}"
 python bin/s3_cli.py create "${UPLOAD_DEFAULT_URL}"
 
+# Set up GCS
+# FIXME bug 1827506: conditionally set UPLOAD_DEFAULT_URL TO GCS URL
+# (i.e. https://gcs-emulator:4443/storage/v1/b/{bucket_name})
+# when CLOUD_SERVICE_PROVIDER == "GCP" and pass that in here instead
+# of publicbucket.
+python bin/gcs_cli.py delete publicbucket
+python bin/gcs_cli.py create publicbucket
```
biancadanforth, line 19: Currently, I'm constructing the `endpoint_url` within `gcs_cli.py` and only accepting a bucket name here instead of either a bucket name or a URL.

Passing in a URL here is a little confusing to me, though it looks like for the S3 CLI, a URL is passed, and this URL is used throughout the source code. I thought I could leave this as-is for now, and perhaps this can be revisited once the [storage API is refactored](https://bugzilla.mozilla.org/show_bug.cgi?id=1564452), and once we [add GCS](https://bugzilla.mozilla.org/show_bug.cgi?id=1827506) into the source code. What do you think? I can make the code comment less prescriptive.

biancadanforth (commented):
@willkg (or others!): I am not done here, but I have enough of a WIP and a few questions I thought I'd put what I've got out there for thoughts.

docker-compose.yml
```diff
@@ -108,6 +110,13 @@ services:
   redis-cache:
     image: redis:6.2.6
 
+  # https://github.com/fsouza/fake-gcs-server
+  # Fake GCP GCS server for local development and testing
+  gcs-emulator:
+    image: fsouza/fake-gcs-server:latest
+    ports:
+      - "${EXPOSE_GCS_EMULATOR_PORT:-5553}:4443"
```
biancadanforth, line 118: Strangely port 4443 for the host was already in use by VS Code Dev Containers once the `devcontainer` service was running in the `tecken` workspace, so without using a different host port, requests to the emulator service were just hanging.

bin/gcs_cli.py
willkg: I add ctx because I often use `ctx.exit(1)` in error situations and I'm just used to setting up click command code that way. If you don't need to do that or use other things in `ctx`, then you don't need it.

bin/gcs_cli.py
willkg: I'm not entirely sure what you mean by this. If `get_bucket` verifies that the bucket was created, then that's good. If it doesn't do that, then we need to do something additional to make sure the bucket was created like list the contents of the bucket.

I don't think I'd worry about testing permissions the client has to the bucket.

bin/setup-services.sh
willkg: @smarnach ^^^ What do you prefer?

docker-compose.yml
willkg: @relud ^^^ Can you weigh in on port usage?

relud: i have no answers, setting this to 4443 seems to be working for me 🤷

bin/gcs_cli.py
biancadanforth: Okay thanks. Yes this `get_bucket` method will throw a `NotFound` error if the bucket doesn't exist, and only in that case do we call `create_bucket` in the `except` block downstream. [`create_bucket` will error](https://cloud.google.com/storage/docs/json_api/v1/buckets/insert) if it cannot create the bucket, or if the bucket already exists.

docker-compose.yml
relud: setup:
```bash
git checkout bug-1827507
echo 'EXPOSE_GCS_EMULATOR_PORT=4443' >> .env
docker compose up gcs-emulator -d
```
how i confirmed it was working:
```bash
$ STORAGE_EMULATOR_HOST=https://localhost:4443 venv/bin/python bin/gcs_cli.py create relud
GCS bucket 'relud' in 'https://localhost:4443/storage/v1/b/relud' created.
$ STORAGE_EMULATOR_HOST=https://localhost:4443 venv/bin/python bin/gcs_cli.py delete relud
GCS bucket 'relud' at 'https://localhost:4443/storage/v1/b/relud' deleted.
```

this was with vscode and devcontainer running the whole time

relud: within my devcontainer i was able to access it with:
```bash
$ STORAGE_EMULATOR_HOST=https://gcs-emulator:4443 python bin/gcs_cli.py create relud
GCS bucket 'relud' in 'https://gcs-emulator:4443/storage/v1/b/relud' created.
$ STORAGE_EMULATOR_HOST=https://gcs-emulator:4443 python bin/gcs_cli.py delete relud
GCS bucket 'relud' at 'https://gcs-emulator:4443/storage/v1/b/relud' deleted.
```

biancadanforth: Thanks for the detailed steps! Testing this the first way is what hangs for me.
```bash
(venv) bdanforth ~/Projects/tecken (bug-1827507) $ STORAGE_EMULATOR_HOST=https://localhost:4443 venv/bin/python bin/gcs_cli.py create relud
```
If I wait long enough, I get a deadline exceeded error from the Google Cloud client library. The request to the emulator is timing out. Sure enough I have the docker host and VS Code listening on the same port when I check via `lsof` and friends.

biancadanforth: I don't consider this thread blocking, so I'm going to mark it as resolved for now to reduce noise. I reverted the default host port to 4443.

bin/setup-services.sh
biancadanforth: I cleaned up the comment to be less prescriptive in the meantime. Other than this detail, the patch is ready for review, so I've marked it as ready!

bin/gcs_cli.py
```diff
@@ -0,0 +1,137 @@
...
+def list_buckets(details):
+    """List GCS buckets"""
+
+    client = get_client()
+
+    buckets = client.list_buckets()
+    for bucket in buckets:
+        if details:
+            # https://cloud.google.com/storage/docs/json_api/v1/buckets#resource-representations
+            click.echo(f"{bucket.name}\t{bucket.time_created}")
```
biancadanforth, line 104: Compared to AWS S3, there is [a lot more data we could print for the bucket](https://cloud.google.com/storage/docs/json_api/v1/buckets#resource-representations). Is there more information that would be helpful?

bin/gcs_cli.py
```diff
@@ -0,0 +1,137 @@
...
+    client = get_client()
+
+    try:
+        client.get_bucket(bucket_name)
+    except NotFound:
+        click.echo(f"GCS bucket {bucket_name!r} does not exist.")
+        return
+
+    blobs = client.list_blobs(bucket_name)
+    is_empty = True
```
biancadanforth, line 124: I couldn't find a great way to check if the iterator returned by `list_blobs` is empty, so I opted for a simple flag. Open to alternatives here if there's a better/clearer way.

biancadanforth (commented):
Leaving some questions in-line.

I also noted in the description that there are no tests, but tests for a simple, dev-only tool are likely overkill? I didn't see any tests for the analogous `s3_cli.py` script. Happy to improve upon that though if it seems beneficial.

willkg:
Re: tests for `s3_cli.py`, I didn't write any because it's such a critical part of the dev environment it would be obvious when it broke and I had been minimizing the amount of code I needed to maintain. Things are different now, so maybe we want to hedge less towards "barebones development" and more towards correctness detection or something like that?

Beyond that, this looks fine to me if it meets the needs you and @smarnach have for GCP migration development.

bin/setup-services.sh
smarnach: I don't see an advantage to using a URL here, so let's keep it the way it is. As Bianca already said, we can always change it later if we need to.

smarnach (approved):

PR #2870, opened 2024-01-05
Add tig to devcontainer

relud:
tig is a command line git tree visualization tool


smarnach (approved):
I usually run all git stuff on my host system rather than inside the container, but it's fine to support different workflows.

PR #2871, opened 2024-01-05
Add devcontainer commands to make

relud:
and additional devcontainer related docs to match https://github.com/mozilla-services/eliot/pull/122


Makefile
```diff
@@ -30,6 +30,9 @@ help:
 .docker-build:
 	make build
 
+.devcontainer-build:
+	make devcontainerbuild
```
smarnach, line 34: I don't think we need a recursive make invocation.
```suggestion
.devcontainer-build: devcontainerbuild
```

smarnach: Oh, I see, we have the same thing for `.docker-build`. Maybe there was a reason for the recursive invocation? I can't think of anything, though.

smarnach (approved):
I've always let VS Code build the dev container. Is there any reason to build it manually?

In any case, this looks good to me.

relud:
> I've always let VS Code build the dev container.

I have to manually build the container for everything to work

PR #2872, opened 2024-01-06
Configure devcontainer to use same uid/gid as docker base

relud:
While investigating issues related to https://github.com/docker/for-mac/issues/6734#issuecomment-1813380412, I found that in docker desktop for mac the mounted volume is retaining file ownership permissions between the test and devcontainer volume mounts.

For example, if vscode starts a devcontainer first, I open a terminal in vscode and run `ls -lha`, I might see that /app is owned by vscode:vscode, but if I then run `docker compose run --rm test bash` and in that container run `ls -lha` the files are owned by 1000:1000 because vscode isn't a user/group in the test container. File ownership then remains sticky for some unknown amount period after that.

This treats the problem by explicitly defining the app user/group in devcontainer with the same ids used in the test container.


relud:
@biancadanforth with this patch, i then run `chown -R app:app /app` from vscode in the devcontainer (it throws exceptions on some git read-only objects that are already owned by app:app, so that's fine), and after that `make lint` works with VirtioFS.

biancadanforth (approved):
TIL: Paraphrasing from relud: On Mac, the Docker host is a VM, and that hides the original file permissions of the `/app` directory. Since `test` and `devcontainer` both mount `$PWD/app` from the same VM, using `chown` in one applies to the VM filesystem (including the other container). Also on Mac, the default user/group id for VS Code's Dev Containers is 1000 instead of trying to match the user id/ group id of the host machine.

smarnach:
Thanks for the fix! I thought the permissions were only an issue on Linux, and since it worked fine for me, I kept it as simple as possible.

smarnach:
Unfortunately, this broke the dec container build for me:

```
 > [devcontainer dev_container_auto_added_stage_label 3/6] RUN groupadd --gid 1000 app &&     useradd -g app --uid 1000 --shell /usr/sbin/nologin --create-home app &&     chown app:app /app/:
0.402 groupadd: GID '1000' already exists
```

This is probably because my uid happens to be 1000. I'll figure out how to solve this after lunch.

PR #2875, opened 2024-01-09
Fix upload-sym.py sym_file generation

willkg:
This fixes a bug introduced in September 2023 in 23daea485d, where I fixed the code to generate the right sym_file filename for Windows, but broke it for Linux/macOS. This fixes Linux/macOS.

For Windows, the sym_file filename involves taking the debug file, dropping `.pdb`, and adding `.sym`. e.g. `libxul.pdb` -> `libxul.sym`.

For Linux/macOS, the sym_file filename involves adding `.sym` to the end of the debug file. e.g. `glxtrace.so` -> `glxtrace.so.sym` and `XUL` -> `XUL.sym`.


relud (approved):

willkg:
Thank you!

PR #2876, opened 2024-01-10
bug-1835880: switch tag pattern to vYYYY.MM.DD[-N]

biancadanforth:
Because:
* We have standardized on tags starting with v

This commit:
* Ensures the `main` job for CI is run for all branches and only tags starting with v, instead of all branches and all tags.
* Updates our `release.py` config to pass in the new tag format when creating the tag as part of a release.

Notes:
* I didn't update any docs, since I didn't see any mentions of the tag format in either the `/docs` directory in the repo or on the [Tecken page in Confluence](https://mozilla-hub.atlassian.net/wiki/spaces/CS1/pages/178061663/Symbols+aka+Tecken)[1]. Please let me know if I missed anywhere.
* I've added two new [tag protection rules](https://github.com/mozilla-services/tecken/settings/tag_protection) in the repo based on the [match patterns used by Eliot for its `build-and-push` GitHub Action](https://github.com/mozilla-services/eliot/blob/ccd40340db031584b9a498e3356bb34a5edc86ec/.github/workflows/build-and-push.yml#L8-L9).
* @willkg or @smarnach : Can one of you help me to complete this step in Jenkins per Willkg's [comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1835880#c1) in the ticket?:
 > update the deploy pipeline in jenkins which has regexps for recognizing what to do based on the tag

[1]: I could update the example output for `bin/service-status.py` in the [Verify deploy section](https://mozilla-hub.atlassian.net/wiki/spaces/CS1/pages/490602899/Runbook+Tecken+prod+deploy#Verify-deploy), but I haven't so far.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,4 +96,4 @@ workflows:
       - main:
           filters:
             tags:
-              only: /.*/
+              only: /^v.*/
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ github_project = "tecken"
 bugzilla_product = "Tecken"
 bugzilla_component = "General"
 main_branch = "main"
-tag_name_template = "%Y.%m.%d"
+tag_name_template = "v%Y.%m.%d"
 
 
 [tool.service-status]


willkg (approved):
Looks good to me.

I looked through the configuration in Jenkins and the deploy code and I don't see anything that restricts on the tag name. I don't remember why I added that step in the bug. Maybe it was because we had something in the eliot deploy code but it's not in the symbols or socorro deploy code.

willkg:
@smarnach Do you know where in the jenkins / deploy pipeline code we might have a regex for matching tag names that we need to update?

biancadanforth:
Merging per our Slack discussion, as the [open question](https://github.com/mozilla-services/tecken/pull/2876#issuecomment-1885798816) pertains to prod, and we don't expect any prod deploys in the immediate future if indeed any changes in Jenkins are required.

smarnach:
Sorry, I missed the ping here. Jenkins indeed has some code to decide what to do whenever a new image is pushed to Docker Hub. Bianca and me just updated that code with the new pattern syntax.

Unfortunately, that code only exists in the Jenkins configuration, and it's not available in any repository. Here's the new version of the code:
```
if(Tag == "latest") { // Execute stage pipeline only
	build job: 'pipelines/symbols', parameters: [
        [$class: 'StringParameterValue', name: 'ImageTag', value: Tag],
        booleanParam(name: 'DeployProd', value: false),
        booleanParam(name: 'CleanProd', value: false),
        ], wait: false
} else if (Tag ==~ /v20[0-9][0-9]\.[01][0-9]\.[0-3][0-9](-[0-9])?/ ) { // Execute stage and prod pipeline
	build job: 'pipelines/symbols', parameters: [
        [$class: 'StringParameterValue', name: 'ImageTag', value: Tag]], wait: false
}
```

The old version simply deployed to prod whenever the tag was not `latest`, so that would have worked as well, but we decided to make it a bit more predictable to avoid accidentally deploying a random other tag.

PR #2878, opened 2024-01-16
bug-1759554: remove django-cache-memoize

biancadanforth:
Because:
* This library is no longer maintained.
* The minimal caching it was doing doesn't warrant the continued maintenance burden (in particular the Redis caching backend).

This commit:
* Removes uses of the library in the source code.
* Removes it from our list of dependencies.
* Fixes tests for the affected functions.

Note: This PR does not remove the Redis config, as per [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1759554#c7) in the bug, we want to monitor some metrics in prod first before proceeding with that step.

### DIFF
--- a/tecken/base/symboldownloader.py
+++ b/tecken/base/symboldownloader.py
@@ -8,7 +8,6 @@ import time
 from typing import Optional
 from urllib.parse import quote
 
-from cache_memoize import cache_memoize
 import logging
 
 from django.conf import settings
@@ -33,11 +32,6 @@ def set_time_took(method):
     return wrapper
 
 
-@cache_memoize(
-    settings.SYMBOLDOWNLOAD_EXISTS_TTL_SECONDS,
-    hit_callable=lambda *a, **k: METRICS.incr("symboldownloader_exists_cache_hit", 1),
-    miss_callable=lambda *a, **k: METRICS.incr("symboldownloader_exists_cache_miss", 1),
-)
 @METRICS.timer_decorator("symboldownloader_exists")
 def get_last_modified(url: str) -> Optional[int]:
     """
@@ -159,7 +153,7 @@ class SymbolDownloader:
                 source.base_url, self.make_url_path(prefix, symbol, debugid, filename)
             )
             logger.debug(f"Looking for symbol file by URL {file_url!r}")
-            if last_modified := get_last_modified(file_url, _refresh=refresh_cache):
+            if last_modified := get_last_modified(file_url):
                 age_days = int(time.time() - last_modified) // 86_400  # seconds per day
                 if i == self.try_url_index:
                     tags = ["storage:try"]
--- a/tecken/base/symboldownloader.py
+++ b/tecken/base/symboldownloader.py
@@ -106,18 +106,6 @@ class SymbolDownloader:
             self._sources = list(self._get_sources())
         return self._sources
 
-    def invalidate_cache(self, symbol, debugid, filename):
-        # Because we can't know exactly which source (aka URL) was
-        # used when the key was cached by exists_in_source() we have
-        # to iterate over the source.
-        for source in self.sources:
-            prefix = source.prefix
-            assert prefix
-            file_url = "{}/{}".format(
-                source.base_url, self.make_url_path(prefix, symbol, debugid, filename)
-            )
-            get_last_modified.invalidate(file_url)
-
     @staticmethod
     def make_url_path(prefix, symbol, debugid, filename):
         """Generates a url quoted path which works with HTTP requests against AWS S3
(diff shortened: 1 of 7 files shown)


relud (approved):

tecken/base/symboldownloader.py
```diff
@@ -159,7 +153,7 @@ def _get(self, symbol, debugid, filename, refresh_cache=False):
                 source.base_url, self.make_url_path(prefix, symbol, debugid, filename)
             )
             logger.debug(f"Looking for symbol file by URL {file_url!r}")
-            if last_modified := get_last_modified(file_url, _refresh=refresh_cache):
+            if last_modified := get_last_modified(file_url):
```
relud, line 156: r+ under the assumption that `refresh_cache` arg throughout this file will be removed in a later update

willkg (changes requested):
I don't see the cache invalidation in this. I think we need to remove that, too:

https://github.com/search?q=repo%3Amozilla-services%2Ftecken%20invalidate&type=code

smarnach:
I wonder why the tests are passing with the `.invalidate()` calls still in place.

tecken/base/symboldownloader.py
biancadanforth: Thanks! I've removed them.

biancadanforth:
> I wonder why the tests are passing with the `.invalidate()` calls still in place.

If you mean the test result showing up in Willkg's [search query results](https://github.com/mozilla-services/tecken/pull/2878#pullrequestreview-1825959469) (i.e. in `tecken/tests/test_upload_utils.py`), I had fixed that test already in my original commit.

willkg:
The invalidate call is wrapped in a try/except that catches `Exception` and ignores it when DEBUG if False which is the case in test and CI environments. I think that's why it didn't kick anything up in the upload tests. :cry: 

https://github.com/mozilla-services/tecken/blob/b96476332562d4dbbb18efa9846d94c8d9d93072/tecken/upload/utils.py#L352-L373

https://github.com/mozilla-services/tecken/blob/b96476332562d4dbbb18efa9846d94c8d9d93072/docker/config/test.env#L8

tecken/upload/utils.py
```diff
@@ -23,8 +22,6 @@
 
 logger = logging.getLogger("tecken")
 
-# FIXME(willkg): This downloader needs to point to all the things that can be downloaded
-# from so it can correctly invalidate the cache
```
biancadanforth, line 27: @willkg : I'm not sure I grok what this FIXME is about, but I gather it's related to caching that we're no longer doing. Is it appropriate to remove?

biancadanforth (commented):
`refresh_cache` is still used by `cached_lookup_by_syminfo` which looks to use the default Django cache (which looks to point to a Redis backend), so IIUC, I can't remove all references to it and more importantly, we wouldn't be able to remove the Django Redis cache backend given it is still being used (granted, without `django-cache-memoize`). Is that right, @willkg ?

willkg (changes requested):
One more thing: the `_refresh` querystring param no longer does anything. We should remove that, too.

https://github.com/mozilla-services/tecken/blob/b96476332562d4dbbb18efa9846d94c8d9d93072/tecken/download/views.py#L152

https://github.com/mozilla-services/tecken/blob/b96476332562d4dbbb18efa9846d94c8d9d93072/docs/download.rst#L105

https://github.com/mozilla-services/tecken/blob/b96476332562d4dbbb18efa9846d94c8d9d93072/docs/download.rst#L174

After that, I think this is good.

**Update:** This is wrong--we should leave `_refresh` as is.

willkg:
Oh, wait--I'm confusing things. There are two caches involved here, but we need to keep the codeinfo -> syminfo one that's used in `cached_lookup_by_syminfo`. That's using Redis, so we need to keep Redis as well.

Also, the download API has a `_refresh` querystring parameter that I said we should remove, but that's a bad idea because it forces the refresh for the cache used by `cached_lookup_by_syminfo` which we need to keep.

biancadanforth:
> Oh, wait--I'm confusing things. There are two caches involved here, but we need to keep the codeinfo -> syminfo one that's used in `cached_lookup_by_syminfo`. That's using Redis, so we need to keep Redis as well.
> 
> Also, the download API has a `_refresh` querystring parameter that I said we should remove, but that's a bad idea because it forces the refresh for the cache used by `cached_lookup_by_syminfo` which we need to keep.

Thanks; I was just about to ask, as that was my conclusion in my above [comment](https://github.com/mozilla-services/tecken/pull/2878#pullrequestreview-1828152573) as well (though I didn't do a good job of linking to the source code).

tecken/upload/utils.py
willkg: You removed the use of this `downloader`, so the `downloader` line (line 28 I think) can be removed as well as this comment.

willkg: For some context about the comment, it's marked as a FIXME because it hardcodes the symbol sources here. We really want the list of symbol sources to be configurable and not hardcoded in a bunch of places. We need the downloader here to cover all the possible sources (regular and try) so that it invalidates the cache correctly when someone uploads a file which might be in the cache as a 404. This fixed bug-1758566.

willkg (approved):
Looks good!

PR #2879, opened 2024-01-16
bug-1873752: add timing metric to remove_orphaned_files

relud:


tecken/upload/management/commands/remove_orphaned_files.py
```diff
@@ -85,6 +85,7 @@ def delete_orphans(self, watchdir, expires):
                         logger.exception("error deleting file: %s", fn)
                         METRICS.incr("remove_orphaned_files.delete_file_error")
 
+    @METRICS.timer_decorator("remove_orphaned_files")
```
willkg, line 88: This is something I get fuzzy on with statsd, telegraf, influxdb, and grafana. Does anything in that mix prevent us from having two metrics where the key of one is a substring of another? e.g. here we have:

* `tecken.remove_orphaned_files` (timing)
* `tecken.remove_orphaned_files.delete_file_error` (incr)

Also, we should add this to the tests. You could add an assertion to this test:

https://github.com/mozilla-services/tecken/blob/main/tecken/tests/test_upload.py#L1073-L1076

willkg (commented):
See the other comment that I submitted via github wrong, but is part of this review.

tecken/upload/management/commands/remove_orphaned_files.py
relud: I wasn't aware of that limitation, but it makes sense. I don't see any violations of it in tekcen, and in download views I see https://github.com/mozilla-services/tecken/blob/85a18162979b2682317b00e015de227c92c83ea2/tecken/download/views.py#L64 and https://github.com/mozilla-services/tecken/blob/85a18162979b2682317b00e015de227c92c83ea2/tecken/download/views.py#L91-L93

So I'll add `.timing` for consistency with the naming above

willkg: I also don't know if it's a limitation, but I've avoided the situation because if it was a problem, I'm not sure how it'd manifest and whether I'd notice.

When Obs Team levels up and multiclasses into statsd and metrics, we can ask them.

willkg (approved):
Looks good--just need to fix the reformat issue.

PR #2880, opened 2024-01-16
bug-1871348: use health checks in docker compose configuration

relud:


willkg (commented):
We had been using `urlwait` for something like this. If this works in local dev and CI, we can take out the `urlwait` requirement:

https://github.com/mozilla-services/tecken/blob/85a18162979b2682317b00e015de227c92c83ea2/requirements.in#L37

And also the place where it's used:

https://github.com/mozilla-services/tecken/blob/85a18162979b2682317b00e015de227c92c83ea2/bin/entrypoint.sh#L28-L33

docker-compose.yml
```diff
@@ -29,6 +29,15 @@ services:
       - localstack
       - oidcprovider
       - statsd
+    depends_on:
```
smarnach, line 32: I think we can remove the `link` section. We only have the default network, so everything can talk to everything by default, and th only purpose of the `link` section was to declare dependencies. Since we have `depends_on` now, the `link` section is redundant.

docker-compose.yml
```diff
@@ -107,18 +127,35 @@ services:
       - POSTGRES_USER=postgres
       - POSTGRES_PASSWORD=postgres
       - POSTGRES_DB=tecken
+    healthcheck:
+      test: ["CMD", "pg_isready"]
+      interval: 10s
```
smarnach, line 132: Maybe we want something shorter here? `pg_isready` returns immediately if the DB isn't up yet, so we'd wait for 10 seconds if it isn't up the first time the health check is run, which can be annoyingly long in interactive usage. If you are worried that a shorter interval isn't long enough, you can always increase the number of retries. Since these checks are all quite cheap, we can afford running them a few times.

(Same for some of the other intervals further down.)

smarnach (approved):
Thanks, this is looking great! I've added a couple of suggestions that are safe to ignore.

It's a bit unfortunate that `depends_on` doesn't default to `condition: service_healthy`. It's the whole point of health checks that dependencies should wait for them, so why is the default to ignore them?

docker-compose.yml
relud: good point, I changed the interval of all of these to 1s and timeout 3s and that seems to work fine for me locally

PR #2882, opened 2024-01-24
bug-1876375: Set devcontainer default shell to bash

relud:


biancadanforth (approved):
R+WC. Can you copy over some version of Sven's note into `.devcontainer/Dockerfile`?

Did you hit an issue where the shell was previously specified?

.devcontainer/Dockerfile
```diff
@@ -8,7 +8,7 @@ WORKDIR /app
 # Add a non-privileged user for installing and running the application.
 # We use --non-unique in case $groupid/$userid collide with the existing "vscode" user.
 RUN groupadd --gid $groupid --non-unique app && \
```
relud, line 10: > Can you copy over some version of Sven's note into `.devcontainer/Dockerfile`

Sven's comment was caused by setting `--shell /usr/sbin/nologin` here, so now that we are setting a sane default shell it's not relevant.

I could add a comment like this?

```suggestion
# We use --non-unique in case $groupid/$userid collide with the existing "vscode" user.
# We use --shell /bin/bash because the default is /bin/sh which doesn't support arrow keys
RUN groupadd --gid $groupid --non-unique app && \
```

biancadanforth: Ah okay thanks for clarifying. SGTM!

PR #2886, opened 2024-01-30
bug 1877534: mark backing services with --no-attach

willkg:
This reduces the copious output of backing services when running `make run`.

To test:

1. `docker compose stop` to reset everything
2. `./slick.sh` to start up backing services and set state for everything
3. `make run` to start up web, fakesentry, and frontend services

Expected: Output for web, fakesentry, and frontend services appears in stdout, but output for backing services (e.g. statsd and friends) does not.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,17 @@ setup: .env  ## | Initialize services.
 
 .PHONY: run
 run: .env .docker-build  ## | Run the web app and services.
-	${DC} up web frontend fakesentry
+	# NOTE(willkg): We tag all the backing services with --no-attach here to
+	# prevent them from spamming stdout
+	${DC} up \
+		--no-attach db \
+		--no-attach statsd \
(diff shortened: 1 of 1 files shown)


relud (approved):
i don't like that this has to list all the dependencies again, and i can't reproduce the original problem on mac or ubuntu, but i think it's still worth merging to fix the problem for willkg

relud:
also i don't see any problem with merging this as an immediate fix that may be removed if we find another way to deal with the problem in the future.

willkg:
^^^ That adds a comment about why the `--no-attach` things are there.

I'm puzzled that I'm having this problem, but no one else is.

This doesn't reduce the output from fakesentry. I wrote up https://github.com/willkg/kent/issues/69 to fix that.

willkg:
I messed up the bug number in the commit summary. Boo.

PR #2887, opened 2024-01-31
bug-1877534: upgrade Kent to 1.2.0

willkg:
This upgrades Kent to 1.2.0 to pick up fixes to logging and other things. That will quell the logging from the healthcheck hitting / every second.

This also updates the Dockerfile to use the same base image Tecken does.


relud (approved):

willkg:
Thank you!

PR #2892, opened 2024-02-13
obs-125: update django to 4.2.10

willkg:
4.2.8 release notes: https://docs.djangoproject.com/en/5.0/releases/4.2.8/

4.2.9 release notes: https://docs.djangoproject.com/en/5.0/releases/4.2.9/

4.2.10 release notes: https://docs.djangoproject.com/en/5.0/releases/4.2.10/

Generally, this just upgrades the version we're using, but shouldn't affect Tecken at all since we don't use most of the things that changed.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -37,4 +37,4 @@ werkzeug==3.0.1
 whitenoise==6.6.0
 
 # NOTE(willkg): We stick with LTS releases.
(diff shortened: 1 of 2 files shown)


biancadanforth (approved):

willkg:
Thank you!

PR #2900, opened 2024-03-01
bug-1881632: update circleci remote docker image

willkg:
Previously, we specified an image that's now deprecated. This removes the version specification which will cause it to use the default which is currently docker 24. default will change over time, but we'll know if CI fails pretty quick and we're switching to GCP soon with a completely different deploy pipeline anyhow.

We did this with Socorro:

https://github.com/mozilla-services/socorro/pull/6533

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,6 @@ jobs:
 
       - setup_remote_docker:
           docker_layer_caching: true
-          version: 20.10.18
 
       - run:
           name: Get info


relud (approved):

willkg:
@relud I looked at antenna and noticed you fixed this by changing the version to `docker24`. Should we be doing that instead? If so, I can change this PR and then go update Socorro.

relud:
i'm in favor of removing the version

willkg:
Thank you!

PR #2902, opened 2024-03-08
bug-1884456: switch to tenacity in systemtests

willkg:
This switches the systemtests code to use tenacity.

* remove `backoff` from requirements--we weren't using it
* add `tenacity` to requirements
* implement `http_get_with_retry` and `http_head_with_retry` which retry and have appropriate timeout default
* implement `http_post` which has an appropriate timeout default
* update the systemtests code to use `http_get_with_retry`, `http_head_with_retry`, and `http_post`
* fix `get_sym_files` to only emit files where completed_at is non-None so we don't get a 404 when trying to download them

While doing that, I fixed some minor issues with the systemtests:

* remove creation and consuming of `@ECHO` lines; the message we had was wrong because download csv can contain _very large_ symbols files that take longer than 1s to download
* remove outer try/except block in `setup_tests.py` so that errors percolate up and we see a traceback
* print file size when downloading a sym file and getting an HTTP 200.

Tenacity: https://tenacity.readthedocs.io/en/latest/

### DIFF
--- a/systemtests/bin/download_sym_files.py
+++ b/systemtests/bin/download_sym_files.py
@@ -13,7 +13,8 @@ from urllib.parse import urljoin
 import click
 import markus
 from markus.backends import BackendBase
-import requests
+
+from systemtestslib.utils import http_get_with_retry
 
 
 class StdoutMetrics(BackendBase):
@@ -72,7 +73,7 @@ def download_sym_files(base_url, csv_file):
 
             # Download the file
             headers = {"User-Agent": "tecken-systemtests"}
-            resp = requests.get(url, headers=headers, timeout=60)
+            resp = http_get_with_retry(url, headers=headers, timeout=60)
 
             for item in resp.history:
                 if item.status_code in (301, 302):
--- a/systemtests/bin/download_sym_files.py
+++ b/systemtests/bin/download_sym_files.py
@@ -59,11 +59,6 @@ def download_sym_files(base_url, csv_file):
             # Skip commented out and blank lines
             continue
 
-        if line.startswith("@ECHO"):
-            # This lets us encode directions for the viewer in the csv files
-            click.echo(click.style(line[5:].strip(), fg="yellow"))
-            continue
-
         with METRICS.timer("download_time"):
             parts = line.split(",")
 
@@ -84,6 +79,8 @@ def download_sym_files(base_url, csv_file):
                     click.echo(f">>> redirect: {item.url}")
             click.echo(f">>> final: {resp.url}")
             click.echo(f">>> status code: {resp.status_code}")
+            if resp.status_code == 200:
+                click.echo(f">>> file size: {len(resp.content):,} bytes")
 
             # Compare status code with expected status code
             if resp.status_code != int(expected_status_code):
--- a/systemtests/bin/setup_download_tests.py
+++ b/systemtests/bin/setup_download_tests.py
@@ -117,9 +117,7 @@ def write_list_of_sym_filenames_to_csv(sym_file_type_to_filename, outputdir):
     date = datetime.datetime.now().strftime("%Y-%m-%d")
     csv_rows.append([f"# File built: {date}"])
     csv_rows.append(["# Format: symfile", "expected status code", "bucket"])
-    csv_rows.append(["@ECHO NOTE: None of these should take over 1000ms to download."])
-    csv_rows.append(["@ECHO"])
-    csv_rows.append(["# These are recent uploads, so they're 200s."])
+    csv_rows.append(["# These are recent uploads, so they should return HTTP 200s."])
     for file_type, sym_filename in sym_file_type_to_filename.items():
         bucket = "regular"
         if file_type == "try":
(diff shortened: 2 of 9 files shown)


systemtests/bin/download_sym_files.py
```diff
@@ -77,13 +73,15 @@ def download_sym_files(base_url, csv_file):
...
-            resp = requests.get(url, headers=headers, timeout=60)
+            resp = http_get_with_retry(url, headers=headers, timeout=60)
 
             for item in resp.history:
                 if item.status_code in (301, 302):
                     click.echo(f">>> redirect: {item.url}")
             click.echo(f">>> final: {resp.url}")
             click.echo(f">>> status code: {resp.status_code}")
+            if resp.status_code == 200:
+                click.echo(f">>> file size: {len(resp.content):,} bytes")
```
willkg, line 84: This gives some helpful. context to why it took so long to download the file.

systemtests/bin/setup_download_tests.py
```diff
@@ -117,9 +117,7 @@ def write_list_of_sym_filenames_to_csv(sym_file_type_to_filename, outputdir):
     date = datetime.datetime.now().strftime("%Y-%m-%d")
     csv_rows.append([f"# File built: {date}"])
     csv_rows.append(["# Format: symfile", "expected status code", "bucket"])
-    csv_rows.append(["@ECHO NOTE: None of these should take over 1000ms to download."])
```
willkg, line 120: Because this file is generated automatically, it can contain `xul.sym` or other _very large_ symbols files which take longer than 1000ms (1s) to download. I couldn't think of a better note, so I removed this note and the `@ECHO` handling.

systemtests/setup_tests.py
```diff
@@ -27,44 +27,41 @@ def setup_tests(ctx):
         os.makedirs(ZIPS_DIR)
 
     click.echo("Generating systemtest data files ...")
-    try:
```
willkg, line 30: Removing this try/except outer layer will cause exceptions to bubble all the way up and then Python will print the exception type, message, _and traceback_ which makes it easier to figure out what went wrong.

systemtests/systemtestslib/utils.py
```diff
@@ -33,15 +124,19 @@ def get_sym_files(auth_token, url, start_page):
...
-                timeout=CONNECTION_TIMEOUT,
+                timeout=HTTP_DOWNLOAD_TIMEOUT,
             )
             resp.raise_for_status()
             data = resp.json()
-            sym_files = [(record["key"], record["size"]) for record in data["files"]]
+            sym_files = [
+                (record["key"], record["size"])
+                for record in data["files"]
+                if record["completed_at"]
```
willkg, line 138: The file list can come back with files that have records in the table, but _haven't been uploaded_, yet. In these cases, `completed_at` is `None`. We don't want to look at those because if we download them right after generating this list, they're likely to kick up an HTTP 404. We never noticed because the code that downloads these files silently ignored HTTP 404s. Now it calls `resp.raise_for_status()`.

biancadanforth (commented):
I'll let Sven's review stand, but I added some comments and questions in line. Nothing is blocking.

systemtests/bin/download_sym_files.py
willkg: f-strings can do a lot of amazing things, but the docs are terrible. Then I found this which is really helpful for me:

https://fstring.help/cheat/

systemtests/bin/setup_download_tests.py
willkg: I don't think we need a _very large_ symbol file for the test for bug #1655944. I think the chunked-transfer encoding will trigger for most of the symbols files we can build. The way jwhitlock reproduced the issue was with a zip file that was 120,684 bytes. That's not that big. The zip files I have in my `systemtests/data/zips/` dir are all > 26mb. We can adjust the code to look for a zip file that's > 120kb and use the first one and if there aren't any, have it raise an error.

Filtering for smaller files is a good idea. I can fix that in this PR.

systemtests/bin/upload_symbols.py
```diff
@@ -15,7 +15,8 @@
 import click
 import markus
 from markus.backends import BackendBase
-import requests
+
```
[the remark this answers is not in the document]
willkg: Since there's already a retry loop, I wanted to just stick with that rather than think through the specific situations that are retry-friendly for HTTP verbs that can be destructive. We should definitely tackle a retry-friendly POST. I'm not sure if we could write a generic one we could use most places or if we have to make each one personalized.

I can think through that for this PR, though. You're right in that it's weird I didn't do a retryable post.

systemtests/systemtestslib/utils.py
```diff
@@ -8,9 +8,100 @@
 import zipfile
 
 import requests
+import tenacity
```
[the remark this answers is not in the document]
biancadanforth: I don't feel very strongly about it, and FWICT VS Code does show the function definition on hover but nothing to indicate which library it's from until you jump to the function definition. Your point is a good one too.

willkg:
This PR is too unwieldy and it's going to be too hard to push it to the finish line. I'm going to close it out and create a new PR with just the fixes to the systemtests for problems I hit when working on this PR. That's more important now.

Then at some point, I'll take all the tenacity-related work and either pass it to someone else or finish that up.

PR #2905, opened 2024-03-26
Systemtest fixes

willkg:
`systemtestslib/utils.py`: Change `get_sym_files` to only emit files where `completed_at` is non-None so we don't get a 404 when trying to download them.

`systemtestslib/utils.py`: Shuffle around parameters in `get_sym_files` and callers.

`systemtestslib/utils.py`: Optimize `get_sym_files` for the case where someone uploaded a ton of try symbols recently--this will move through the pages faster reducing the number of GET requests.

`test_env.py`: Converts integer values to bools for flags, tweaks the output code, and improves a comment about the bad token test. This also adjusts the bad token test to make sure it uses a zip file that's greater than 120kb which we think we need to trigger the situation the bad token test tests.

`bin/setup_upload_tests.py`:
`bin/setup_download_tests.py`: When looking for appropriate symbols to use in the tests, restrict files by size to reduce the time it takes to download, zip, and upload the files which will reduce the time it takes to set up and run the tests.

`setup_tests.py`: Drops the zip file sizes for upload tests to 2mb and 4mb because that's plenty big; this reduces the number of files setup has to download and add to the zip file.

Add timings for setup scripts.

Fix some documentation, comments, curiosities, etc.

Remove backoff dependency--we're not using it.

### DIFF
--- a/systemtests/README.rst
+++ b/systemtests/README.rst
@@ -14,7 +14,10 @@ Contents of this directory::
 Setting up tests
 ================
 
-Set up the tests this way::
+Before running the systemtests, you need to `build a local dev environment
+<https://tecken.readthedocs.io/en/latest/dev.html>`__.
+
+Then set up the tests this way::
 
     $ make shell
     root@e62fb7ae586f:/app# cd systemtests
--- a/systemtests/README.rst
+++ b/systemtests/README.rst
@@ -22,15 +22,18 @@ Set up the tests this way::
 
 That creates files in directories under ``data/``.
 
-You only need to set up the tests once to run the system tests against all environments.
+You only need to set up the tests once to run the system tests against all
+environments.
 
 
 Running tests
 =============
 
-First, make sure you have valid, unexpired API tokens for the environment you're testing.
+First, make sure you have valid, unexpired API tokens for the environment
+you're testing.
 
-For destructive tests run in local and stage, you need separate auth tokens for try uploads with "Upload Try Symbols Files" permissions. See Bug 1759740.
+For destructive tests run in local and stage, you need separate auth tokens for
+try uploads with "Upload Try Symbols Files" permissions. See Bug 1759740.
 
 To set auth tokens, add these to your .env file:
 
@@ -56,8 +59,7 @@ where ``ENVIRONMENT`` is one of the following:
 Rules of systemtest
 ===================
 
-1. Thou shalt not import anything from ``tecken``. Test code must be
-   self-contained.
+1. Don't run destructive tests against the prod server environment.
 
-2. Destructive tests get added in ``test_env.sh`` in the destructive tests
+2. Destructive tests get added in ``test_env.py`` in the destructive tests
    section.
--- a/systemtests/bin/list_firefox_symbols_zips.py
+++ b/systemtests/bin/list_firefox_symbols_zips.py
@@ -67,13 +67,21 @@ def list_artifacts(taskid):
 
 
 def get_symbols_urls():
-    for taskid in tasks_by_changeset(10):
+    for taskid in tasks_by_changeset(revisions_limit=10):
         artifacts = list(list_artifacts(taskid))
-        full_zip = [a for a in artifacts if a.endswith(SYMBOLS_FULL_ZIP_SUFFIX)]
+        full_zip = [
+            artifact
+            for artifact in artifacts
+            if artifact.endswith(SYMBOLS_FULL_ZIP_SUFFIX)
+        ]
         if full_zip:
             yield QUEUE + f"task/{taskid}/artifacts/{full_zip[0]}"
         else:
-            nonfull_zip = [a for a in artifacts if a.endswith(SYMBOLS_ZIP_SUFFIX)]
+            nonfull_zip = [
+                artifact
+                for artifact in artifacts
+                if artifact.endswith(SYMBOLS_ZIP_SUFFIX)
+            ]
             if nonfull_zip:
                 yield QUEUE + f"task/{taskid}/artifacts/{nonfull_zip[0]}"
 
--- a/systemtests/bin/setup_download_tests.py
+++ b/systemtests/bin/setup_download_tests.py
@@ -17,6 +17,8 @@ import tempfile
 from urllib.parse import urljoin
 
 import click
+import markus
+from markus.backends import BackendBase
 
 from systemtestslib.utils import build_zip_file, download_sym_file, get_sym_files
 
@@ -36,13 +38,27 @@ REQUIRED_FILE_TYPES = ["try", "regular"]
 
 
 def iterate_through_symbols_files(
+    baseurl,
     auth_token,
-    sym_files_url,
     start_page,
+    try_symbols,
     sym_file_type_to_filename,
     end_condition_num_files,
 ):
-    sym_files_generator = get_sym_files(auth_token, sym_files_url, start_page)
+    params = {
+        "page": start_page,
+        "size": "< 10mb",
+        # NOTE(willkg): this restricts the search to symbols files only; when we decide
+        # to download other kinds of files, we'll need to remove this
+        "key": [".sym"],
+    }
+    if try_symbols:
+        params["key"].append("try/")
+
+    sym_files_generator = get_sym_files(
+        baseurl=baseurl, auth_token=auth_token, params=params
+    )
+
     for sym_filename, _ in sym_files_generator:
         if sym_filename.endswith(".0"):
             # Skip these because there aren't SYM files for them.
@@ -52,18 +68,17 @@ def iterate_through_symbols_files(
             # We only test for sym files currently
             continue
 
-        is_try = False
-
         if sym_filename.startswith("try/"):
+            if not try_symbols:
+                continue
             sym_filename = sym_filename[4:]
-            is_try = True
 
         if sym_filename.startswith("v1/"):
             sym_filename = sym_filename[3:]
 
         build_list_of_sym_filenames(
             sym_filename,
-            is_try,
+            try_symbols,
             sym_file_type_to_filename,
         )
 
@@ -117,9 +132,7 @@ def write_list_of_sym_filenames_to_csv(sym_file_type_to_filename, outputdir):
     date = datetime.datetime.now().strftime("%Y-%m-%d")
     csv_rows.append([f"# File built: {date}"])
     csv_rows.append(["# Format: symfile", "expected status code", "bucket"])
-    csv_rows.append(["@ECHO NOTE: None of these should take over 1000ms to download."])
-    csv_rows.append(["@ECHO"])
-    csv_rows.append(["# These are recent uploads, so they're 200s."])
+    csv_rows.append(["# These are recent uploads, so they should return HTTP 200s."])
     for file_type, sym_filename in sym_file_type_to_filename.items():
         bucket = "regular"
         if file_type == "try":
@@ -179,6 +192,15 @@ def get_subset_to_zip(is_try, sym_file_type_to_filename):
     return subset_to_zip
 
 
+class StdoutMetrics(BackendBase):
+    def emit(self, record):
+        click.echo(f"metric: {record.stat_type} {record.key} {record.value/1000:,.2f}s")
+
+
+markus.configure([{"class": StdoutMetrics}], raise_errors=True)
+METRICS = markus.get_metrics()
+
+
 @click.command()
 @click.option(
     "--auth-token",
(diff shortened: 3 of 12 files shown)


systemtests/bin/list_firefox_symbols_zips.py
```diff
@@ -67,13 +67,21 @@ def list_artifacts(taskid):
...
 def get_symbols_urls():
-    for taskid in tasks_by_changeset(10):
+    for taskid in tasks_by_changeset(revisions_limit=10):
         artifacts = list(list_artifacts(taskid))
-        full_zip = [a for a in artifacts if a.endswith(SYMBOLS_FULL_ZIP_SUFFIX)]
+        full_zip = [
+            artifact
+            for artifact in artifacts
+            if artifact.endswith(SYMBOLS_FULL_ZIP_SUFFIX)
+        ]
```
willkg, line 76: The only change here and below is changing the `a` variable to `artifact`.

systemtests/bin/setup_download_tests.py
```diff
@@ -36,13 +38,27 @@
...
-    sym_files_generator = get_sym_files(auth_token, sym_files_url, start_page)
+    params = {
+        "page": start_page,
+        "size": "< 10mb",
+        # NOTE(willkg): this restricts the search to symbols files only; when we decide
+        # to download other kinds of files, we'll need to remove this
+        "key": [".sym"],
+    }
+    if try_symbols:
+        params["key"].append("try/")
```
willkg, line 56: This filters on files that have `.sym` and files that have `try/`. That helps for when we're looking for try symbols since it reduces the results to a set closer to our needs.

However, there's no way to say "not try symbols", so it's not helpful for finding regular symbols. :cry:

systemtests/bin/setup_download_tests.py
```diff
@@ -52,18 +68,17 @@ def iterate_through_symbols_files(
             # We only test for sym files currently
             continue
 
-        is_try = False
-
         if sym_filename.startswith("try/"):
+            if not try_symbols:
+                continue
```
willkg, line 73: If this is a symbols file in the try bucket and we're not looking for those, this skips to the next filename to look at.

systemtests/bin/setup_download_tests.py
```diff
@@ -199,75 +221,79 @@ def setup_download_tests(start_page, auth_token, csv_output_path, zip_output_dir
...
-    sym_files_url = urljoin(SYMBOLS_URL, "/api/uploads/files?key=try/")
-    end_condition_num_files = 1
-    iterate_through_symbols_files(
-        auth_token,
-        sym_files_url,
-        start_page,
-        sym_file_type_to_filename,
-        end_condition_num_files,
-    )
+    with METRICS.timer("elapsed_time"):
```
willkg, line 233: This whole section is wrapped in a timer which spits out the elapsed time after the wrapped block is done. Thus the next bunch of lines are all indented.

systemtests/bin/setup_download_tests.py
```diff
@@ -199,75 +221,79 @@ def setup_download_tests(start_page, auth_token, csv_output_path, zip_output_dir
...
+        click.echo("Fetching try symbols files ...")
+        end_condition_num_files = 1
+        iterate_through_symbols_files(
+            baseurl=SYMBOLS_URL,
+            auth_token=auth_token,
+            start_page=start_page,
+            try_symbols=True,
+            sym_file_type_to_filename=sym_file_type_to_filename,
+            end_condition_num_files=end_condition_num_files,
+        )
```
willkg, line 244: I adjusted the arguments here and future calls to match the changes to `iterate_through_symbols_files`.

systemtests/bin/setup_upload_tests.py
```diff
@@ -70,57 +82,68 @@ def setup_upload_tests(max_size, start_page, auth_token, outputdir):
...
-                    if not os.path.exists(zip_path):
-                        shutil.copy(tmp_zip_path, zip_path)
-                    break
-
-                # This zip file isn't too big, so copy it over.
-                shutil.copy(tmp_zip_path, zip_path)
-
-            click.echo(
-                click.style(f"Adding {sym_filename} ({sym_size:,}) ...", fg="yellow")
+    with METRICS.timer("elapsed_time"):
```
willkg, line 85: I wrapped this block in a timer and indented it.

systemtests/bin/setup_upload_tests.py
```diff
@@ -70,57 +82,68 @@ def setup_upload_tests(max_size, start_page, auth_token, outputdir):
...
+        click.echo(f"Generating ZIP file {zip_path} ...")
+        with tempfile.TemporaryDirectory(prefix="symbols") as tmpdirname:
+            sym_dir = os.path.join(tmpdirname, "syms")
+            tmp_zip_path = os.path.join(tmpdirname, zip_filename)
+
+            params = {
+                "page": start_page,
+                "size": "< 10mb",
```
willkg, line 93: Restricting the size reduces the likelihood we pick up a xul symbols file--they're huge and slow everything down.

systemtests/bin/upload_symbols.py
```diff
@@ -85,8 +83,8 @@ def upload_symbols(ctx, expect_code, auth_token, base_url, symbolsfile):
                 fg="yellow",
             )
         )
-        try:
-            with METRICS.timer("upload_time"):
+        with METRICS.timer("upload_time"):
+            try:
```
willkg, line 87: I swapped these contexts so it prints an elapsed time even if it fails with an unhandled exception.

systemtests/bin/upload_symbols.py
```diff
@@ -103,33 +101,32 @@ def upload_symbols(ctx, expect_code, auth_token, base_url, symbolsfile):
                     )
                 )
 
+                # 403 means the auth token is bad which is not a retryable error
```
willkg, line 104: This section does the same thing, but the code is a little cleaner with fewer if/elif/else blocks to reason about.

systemtests/setup_tests.py
```diff
@@ -27,44 +27,42 @@ def setup_tests(ctx):
...
-                outputdir=f"{ZIPS_DIR}",
-            )
-        else:
-            click.echo(f"Already have {zips_count} zip files.")
-    except Exception as exc:
-        raise click.ClickException("Unexpected error") from exc
+        # Generate some symbols ZIP files to upload
+        ctx.invoke(
+            setup_upload_tests,
+            max_size=2_000_000,
```
willkg, line 52: I reduced the size from 10mb to 2mb. We don't need huge files for this and they take longer to put together and upload/download.

systemtests/setup_tests.py
```diff
@@ -27,44 +27,42 @@ def setup_tests(ctx):
...
+            start_page=1,
+            auth_token=PROD_AUTH_TOKEN,
+            outputdir=ZIPS_DIR,
+        )
+        ctx.invoke(
+            setup_upload_tests,
+            max_size=4_000_000,
```
willkg, line 59: I dropped this from 50mb to 4mb. Same reasoning--this file doesn't have to be huge, so we should spend the time on big files.

systemtests/setup_tests.py
```diff
@@ -27,44 +27,42 @@ def setup_tests(ctx):
         os.makedirs(ZIPS_DIR)
 
     click.echo("Generating systemtest data files ...")
-    try:
```
willkg, line 30: I removed the try block here. If this fails, we want to know about it and get a traceback so we can either run it again or improve the script.

systemtests/systemtestslib/utils.py
```diff
@@ -13,36 +14,51 @@
...
+    url = urljoin(baseurl.rstrip("/"), "/api/uploads/files")
     headers = {"auth-token": auth_token, "User-Agent": "tecken-systemtests"}
+    params = params or {}
+
+    sym_files = []
+    page = params.get("page", 1)
 
     while True:
-        if sym_files:
-            yield sym_files.pop(0)
```
willkg, line 33: I reworked this loop to use a `yield from ...` so it doesn't have multiple modes.

systemtests/systemtestslib/utils.py
```diff
@@ -13,36 +14,51 @@
...
+        data = resp.json()
+        sym_files = [
+            (record["key"], record["size"])
+            for record in data["files"]
+            if record["completed_at"]
+        ]
+        yield from sym_files
+
+        # NOTE(willkg): sometimes a whole bunch of try uploads happen at the same
+        # time and since there's no way to filter for just regular symbols files (as
```
biancadanforth, line 52: Are there reasons outside of the system tests that someone would want to filter for regular symbols? Is there a bug to be able to do this?

I've seen you mention this limititation a couple of times ([other mention](https://github.com/mozilla-services/tecken/pull/2905/files#r1540332442)) in this test as being inconvenient.

systemtests/test_env.py
```diff
@@ -89,56 +88,71 @@ def test_env(ctx, env):
...
-        URL = ctx.invoke(list_firefox_symbols_zips, number=1, max_size=1000000000)[0]
+        url = ctx.invoke(list_firefox_symbols_zips, number=1, max_size=1000000000)[0]
         ctx.invoke(
             upload_symbols_by_download,
-            base_url=f"{tecken_host}",
-            auth_token=f"{auth_token}",
-            url=f"{URL}",
+            base_url=tecken_host,
+            auth_token=auth_token,
+            url=url,
```
biancadanforth, line 105: Ah I guess we don't need an f-string if we're just passing the variable directly here, and it's not a substring for something else.

systemtests/test_env.py
```diff
@@ -89,56 +88,71 @@ def test_env(ctx, env):
...
+            )
+        ]
 
-    click.echo("\n")
+        if not zip_files:
+            click.echo(
+                click.style(
+                    ">>> No zip files > 120kb; skipping bad token test", fg="red"
+                )
+            )
```
biancadanforth, line 139: Do we want to make sure there's a zip file that's > 120kb as part of setup_tests.py, so that we don't risk not running this test?

systemtests/README.rst
```diff
@@ -22,15 +22,18 @@ Set up the tests this way::
 
 That creates files in directories under ``data/``.
 
-You only need to set up the tests once to run the system tests against all environments.
+You only need to set up the tests once to run the system tests against all
+environments.
```
biancadanforth, line 26: While we're here, I realized it's not mentioned that the local Tecken stack needs to be running for the local tests to work (and that means `./slick.sh` as well as `make run`). I only did the former, and hit an error since the `web` container wasn't running. Is it worth mentioning this in the README?

biancadanforth (approved):
Works great and much faster. I should have included some of the size restrictions when I wrote `setup_download_tests.py` -- I thought about it, but the PR was so big already. I know that's only one of many optimizations you made here. The result: it runs way faster.

Dropped a bunch of non-blocking questions inline. The only two more moderate questions are around:
* the `bad_token_test`, and if we want to guarantee there's a file large enough, so the test doesn't risk getting skipped.
* if/how to document the fact that we're only looking at .sym files for these tests and not any other file types, and documenting the meaning of the other file types we support.

systemtests/README.rst
willkg: I'll add a note that you have to follow the local dev environment set up process.

systemtests/bin/setup_download_tests.py
```diff
@@ -36,13 +38,27 @@
...
     sym_file_type_to_filename,
     end_condition_num_files,
 ):
-    sym_files_generator = get_sym_files(auth_token, sym_files_url, start_page)
+    params = {
+        "page": start_page,
+        "size": "< 10mb",
+        # NOTE(willkg): this restricts the search to symbols files only; when we decide
+        # to download other kinds of files, we'll need to remove this
+        "key": [".sym"],
```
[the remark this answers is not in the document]
willkg: This is covered in https://bugzilla.mozilla.org/show_bug.cgi?id=1884367 . We can add to that.

systemtests/systemtestslib/utils.py
willkg: There is no bug I'm aware of. The "API" we're using here isn't an official API. It's something Peter wrote to support the React-based web frontend so the ergonomics of it are determined by the webapp ui. I haven't made any changes since I don't want to extend the API beyond what we're using in the webapp and I don't want to make changes to the webapp because we have no tests and it's antiquated. I keep hoping we redo the ui which would also cause us to redo this API.

willkg:
Thank you!

PR #2917, opened 2024-04-20
 bug-1889200: specify release in sentry-wrap.py

biancadanforth:
Because:
* We want to be able to mark an error hit in `sentry_wrap.py` as "Resolved in the next release" in Sentry, but we weren't passing in the `release` in this context (see this [reference error's events](https://mozilla.sentry.io/issues/4723701783/events/?project=6700123)).

Screenshot 2024-04-23 at 4 57 46 PM

This commit:
* Passes `release` when calling `sentry_sdk.init` in `sentry_wrap.py`.

### DIFF
--- a/bin/sentry_wrap.py
+++ b/bin/sentry_wrap.py
@@ -19,7 +19,6 @@ import os
 from pathlib import Path
 import shlex
 import subprocess
-import sys
 import time
 import traceback
 
@@ -63,13 +62,7 @@ def get_release_name(basedir):
     return f"{version}:{commit}"
 
 
-def set_up_sentry():
-    sentry_dsn = os.environ.get("SENTRY_DSN")
-
-    if not sentry_dsn:
-        click.echo("SENTRY_DSN is not defined. Exiting.", err=True)
-        sys.exit(1)
-
+def set_up_sentry(sentry_dsn):
     current_dir = os.path.dirname(os.path.abspath(__file__))
     base_dir = os.path.dirname(current_dir)
     release = get_release_name(base_dir)
@@ -84,7 +77,13 @@ def cli_main():
 @cli_main.command()
 @click.pass_context
 def test_sentry(ctx):
-    set_up_sentry()
+    sentry_dsn = os.environ.get("SENTRY_DSN")
+
+    if not sentry_dsn:
+        click.echo("SENTRY_DSN is not defined. Exiting.", err=True)
+        ctx.exit(1)
+
+    set_up_sentry(sentry_dsn)
 
     capture_message("Sentry test")
     click.echo("Success. Check Sentry.")
@@ -107,9 +106,15 @@ def wrap_process(ctx, timeout, verbose, cmd):
     if not cmd:
         raise click.UsageError("CMD required")
 
-    set_up_sentry()
+    sentry_dsn = os.environ.get("SENTRY_DSN")
+
+    if not sentry_dsn:
+        click.echo("SENTRY_DSN is not defined. Exiting.", err=True)
+        ctx.exit(1)
 
     start_time = time.time()
+
+    set_up_sentry(sentry_dsn)
 
     cmd = " ".join(cmd)
     cmd_args = shlex.split(cmd)
--- a/bin/sentry_wrap.py
+++ b/bin/sentry_wrap.py
@@ -7,10 +7,10 @@
 # Wraps a command such that if it fails, an error report is sent to the Sentry service
 # specified by SENTRY_DSN in the environment.
 #
-# Usage: python bin/sentry-wrap.py wrap-process -- [CMD]
+# Usage: python bin/sentry_wrap.py wrap-process -- [CMD]
 #    Wraps a process in error-reporting Sentry goodness.
 #
-# Usage: python bin/sentry-wrap.py test-sentry
+# Usage: python bin/sentry_wrap.py test-sentry
 #    Tests Sentry configuration and connection.
 
 
--- a/bin/sentry_wrap.py
+++ b/bin/sentry_wrap.py
@@ -14,7 +14,9 @@
 #    Tests Sentry configuration and connection.
 
 
+import json
 import os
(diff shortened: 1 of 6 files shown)


tecken/tests/test_sentry.py
```diff
@@ -226,3 +232,33 @@ def test_count_sentry_scrub_error(metricsmock):
...
+    port = os.environ.get("EXPOSE_SENTRY_PORT", 8090)
+
+    # Flush fakesentry to ensure we fetch only the desired error downstream
+    requests.post(f"http://fakesentry:{port}/api/flush/")
+
+    current_dir = os.path.dirname(os.path.abspath(__file__))
+    base_dir = os.path.join(os.path.dirname(current_dir), "..")
+    release = get_release_name(base_dir)
+
+    expected_event = {"release": release}
```
biancadanforth, line 247: I couldn't figure out how to mock `get_release_name` when `sentry_wrap.py` is executed in a subprocess. Previously, when I was running this test in the test process, the `unittest` `@patch` decorator did this.

tecken/tests/test_sentry.py
```diff
@@ -226,3 +232,33 @@ def test_count_sentry_scrub_error(metricsmock):
...
+
+    # Pass a non-Django command that will error to sentry_wrap
+    non_app_command = "ls -2"
+    sentry_wrap_command = f"python bin/sentry_wrap.py wrap-process -- {non_app_command}"
+    cmd_args = shlex.split(sentry_wrap_command)
+    subprocess.run(cmd_args, timeout=10)
+
+    # Avoid a race condition where fakesentry hasn't processed the event yet.
+    time.sleep(1)
```
biancadanforth: I ran the test several times locally at various sleep times (including sub-1s). This has a safety factor of at least 3x, so hopefully we don't see any intermittent failures in CI downstream.

tecken/tests/test_sentry.py
```diff
@@ -226,3 +232,38 @@ def test_count_sentry_scrub_error(metricsmock):
...
+    while not is_condition_met:
+        time.sleep(1)
+        errors_resp = requests.get(f"http://fakesentry:{port}/api/errorlist/")
+        errors_resp.raise_for_status()
+        if len(errors_resp.json()) > 0:
+            error_id = errors_resp.json()["errors"][0]
+            error_resp = requests.get(f"http://fakesentry:{port}/api/error/{error_id}")
+            error_resp.raise_for_status()
+            actual_event = error_resp.json()["payload"]
+            is_condition_met = True
```
biancadanforth, line 267: I know this is generally an antipattern, but I couldn't find a better way to do this. Based on local testing, this `while` loop will typically only run through once (likely since we're making a request to another local docker container).

willkg: One nice thing about running `sentry_wrap.py` in a subprocess is that when that process is done, we're guaranteed that sentry has sent the event and it's been handled by Kent and stored. There shouldn't be a race condition anymore.

After reworking the test to run `sentry_wrap.py` in a subprocess, did you continue to have problems with Kent not having any errors, yet?

willkg (changes requested):

tecken/tests/test_sentry.py
biancadanforth: Thanks! Yeah I thought about trying to mess with `version.json`, but it seemed more complicated, given that whether or not it exists and what its contents are vary by environment. I'll keep it this way in that case.

tecken/tests/test_sentry.py
biancadanforth: Ah good point, because part of the code in `sentry_wrap.py` handles `kent`'s response. Yep I just ran in 20x locally and don't see the race condition with 0 sleeps. Nice!

bin/sentry_wrap.py
```diff
@@ -34,13 +84,8 @@ def cli_main():
 @cli_main.command()
 @click.pass_context
 def test_sentry(ctx):
-    sentry_dsn = os.environ.get("SENTRY_DSN")
+    set_up_sentry()
```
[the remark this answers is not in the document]
biancadanforth: Sure I can change that -- is `ctx.exit` better than `sys.exit` for some reason? Outwardly (from the command line), I see no difference. The [`click` docs](https://click.palletsprojects.com/en/8.1.x/api/#click.Context.exit) are pretty sparse on `exit` as well.

tecken/tests/test_sentry.py
```diff
@@ -226,3 +231,34 @@ def test_count_sentry_scrub_error(metricsmock):
...
+
+def test_sentry_wrap_non_app_error_has_release():
+    port = os.environ.get("EXPOSE_SENTRY_PORT", 8090)
+
+    # Flush fakesentry to ensure we fetch only the desired error downstream
+    requests.post(f"http://fakesentry:{port}/api/flush/")
+
+    current_dir = os.path.dirname(os.path.abspath(__file__))
+    base_dir = os.path.join(os.path.dirname(current_dir), "..")
+    release = get_release_name(base_dir)
```
willkg, line 244: We might want to assert that the release is in an expected shape. Namely something like:

* it's "unknown", OR
* value has a ":" in it

But this is good enough for now.

willkg (approved):
Looks good!

tecken/tests/test_sentry.py
biancadanforth: There are also tests for `get_release_name` already in `test_libdockerflow.py` which test that sort of thing.

bin/sentry_wrap.py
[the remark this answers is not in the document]
willkg: I'm pretty sure `ctx.exit` works better when we chain commands using click. I don't think we'd do that here, but one never knows and generally I worry about copy-pasting wrong things around. It should have been a `ctx.exit` originally--not sure why I used `sys.exit`. Maybe sentry-wrap evolved from a non-click script to a click script or something like that.

PR #2923, opened 2024-05-06
bug-1873752: double remove_orphaned_files timeout again

relud:
the timeout occurred again, as discussed doubling the timeout again

### DIFF
--- a/bin/run_web_disk_manager.sh
+++ b/bin/run_web_disk_manager.sh
@@ -17,7 +17,7 @@ cd /app/
 export PROCESS_NAME=disk_manager
 
 SLEEP_SECONDS=60
-PROCESS_TIMEOUT_SECONDS=120
+PROCESS_TIMEOUT_SECONDS=240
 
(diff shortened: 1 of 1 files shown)


bin/run_web_disk_manager.sh
```diff
@@ -17,7 +17,7 @@ cd /app/
 export PROCESS_NAME=disk_manager
 
 SLEEP_SECONDS=60
-PROCESS_TIMEOUT_SECONDS=120
+PROCESS_TIMEOUT_SECONDS=240
```
willkg, line 20: This feels weird to me. If it takes more than 2 minutes to run, I wonder if the problem is less that it doesn't have time to finish and more that it's hanging somewhere.

willkg: I fixed the `remove_orphaned_files` timing graph and then looked at the last 7 days and it does get up to 2 minutes. I double-checked the code and there's no obvious places it would hang. If it's hanging, it's in the python vm or function implementations.

I'm game for increasing this to 240.

willkg (approved):

willkg:
image

PR #2928, opened 2024-05-07
Remove reviewers from Dependabot configuration.

smarnach:
The Dependabot PRs get assigned to mozilla-services/obs-team automatically due to our branch protection rules, so we don't need a reviewer in the Dependabot configuration. Moreover, the current configuration tried to assign the PRs to https://github.com/obs-team, which is different from https://github.com/orgs/mozilla-services/teams/obs-team.


willkg (approved):
Thank you for doing this!

PR #2932, opened 2024-05-15
bug-1896962: improve Token admin

willkg:
This impproves the Token admin making it effectively like the one in Crash Stats. One nice thing is that tokens are searchable now.

To test:

1. create a token:
   ```
   docker compose run --rm web bash python manage.py createtoken EMAIL TOKEN
   ```
2. run everything with `make run`
3. log in, go to the admin, go to the token manager
4. view token list
5. search tokens using email address substring

### DIFF
--- a/tecken/tokens/admin.py
+++ b/tecken/tokens/admin.py
@@ -24,6 +24,7 @@ class TokenAdmin(admin.ModelAdmin):
     def key_truncated(self, obj):
         return obj.key[:12] + "..."
 
+    @admin.display(description="Permissions")
     def get_permissions(self, obj):
         return ", ".join(perm.codename for perm in obj.permissions.all())
 
--- a/tecken/tokens/admin.py
+++ b/tecken/tokens/admin.py
@@ -28,5 +28,6 @@ class TokenAdmin(admin.ModelAdmin):
     def get_permissions(self, obj):
         return ", ".join(perm.codename for perm in obj.permissions.all())
 
+    @admin.display(description="Email")
     def get_user_email(self, obj):
         return obj.user.email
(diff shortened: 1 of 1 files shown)


tecken/tokens/admin.py
```diff
@@ -9,12 +9,22 @@
...
         "expires_at",
-        "created_at",
+        "notes",
     ]
 
-    def user_email(self, obj):
+    list_filter = ["permissions"]
+    search_fields = ["user__email", "notes"]
+
+    def key_truncated(self, obj):
```
smarnach, line 23:
```suggestion
    @admin.display(description="Key")
    def key_truncated(self, obj):
```

tecken/tokens/admin.py
```diff
@@ -9,12 +9,22 @@
...
+        return obj.key[:12] + "..."
+
+    def get_permissions(self, obj):
```
smarnach, line 26:
```suggestion
    @admin.display(description="Permissions")
    def get_permissions(self, obj):
```

tecken/tokens/admin.py
```diff
@@ -9,12 +9,22 @@
...
+        return ", ".join(perm.codename for perm in obj.permissions.all())
+
+    def get_user_email(self, obj):
```
smarnach, line 29:
```suggestion
    @admin.display(description="Email")
    def get_user_email(self, obj):
```

smarnach (approved):
I tested this and it works fine. Only minor cosmetic issue is that the column titles are automatically derived from the method names, so the column with the email address is called "Get User Email", which is a bit weird. I added a suggestion with different names, which I tested and confirmed to be working.

willkg:
Thank you!

smarnach:
I didn't realize every suggestion becomes a separate commit. I should probably have made that a single suggestion instead.

willkg:
You did fine! No worries!

PR #2945, opened 2024-06-24
bug-1904042: Remove all uses of the Upload.bucket_region field.

smarnach:
This removes all code using the bucket_region field while keeping the field in the model. New records will be created with the field set to NULL, and the field will be removed from the model and the database in a subsequent change.


willkg (approved):
Looks good to me.

I built Tecken from the main branch, did some uploads, switched to this branch, verified that the upload pages work, then did some more uploads. Everything looks fine.

I'm not sure why Tecken was tracking bucket_region in the first place. It's not something that was related to the private symbols bucket or anything I remember adjusting/removing over the last 4 years.

PR #2954, opened 2024-07-03
bug-1827507: Fix GCS emulator settings.

smarnach:
The GCS emulator needs special settings to allow public downloads and retriable uploads.

Public downloads are required so we can redirect to files in the GCS emulator from the download API. They are enabled by the `-public-host` command-Line option.

Retriable uploads are used by the `Blob.upload_from_file()` method. They return a redirect in the initial request, which requires the server to know its external URL, which can be set using `-external-url`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -156,7 +156,11 @@ services:
     build:
       context: docker/images/gcs-emulator
     image: local/tecken_gcs_emulator
-    command: -port 8001 -scheme http
+    command: >-
+      -port 8001
+      -scheme http
+      -external-url http://gcs-emulator:"${EXPOSE_GCS_EMULATOR_PORT:-8001}"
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -156,7 +156,11 @@ services:
     build:
       context: docker/images/gcs-emulator
     image: local/tecken_gcs_emulator
-    command: -port 8001 -scheme http
+    command: >-
```
willkg, line 159: Things I learned: the `>-` replaces newlines with spaces and chomps the last newline.

willkg (approved):
Looks good!

docker-compose.yml
smarnach: I find it impossible to remember the various options for YAML multiline strings. The only thing I remember is that there is https://yaml-multiline.info/, and everytime I need something unsual I look it up there.

PR #2962, opened 2024-07-16
bug-1908096: Remove unused Python requirements and Django settings.

smarnach:


relud (approved):
perhaps we _should_ be using ujson and falling back to stdlib? it's my understanding that ujson is the go-to implementation for json parsing speed.

regardless lgtm

willkg:
I skimmed the codebase and Tecken doesn't really use JSON except for some trivial things like the manifest and version endpoints. It used to back when it did symbolication, but we spun that out into a separate service. It'd probably be fine to use json here.

But we might want to look at Eliot again. I can't remember if I stuck with ujson or json there.

PR #2965, opened 2024-07-16
bug-1905455: Enable anonymous access to storage buckets.

smarnach:


biancadanforth:
I had more of a general question -- from testing the STR for the initial bug reported locally, the error looks to be coming from Django's DB `migrate` command as a system check error. I'm having a hard time telling where `SYMBOL_URLS` comes into play. Could either of you point me in the right direction from here?

willkg (approved):
Looks good! I can access symbols files via redirect in my local dev env now.

willkg:
@biancadanforth I bumped into the original issue when testing the setup we use in stage where `SYMBOL_URLS` has an additional HTTP-access-only bucket url which is the prod bucket. Since we don't upload to stage very often, it allows us to test download API with stage and stage will check the stage buckets, then the prod bucket. In the original issue, that kicks up a Django check error because the prod bucket url is an HTTP-access-only url that stage is accessing via HTTP and not S3 protocols. When the `.exists()` check runs, it only works with S3 protocols. Therefore it failed.

When we redid how configuration works in the last couple of weeks, we dropped this feature. Stage will now be configured just like the local dev environment and prod and won't additionally check the prod bucket via HTTP-access-only methods.

PR #2966, opened 2024-07-18
bug-1908689: Allow running systemtests against the new GCP environments.

smarnach:
I refactored the code a bit as well.

* The information about the different environments is now data rather than code.
* I simplified iterating of the list of zip files a bit.

### DIFF
--- a/systemtests/README.rst
+++ b/systemtests/README.rst
@@ -32,31 +32,26 @@ environments.
 Running tests
 =============
 
-First, make sure you have valid, unexpired API tokens for the environment
-you're testing.
+The system tests are run using the ``test_env.py`` Python script. You can get
+help about the command-line invocation of that script using::
+
+    $ make shell
+    root@e62fb7ae586f:/app# cd systemtests
+    root@e62fb7ae586f:/app/systemtests# ./test_env.py --help
+
+The help includes a list of available environments.
+
+You need to make sure you have valid, unexpired API tokens for the environment
+you're testing. Add these tokens to your ``.env`` file using the environment
+variable names in the help output, e.g. ``STAGE_AUTH_TOKEN`` and
+``STAGE_AUTH_TOKEN_TRY`` for the stage environment.
 
 For destructive tests run in local and stage, you need separate auth tokens for
 try uploads with "Upload Try Symbols Files" permissions. See Bug 1759740.
 
-To set auth tokens, add these to your .env file:
-
-* `LOCAL_AUTH_TOKEN`
-* `LOCAL_AUTH_TOKEN_TRY`
-* `STAGE_AUTH_TOKEN`
-* `STAGE_AUTH_TOKEN_TRY`
-* `PROD_AUTH_TOKEN`
-
 To run the systemtests, do::
 
-   $ make shell
-   root@f09b3cdf8570:/app# cd systemtests/
-   root@e62fb7ae586f:/app/systemtests# ./test_env.py ENVIRONMENT
-
-where ``ENVIRONMENT`` is one of the following:
-
-* ``local``: run all tests against your local dev environment
-* ``stage``: run all tests against stage
-* ``prod``: run non-destructive tests against prod
+    root@e62fb7ae586f:/app/systemtests# ./test_env.py <ENV_NAME>
 
 
 Rules of systemtest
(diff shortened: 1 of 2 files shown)


systemtests/test_env.py
```diff
@@ -40,75 +89,51 @@ def test_env(ctx, env):
...
         click.echo(
             click.style(">>> UPLOAD BY DOWNLOAD TEST (DESTRUCTIVE)", fg="yellow")
         )
-        url = ctx.invoke(list_firefox_symbols_zips, number=1, max_size=1000000000)[0]
+        url = ctx.invoke(list_firefox_symbols_zips, number=1, max_size=1_000_000_000)[0]
         ctx.invoke(
             upload_symbols_by_download,
-            base_url=tecken_host,
+            base_url=env.base_url,
             auth_token=auth_token,
```
relud, line 129: `auth_token` has the potential to be undefined here because it's set inside a for loop above

biancadanforth: The presumption is that you've run `./setup_tests.py` prior to running this script, so that there are some `ZIP_FILES`, but we don't check that it's empty here. We could add a check to do so and remind people to setup the tests first?

biancadanforth (approved):
R+WC. A really nice way to clean up the test config that varies across environments.

Mostly just hoping you can update the `--help` for the `test_env` command and the README with the CLI argument changes and env var changes?

WFM against GCP stage, though I got a lot of 429s, and the negative download tests that are supposed to 404 in AWS return a 500 in GCP. I know actually getting the systemtests to pass for GCP are out of scope here, just an observation.

smarnach:
> Mostly just hoping you can update the `--help` for the `test_env` command and the README with the CLI argument changes and env var changes?

I did that, and took the opportuinity to reword the help string a bit.

> WFM against GCP stage, though I got a lot of 429s, and the negative download tests that are supposed to 404 in AWS return a 500 in GCP. I know actually getting the systemtests to pass for GCP are out of scope here, just an observation.

In GCP stage, we only have a single instance. The rate-limiting set-up is exactly the same as in AWS, though, so you should get approximately the same number of 429s for AWS stage.

The 500 was a copy-and-paste error that [should be fixed now](https://github.com/mozilla-services/tecken/pull/2957/commits/01311fa3b2b0a9ce32c765b0723934a4af65e4d7).

smarnach:
@biancadanforth Could you take another look, please?

systemtests/README.rst
```diff
@@ -32,31 +32,26 @@ environments.
 Running tests
 =============
 
-First, make sure you have valid, unexpired API tokens for the environment
-you're testing.
+The system tests are run using the ``test_ven.py`` Python script. You can get
```
biancadanforth, line 35: Nit: typo on the script name.

systemtests/README.rst
```diff
@@ -32,31 +32,26 @@ environments.
...
+For destructive tests run in local and stage, you need separate auth tokens for
+try uploads with "Upload Try Symbols Files" permissions. See Bug 1759740.
 
-where ``ENVIRONMENT`` is one of the following:
+To run the systemtests, do::
 
-* ``local``: run all tests against your local dev environment
-* ``stage``: run all tests against stage
-* ``prod``: run non-destructive tests against prod
+    root@e62fb7ae586f:/app/systemtests# ./test_env.py <ENV_NAME>
```
biancadanforth, line 54: This assumes you've done a bunch of setup work now listed under `USAGE` in the `test_env.py` `--help`. Why not keep those in sync, else reference `USAGE` in the `--help` here?

systemtests/test_env.py
```diff
@@ -29,74 +33,87 @@ class Environment:
...
+CLI_HELP = f"""
+Runs upload and download tests against the environment specified in ENV_NAME.
+
+Available environments:
+
+{ENV_HELP}
+
+\b
+Usage:
+1. run "make shell" to get a shell in the container
```
biancadanforth, line 103: Note: To run the systemtests against `local`, `make shell` is insufficient. They also need to ensure their local stack is running (i.e. `make run`). Probably worth mentioning?

biancadanforth (approved):
LGTM! Made some suggestions on the docs, but in general this is a really nice improvement all around.

systemtests/README.rst
smarnach: The setup is documented in the first section of the README file. The first section is "Setting up the tests". The second section is "Running the tests".

systemtests/test_env.py
smarnach: Hmm, I'd think that's clear enough – if you want to test the local environment, you need to run it. How to do that is documented elsewhere, so I'd prefer to restrict this documentation to aspects specific to the system tests. If you forget `make run`, the error message will make that sufficiently clear.

PR #2971, opened 2024-07-25
bug-1906139: systemtests download API improvements

biancadanforth:
Because:
* We will rely on systemtests as a primary way to validate the Tecken GCP migration.
* We were not testing HEAD requests or verifying response headers in the systemtests for the download API.

This commit:
* Tests the HTTP response for HEAD requests.
* Checks response headers for both the Tecken redirect response and the storage backend response.
  * What headers to check for and their values is based largely on the specifications in the [Tecken GCP migration plan](https://docs.google.com/document/d/1tLoGhMJQrP1m6A5H3D5zElNbTHQ9_vzYbOE9QnrxLUc/edit?usp=sharing).

### DIFF
--- a/systemtests/bin/download_sym_files.py
+++ b/systemtests/bin/download_sym_files.py
@@ -20,7 +20,7 @@ from unittest.mock import ANY
 TECKEN_RESPONSE_HEADERS = {
     "X-Content-Type-Options": "nosniff",
     "X-Frame-Options": "DENY",
-    "Content-Security-Policy": "font-src 'self'; object-src 'none'; script-src 'self'; img-src 'self'; default-src 'self'; style-src 'self'; connect-src 'self'; frame-ancestors 'none'",
+    "Content-Security-Policy": ANY,
     "Strict-Transport-Security": "max-age=31536000",
 }
 
@@ -182,10 +182,10 @@ def download_sym_files(base_url, test_headers, csv_file):
                 else:
                     click.echo(click.style(f"SUCCESS: {method} request!", fg="green"))
 
-                # If we're testing a 404, we should not expect the same response
-                # headers from the storage backend as for a 200 (e.g. Content-Encoding
-                # isn't included in a 404 response).
-                if method == "GET" and expected_status_code != "404" and test_headers:
+                # We're interested in the success case for a request when testing headers.
+                # Non-200 responses won't necessarily have the same response headers;
+                # e.g. Content-Encoding isn't included in a 404 response.
+                if method == "GET" and expected_status_code == "200" and test_headers:
                     check_headers(resp)
 
             click.echo(f">>> {method} final: {resp.url}")
--- a/systemtests/bin/download_sym_files.py
+++ b/systemtests/bin/download_sym_files.py
@@ -14,6 +14,20 @@ import click
 import markus
 from markus.backends import BackendBase
 import requests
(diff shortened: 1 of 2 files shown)


systemtests/bin/download_sym_files.py
```diff
@@ -25,18 +39,92 @@ def emit(self, record):
...
+                            fg="green",
+                        )
+                    )
+                else:
+                    click.echo(
+                        click.style(
+                            f"""
+>>> FAIL: GET expected response header {expected_key} does not have expected value:
+actual: {item.headers[expected_key]}
+expected: {expected_value}
```
biancadanforth, line 71: This is a bit ugly. The motivation is to get `actual` and `expected` on new lines. There's probably a better way to do this in Python.

I know we're going to be moving some of these tests to pytest soon, so maybe it doesn't matter that much in the short term.

systemtests/bin/download_sym_files.py
```diff
@@ -14,6 +14,20 @@
 import markus
 from markus.backends import BackendBase
 import requests
+from unittest.mock import ANY
+
+# These are exclusively security headers added by nginx
+TECKEN_RESPONSE_HEADERS = {
+    "X-Content-Type-Options": "nosniff",
+    "X-Frame-Options": "DENY",
+    "Content-Security-Policy": "font-src 'self'; object-src 'none'; script-src 'self'; img-src 'self'; default-src 'self'; style-src 'self'; connect-src 'self'; frame-ancestors 'none'",
```
biancadanforth, line 23: While the new HEAD request checks pass, there is one response header check that is failing for AWS stage and prod: `Content-Security-Policy` expected versus actual do not match for the Tecken redirect response. The difference is that the actual header response has: `; report-uri /__cspreport__`.

MDN [indicates](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri) that this `report-uri` directive is deprecated and no longer recommended in favor of `report-to`, but `report-to` isn't yet supported in all browsers. Should we include both the `report-uri` and `report-to` directives? Do we still want these directives?

Also who looks at these reports?

biancadanforth (commented):
Added some specific questions inline for the review. Thanks in advance!

systemtests/bin/download_sym_files.py
smarnach: Yeah, it's fine. We'll let pytest take care of the details here soon.

systemtests/bin/download_sym_files.py
```diff
@@ -68,28 +156,46 @@ def download_sym_files(base_url, csv_file):
...
-                )
-            else:
-                click.echo(click.style("Success!", fg="green"))
+                else:
+                    click.echo(click.style(f"SUCCESS: {method} request!", fg="green"))
+
+                # If we're testing a 404, we should not expect the same response
+                # headers from the storage backend as for a 200 (e.g. Content-Encoding
+                # isn't included in a 404 response).
+                if method == "GET" and expected_status_code != "404" and test_headers:
```
smarnach, line 188: I'd go a step further and use `excpected_status_code == "200"`. Testing the storage backend headers for any other status code doesn't make sense.

smarnach (approved):
This is looking good to me. I've just had one tiny comment.

PR #2972, opened 2024-07-29
bug-1909870: Rename old system tests to smoke tests.

smarnach:
We would like to introduce a new set of system tests, but we'd like to keep the old system tests around, since they test some interesting aspects or the public API. This PR renames the old tests from "sytem tests" to "smoke tests".


docs/dev.rst
```diff
@@ -634,19 +634,19 @@ this code and it has to be tested manually. You can do something like this:
 11. click on "Sign out"
 
 
-System tests
+Smoke tests
 ------------
```
willkg, line 638:
```suggestion
-----------
```

docs/overview.rst
```diff
@@ -52,12 +52,11 @@ Here's a bunch of top-level directories and what's in them::
     docs/                   -- documentation
     favicons/               -- favicons used by Symbols webapp
     frontend/               -- Symbols webapp JavaScript frontend
-    schemas/                -- API schemas
```
willkg, line 55: For my notes: The schemas held the symbolication v4 and v5 reponse jsonschema specs. When we built Eliot and then moved that to a new repo, we didn't have anymore schemas.

smoketests/README.rst
```diff
@@ -1,5 +1,5 @@
 =================
-Systemtest README
+Smoke tests README
 =================
```
willkg, line 3: It doesn't let me make a suggestion here for whatever reason, but the top and bottom lines need an additional `=`.

willkg (approved):
Looks good!

smoketests/README.rst
smarnach: Looks like a GitHub bug. The tool tip says "Applying suggestions on deleted lines is not supported". My guess is it's because I moved and modified the file at the same time.

PR #2973, opened 2024-07-29
bug-1827506: Add setting to configure the S3 region in AWS.

smarnach:
I forgot the S3 region in #2957.


willkg (approved):
Looks fine as far as I can tell.

PR #2974, opened 2024-07-29
bug-1909870: Implement a new system test suite.

smarnach:
This is a starting point for a new system test suite using pytest. You can run the tests against the local environment by running `pytest` in the `systemtests/` directory inside the Docker container. (More information in the README file included in this PR)

Some notes:

* Selecting the "prod" or "gcp_prod" environments currently skips all tests. If we don't come up with tests that run in these environments, we can remove them from the `conftests.py`.

* All uploads made by the system tests have keys that start with "tecken-system-tests-". This allows deleting all these files from the storage bucket, or creating a lifecycle rule to expire them early.

* The fake symbol file generator is very basic. It's also very slow for big files. I'm sure there are some easy improvements to make.

* The fake data generators accept random number generator seeds to allow reproducing files. The code currently uses randomized seeds. We could log these seeds to be able to reproduce files that caused problems, but I'm not sure it's worth it.

* The GCS bucket is currently only used to upload generated zip archives for the upload-by-download tests. We could also upload some real sym files or zip archives to that bucket for some tests. These files would only be properly uploaded when they are used for the first time, though, and after that all files are skipped.

* The Tecken client should log timings, but doesn't yet.

* I squashed the git history at some point because I messed up rebasing, so there's essentially just one commit for the new tests.

* Further tests we could add:
    * A test that body sizes above 2 GB are rejected for direct uploads.
    * Test with bad request methods for both upload and download endpoints

* The new system tests as well as the distinction between smoke tests and system tests  should be documented.

* The `large_files` upload-by-download test tries uploading a 3 GB file. This test causes the local environment to OOM for some reason, which shouldn't be happening. I won't have time to look into it until I'm back.


systemtests/tests/test_direct_upload.py
```diff
@@ -0,0 +1,155 @@
...
+        }
+        assert actual_tecken_headers == expected_tecken_headers
+
+        actual_storage_headers = {
+            name: response.headers[name] for name in expected_storage_headers
+        }
+        assert actual_storage_headers == expected_storage_headers
+
+    @small_archive
+    def test_code_info_lookup(
```
willkg, line 94: This test should only use Windows sym files. Linux sym files and macOS sym files don't have code file data.

systemtests/tests/test_direct_upload.py
```diff
@@ -0,0 +1,155 @@
...
+            response = tecken_client.download(
+                sym_file.code_info_key(), allow_redirects=False
+            )
+            assert response.status_code == 302
+            redirect_key = unquote(response.headers["location"])[1:]
+            assert redirect_key == sym_file.key()
+
+
+class TestDirectTryUpload:
+    # Tests within a class scope are executed in definition order.
```
willkg, line 110: I'm not sure why you wrote this. Is this implying that one test relies on another test passing?

willkg (approved):
I put in some suggested changes. Those could be approved and go with this PR--they don't fundamentally change anything.

I noted a bunch of things that should get changed. We can make those changes in follow-up PRs.

One thing I noticed is that the output of running the tests is really hard to parse. It's difficult to see where the test starts and ends so all the output kind of blurs together and I can't figure out what I'm reading. For the antenna tests, we don't log anything beyond the test reporting, so it's much easier to read by default. I think I would shut off `log_cli` and add `-vv` to the options for the default.

This is a good start and we can land this now and improve it in follow up PRs. Thank you for doing this! Structurally, it'll be a lot easier to add tests to validate specific requirements to this suite than the smoke test suite.

systemtests/lib/tecken_client.py
```diff
@@ -0,0 +1,109 @@
...
+    include_marks: set[str]
+
+    def env_var_name(self, try_storage: bool) -> str:
+        env_var_name = self.name.upper() + "_AUTH_TOKEN"
+        if try_storage:
+            env_var_name += "_TRY"
+        return env_var_name
+
+    def auth_token(self, try_storage: bool) -> str:
+        return os.environ.get(self.env_var_name(try_storage), None)
```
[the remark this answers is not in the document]
smarnach: I opted to catch the `KeyError` right in this method.

systemtests/lib/fake_data.py
```diff
@@ -0,0 +1,173 @@
...
+        self.build_id = datetime.now().strftime("%Y%m%d%H%M%S")
+
+    def key(self) -> str:
+        return f"{self.debug_file}/{self.debug_id}/{self.sym_file}"
+
+    def code_info_key(self) -> str:
+        return f"{self.code_file}/{self.code_id}/{self.sym_file}"
+
+    def header(self) -> bytes:
+        return textwrap.dedent(f"""\
```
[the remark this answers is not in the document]
willkg: I completely missed the code-file thing. Ignore what I said here. Sorry about that.

systemtests/lib/tecken_client.py
```diff
@@ -0,0 +1,109 @@
...
+        sym_file_key: str,
+        try_storage: bool = False,
+        method: str = "GET",
+        allow_redirects: bool = True,
+    ) -> Response:
+        url = f"{self.base_url}/{quote(sym_file_key)}"
+        if try_storage:
+            url += "?try"
+        LOGGER.info("downloading %s", url)
+        return self.session.request(method, url, allow_redirects=allow_redirects)
```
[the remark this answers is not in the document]
willkg: I see that now. :+1:

systemtests/tests/test_direct_upload.py
willkg: We talked about this. I want to adjust the structure here so it's easier to see that. We'll tinker with that in a followup PR.

systemtests/tests/test_direct_upload.py
willkg: We talked about this in Zoom. We're going to adjust the limiting for uploads (bug 1910613) and that might reduce the problem we're trying to work around here where we want to minimize the number of uploads as much as possible.

willkg:
I talked with @smarnach about this. Currently, uploads are rate-limited, so every time the system tests have to upload something, it greatly increases the time it takes to run the system tests. Thus the system tests are currently structured so as to minimize the number of uploads.

In bug 1910613, @smarnach is going to switch GCP and I'll do AWS to switch from rate-limiting to connection limiting. That should reduce limits on uploading especially against the stage environment where it's unlikely anyone else is uploading. After that, we can adjust the structure of these tests to be more self-contained.

Follow-up work to do:

1. figure out a structure for the tests so tests are self-contained and don't require tests to run in a certain order
2. adjust the fake-sym-file generation code so it's generating valid platform-specific symbols files
3. add relevant docstrings
4. add additional tests for things we need to validate in environments -- this will reduce the manual validation work we're doing now
5. add a "Rules of system test" to the README
6. think about changing the expiration for the stage bucket to something shorter -- if we do this, we need to adjust the clean up command so it deletes the related files in the db as well

I'll capture this in the systemtests bug.

PR #2984, opened 2024-08-02
bug-1910613: update guidance on 429s and symbols zip archive sizes

willkg:
This updates the guidance on retrying to wait 15 seconds and try again.

This also updates the discussion around which HTTP POST method to use in relation to the symbol zip archive size. We now recommend this new model where we encourage people to have symbol zip archive sizes < 2gb and they should split the archive up across multiple HTTP POSTs.

### DIFF
--- a/docs/upload.rst
+++ b/docs/upload.rst
@@ -116,8 +116,8 @@ symbols.
    that URL in the HTTP POST.
 
 
-Upload by HTTP POST (payload < 2gb size)
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Upload by HTTP POST
+~~~~~~~~~~~~~~~~~~~
 
 Include the ZIP file in the HTTP POST to ``/upload/`` as a
 ``multipart/form-data`` payload.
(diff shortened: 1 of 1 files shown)


docs/upload.rst
```diff
@@ -116,8 +116,8 @@ symbols.
    that URL in the HTTP POST.
 
 
-Upload by HTTP POST (payload < 2gb size)
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Upload by HTTP POST
+~~~~~~~~~~~~~~~~~~~
```
willkg, line 120: We no longer suggest picking one based on the size of the file. Instead we suggest breaking up large files into smaller ones regardless of the upload method.

docs/upload.rst
```diff
@@ -142,24 +142,38 @@ Here's a Python example using the ``requests`` library:
...
+Do an HTTP POST to ``/upload/`` as a ``application/x-www-form-urlencoded``
 payload and specify the url to the symbols file as a value to ``url``.
 
 Domains that Tecken will download from is specified in the
 ``DJANGO_ALLOW_UPLOAD_BY_DOWNLOAD_DOMAINS`` environment variable and at the
 time of this writing is set to::
 
     queue.taskcluster.net
-    public-artifacts.taskcluster.net
+    firefox-ci-tc.services.mozilla.com
```
willkg, line 175: I updated this to match our settings.

docs/upload.rst
```diff
@@ -333,12 +367,6 @@ filename.
 
 Records of the upload and what files were in it are available on the website.
 
-.. Note::
-
-   Symbols files can not be overwritten. Once a file is uploaded, any future
-   upload attempts for that debug_filename/debug_id combination will be
-   skipped.
```
willkg, line 340: This was never true. This was a misunderstanding I had for a long time and should have fixed this ages ago.

docs/upload.rst
```diff
@@ -314,10 +342,16 @@ Upload: /upload/
...
+   :statuscode 429: wait and retry
+   :statuscode 500: wait and retry; if retrying continues to fail, then please
        file a bug report
-   :statuscode 503: sleep for a bit and retry
+   :statuscode 503: wait and retry
+
+
+.. Note::
+
+   For retrying, we suggest waiting between 15 seconds between retry attempts.
```
relud, line 353: seems like there's an extra "between" here

```suggestion
   For retrying, we suggest waiting 15 seconds between retry attempts.
```

relud (approved):
r+wc

willkg:
Thank you!

PR #2991, opened 2024-08-14
bug-1907985: fix sym file header parsing

willkg:
This fixes sym file header parsing so it parses the entire sym file header rather than stopping after INFO CODE_ID.

While doing this, I also updated the tests for windows, mac, and linux sym file headers with dump_syms 2.3.3 output.

### DIFF
--- a/tecken/libsym.py
+++ b/tecken/libsym.py
@@ -34,21 +34,22 @@ def extract_sym_header_data(file_path):
                     data["debug_filename"] = debug_filename
                     data["debug_id"] = debug_id.upper()
 
-                elif line.startswith("INFO CODE_ID"):
+                elif line.startswith("INFO"):
                     parts = line.strip().split()
-                    # NOTE(willkg): Non-Windows module sym files don't have a code_file
-                    if len(parts) == 3:
-                        _, _, code_id = parts
-                        code_file = ""
-                    elif len(parts) == 4:
-                        _, _, code_id, code_file = parts
+                    if parts[1] == "CODE_ID":
+                        # NOTE(willkg): Non-Windows module sym files don't have a code_file
+                        if len(parts) == 3:
+                            _, _, code_id = parts
+                            code_file = ""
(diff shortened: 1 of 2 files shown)


tecken/libsym.py
```diff
@@ -34,21 +34,22 @@ def extract_sym_header_data(file_path):
...
-                elif line.startswith("INFO GENERATOR"):
-                    _, _, generator = line.strip().split(maxsplit=2)
-                    data["generator"] = generator
+                    if parts[1] == "CODE_ID":
+                        # NOTE(willkg): Non-Windows module sym files don't have a code_file
+                        if len(parts) == 3:
+                            _, _, code_id = parts
+                            code_file = ""
+                        elif len(parts) == 4:
+                            _, _, code_id, code_file = parts
```
relud, line 45: seems like this should have an else ValueError condition in case the file is corrupted or something

relud (approved):
r+wc

tecken/libsym.py
willkg: When I initially wrote the extraction code, I thought about validation a bit, but decided that's outside of the scope. This code parses the bits it understands so it can fill in some metadata. If the header is malformed or invalid in some way, we don't have any mechanisms for doing anything with that knowledge. Given that, I wrote it such that it just ignores those cases.

[Bug 1814430](https://bugzilla.mozilla.org/show_bug.cgi?id=1814430) covers writing a better information view for symbols files which would cover whether they're valid. Adding header validation to that would be helpful.

A while back, I wrote [bin/debug-sym-file.py](https://github.com/mozilla-services/tecken/blob/main/bin/debug-sym-file.py) that downloads a file and tells you things about it including whether it parses in symbolic. I used that to debug malformed symbols files for [bug 1791785](https://bugzilla.mozilla.org/show_bug.cgi?id=1791785). We could add header validation to that.

willkg:
Thank you!

PR #2992, opened 2024-08-14
bug-1909637: allow searching for files from regular and try symbol uploads

willkg:
Previously, you could sort of search for try uploads by searching for "try/" in the key. But you couldn't search for regular uploads.

We recently changed the key data in the model, so "try/" no longer appears in the key which breaks this flimsy filtering method.

This introduces a new upload_type filter field that lets you filter on nothing (""), try upload files ("try"), and regular upload files ("regular").


relud (approved):

willkg:
Thank you!

PR #2994, opened 2024-08-15
bug-1909013: remove upload now view forms

willkg:
This removes the forms from the Symbols Upload Now page. I kept the page because it contains a link to instructions on how to upload symbols using Python and curl. That seems helpful to keep.

This PR gets into a lot of react code I'm not wildly familiar with (and has no tests) so I might have missed some bits when cleaning up. However, only a handful of people in the world have the access on Mozilla Symbols to see this page, so this is pretty low risk even if we mess up.


willkg:
I opened a new PR, but CircleCI keeps the same pipeline. I'll try pushing a new commit.

PR #3005, opened 2024-09-10
bug-1909870: Make system test cases independent of each other.

smarnach:
This change makes the system test cases independent of each other. I combined the basic upload and download tests into a single test case. All other tests that need an upload simply upload a small archive first.

This change leads to a higher number of uploads when the system tests run. Since we switched to connection limiting for our Tecken deployments, this no longer leads to any rate limiting on the server side. We also introduced a lifecycle rule for our GCS buckets that expires the uploaded files early.


willkg (approved):
This is a great improvement. Nice job!

PR #3031, opened 2024-10-07
Update pyopenssl and cryptography

relud:
because dependabot is failing to do it (See logs in https://github.com/mozilla-services/tecken/network/updates/897084498)

created by removing the dependencies from requirements.txt and running make rebuildreqs


willkg (approved):
:medal_sports:

PR #3038, opened 2024-10-17
Prevent dependabot from proposing Python version upgrades beyond 3.11.

smarnach:
While we want to upgrade Python to a newer version at some point, I don't think it's useful for depednabot to suggest this to us every week.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,6 +14,9 @@ updates:
       interval: "weekly"
     rebase-strategy: "disabled"
     open-pull-requests-limit: 10
+    ignore:
+      - dependency-name: python
+        versions: [">=3.12"]
 
   # Set update schedule for GitHub Actions
   - package-ecosystem: "github-actions"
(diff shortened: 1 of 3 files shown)


.github/dependabot.yml
```diff
@@ -14,6 +14,9 @@ updates:
       interval: "weekly"
     rebase-strategy: "disabled"
     open-pull-requests-limit: 10
+    ignore:
+      - dependency-name: python
+        versions: [">=3.12"]
```
willkg, line 19: I never know what to do in situations like this where there are multiple places in the code base related to a single thing, but they're not connected.

For example, we have this ignore bit here and it's tied to the docker base image we use. When we update the docker base image, we need to remember to update this, too. Will we? What helps us remember? Should we add a comment to `docker/Dockerfile`? Should we have a runbook for updating the Python version?

willkg (approved):
I tried to verify the dependency name we should use here. I landed on this PR which has a blurb with the related ignore conditions.

https://github.com/mozilla-services/tecken/pull/3034

image

I think that suggests this change does what we want it to.

.github/dependabot.yml
smarnach: The Python version is indeed used in many places. Having a list of all these places in Dockerfile seems useful. I'll add it.

PR #3080, opened 2024-12-05
OBS-144: use docker images from obs-common

relud:


willkg (approved):
This looks good to me. I can't verify that these are the most recent tags for these images, but once we land this dependabot will update them if they aren't.

PR #3081, opened 2024-12-05
OBS-144: migrate from socorro-release to obs-common

relud:


tecken/tests/test_sentry.py
```diff
@@ -234,19 +234,16 @@ def test_count_sentry_scrub_error(metricsmock):
 
 
 def test_sentry_wrap_non_app_error_has_release():
-    port = os.environ.get("EXPOSE_SENTRY_PORT", 8090)
-    fakesentry_url = f"http://fakesentry:{port}/"
+    fakesentry_url = f"http://fakesentry:8090/"
```
relud, line 237: this shouldn't have been checking `EXPOSE_SENTRY_PORT`, because that's for use with `localhost` when outside of a docker container. when inside a container the port is always 8090.

bin/run_lint.sh
```diff
@@ -31,11 +31,11 @@ else
     if [[ -d ".git" ]]; then
         # If the .git directory exists, we can let license-check.py do
```
willkg, line 32:
```suggestion
        # If the .git directory exists, we can let license-check do
```

willkg (approved):
I can rebuild Tecken, run tests, and run the service itself. `make servicestatus` works. I checked for lingering documentation of things that were changed. Looks good!

PR #3082, opened 2024-12-06
OBS-410: Add script for graceful gunicorn shutdown.

smarnach:
https://mozilla-hub.atlassian.net/browse/OBS-410

### DIFF
--- a/bin/stop_gunicorn.sh
+++ b/bin/stop_gunicorn.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# Send the TERM signal to gunicorn to trigger a graceful shutdown, and wait for it to exit.
+
+set -euo pipefail
+
+pid=`cat /tmp/gunicorn.pid`
+
+if [ -n "$pid" ]; then
+    echo "Sending TERM signal to gunicorn."
+    kill -TERM "$pid"
+    echo "Waiting for gunicorn to exit..."
+    while kill -0 $pid 2>/dev/null; do
+        sleep 1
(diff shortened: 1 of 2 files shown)


relud (approved):

bin/stop_gunicorn.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
```
willkg, line 2: This should have the license header.

```
# This Source Code Form is subject to the terms of the Mozilla Public
# file, You can obtain one at https://mozilla.org/MPL/2.0/.
```

bin/stop_gunicorn.sh
```diff
@@ -0,0 +1,16 @@
...
+# Send the TERM signal to gunicorn to trigger a graceful shutdown, and wait for it to exit.
+
+set -euo pipefail
+
+pid=`cat /tmp/gunicorn.pid`
```
willkg, line 7: We should be using `$( ... )` instead of the legacy backticks form.

bin/stop_gunicorn.sh
```diff
@@ -0,0 +1,16 @@
...
+
+if [ -n "$pid" ]; then
+    echo "Sending TERM signal to gunicorn."
+    kill -TERM "$pid"
+    echo "Waiting for gunicorn to exit..."
+    while kill -0 $pid 2>/dev/null; do
```
willkg, line 13: `$pid` here should be quoted.

PR #3112, opened 2025-01-30
OBS-463: Use custom time as modification time for GCS objects if set.

smarnach:
https://mozilla-hub.atlassian.net/browse/OBS-463

### DIFF
--- a/tecken/ext/gcs/storage.py
+++ b/tecken/ext/gcs/storage.py
@@ -126,7 +126,7 @@ class GCSStorage(StorageBackend):
             content_encoding=blob.content_encoding,
             original_content_length=original_content_length,
             original_md5_sum=gcs_metadata.get("original_md5_hash"),
-            last_modified=blob.updated,
+            last_modified=blob.custom_time or blob.updated,
         )
         return metadata
 


relud (approved):

tecken/ext/gcs/storage.py
```diff
@@ -126,7 +126,7 @@ def get_object_metadata(self, key: str) -> Optional[ObjectMetadata]:
             content_encoding=blob.content_encoding,
             original_content_length=original_content_length,
             original_md5_sum=gcs_metadata.get("original_md5_hash"),
-            last_modified=blob.updated,
+            last_modified=blob.custom_time or blob.updated,
```
willkg, line 129: I'm not sure I understand this. For files that we migrated from S3 to GCS, `custom_time` will be set and will be the same value type as `updated` (e.g. seconds since epoch, Python datetime in UTC, etc). Is that right? What is that value type?

For files that were not migrated from S3, what does `custom_time` get set to? Does it exist as an attribute of `blob` object?

relud: migrated files will have a custom time that matches the s3 upload time and an upload time in the last week. when custom time isn't present the attribute will be None, which is false-y.

smarnach: The type of both `blob.custom_time` and `blob.updated` is `datetime.datetime | None`.

For objects directly uploaded to GCS, `blob.updated` is the time the upload finished, and `blob.custom_time` is `None`.

For objects migrated from S3, `blob.updated` is the time the object was migrated, and `blob.custom_time` is the original upload time. 

In both cases, ` `blob.custom_time or blob.updated` evaluates to the original upload time, which is what we want.

PR #3116, opened 2025-02-03
bug-1945509: switch from make to just

willkg:
This switches from make to just. While doing that, this also fixes a couple of other things:

1. fixes running the Django webapp using `manage.py runserver` when in a local dev environment; I think I broke this in 2023 when I switched the container to use Honcho to run the web and disk manager processes
2. removes the test-ci and frontend-ci services and instead uses docker-compose.override.yml to add volume mounts

Tecken should work like other crash ingestion services now.

While working through this, I pulled in `justfile` recipes from Socorro and Antenna. I fixed the volumes thing so it works the same way without the `test-ci` and `frontend-ci` containers. I updated all the documentation. I looked through the repository for "make" and updated anything relevant. I updated CI actions to match Socorro.

I went through and tested all the `justfile` recipes.

I think I caught everything. However, since I made some fundamental changes to make Tecken match Socorro and Antenna, it's possible I missed something. We'll fix things as we move forward.


relud (approved):

willkg:
Thank you!

PR #3131, opened 2025-02-24
OBS-515: Add prototype release workflow.

smarnach:
This is a lightly tested prototype for a release workflow. Creating the release can be done with a simple invocation of `gh release create`. Determining the right version tag is a bit awkward though. I wonder whether I should have reused the implementation from obs-common, instead of rewriting this in POSIX shell. This version doesn't require checking out any repos and is completely self-contained, which has advantages, too.

https://mozilla-hub.atlassian.net/browse/OBS-515

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,34 @@
+name: Draft a new production release
+
+on:
+  workflow_dispatch: {}
+
+jobs:
+  build:
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+    - name: Draft a new production release
+      run: |
+        # Determine tag
+        base_tag="$(date +v%Y.%m.%d)"
+        tag="$base_tag"
+        gh api "repos/${{ github.repository }}/tags" --paginate \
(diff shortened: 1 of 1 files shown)


relud (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,34 @@
...
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+    - name: Create new production release
+      run: |
+        # Determine tag
+        base_tag="$(date +v%Y.%m.%d)"
+        tag="$base_tag"
+        # write last 30 tags to file. We shouldn't need to request further pages.
+        gh api "repos/${{ github.repository }}/tags" | jq -r '.[].name' > all_tags
```
biancadanforth, line 18: This looks to list the tags in reverse chronological order by default, but is it worth being explicit? I couldn't find docs (or figure out how to get the manpage from the gh cli) about the sort order.

biancadanforth (approved):
LGTM! IIUC this is a workflow we would manually trigger in place of filling out the form on the GitHub Releases page for each repo as part of a prod deployment. Had one minor, non-blocking question.

PR #3153, opened 2025-03-25
bug 1954518: support tildes in debug_filename

willkg:
This adds support for the `~` character in the debug_filename field for symbol download requests. The GCS object naming docs don't list this character as an invalid character, so we should be good.

The code was written with AWS S3 conventions and requirements in mind. I skimmed the GCS naming conventions and requirements and I don't think we have new exclusions we should add. I didn't want to completely change it to match GCS requirements figuring the smallest change for this issue was better for now.

This also adds tests for the invalid key function for both the valid and invalid cases. I didn't spend a lot of time fleshing those cases out--mostly I seeded those test with what we have from the download view tests.

I also changed one of the download view tests to be parametrized rather than one big test with lots of assertions in it. I did that work and then realized I should be testing the invalid key function directly, but then thought I should keep this work around since it's an improvement.

To test:

1. build tecken
2. curl a file with a `~` in it:
   ```
   curl 'http://localhost:8000/try/libgallium-24.2.8-1ubuntu1~24.04.1.so/5BFAF14817989BCC1A495564C09FFD540/libgallium-24.2.8-1ubuntu1~24.04.1.so.sym'
   ```
   With main, you get an HTTP 400. With this change you get an HTTP 404 because the file isn't uploaded in your local dev environment.


smarnach (approved):
Looks good to me.

I'm wondering whether we could go further than this PR.

1. The naming conventions you linked are much less restrictive – essentially carriage return and line feed are the only two characters that are not allowed. The "recommendations" further down on the page are only relevant for the XML API, which we don't use, and have no intention of using.

1. Since the set of valid object names depends on the storage interface, it would make some sense to move the function to test for valid object names to `StorageBackend`. Given that we don't plan to actually support multiple storage backends going forward, refactoring the code accordingly probably isn't worth it.

1.  We could simply support any filename by encoding the filename when converting it to a storage key (e.g. URL-encoding). This would require a painful migration and result in confusing double-URL-encoded download URLs, so it's probably not worth it.

Either way, this PR improves on the current situation, and is probably all we need for the time being.

willkg:
In the PR comment, I mentioned that I tried to keep this small and focused on the specific issue. I agree it's worth rethinking this code in respects to storage buckets and other considerations for filenames in the future. That's a good future thing to do.

Thank you!

PR #3294, opened 2026-03-30
bug-2027684: Move recording of file uploads to the main thread.

smarnach:
This moves recording file uploads in the database to the main thread. This way, only the main thread needs a database connection, reducing the total number of database connections we need.

This changes Tecken's behaviour. Currently, we record an upload in the database before uploading it to object storage, and update the record with the completion time once the upload finishes. With this change, we only record uploads that actually finish. I think this change is acceptable – the database entries for unfinished file uploads don't provide much value.

The test suite is run without a thread pool; everything happens in the main thread. For this reason it's not possible to add a test that verifies that we don't open database connections from the thread pool without changing how the tests are run. Using threads in the tests has the potential to cause a lot of problems, so I think it's not worth it. Moreover, if we go ahead with our plan to change the upload API, the whole thread pool will go away in a couple of months.

### DIFF
--- a/tecken/tests/test_upload.py
+++ b/tecken/tests/test_upload.py
@@ -416,10 +416,8 @@ def test_upload_archive_one_uploaded_one_errored(
     assert upload.user == uploaderuser
     assert not upload.completed_at
 
-    assert FileUpload.objects.all().count() == 1
-    assert FileUpload.objects.get(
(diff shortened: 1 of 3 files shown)


biancadanforth (approved):

biancadanforth:
@smarnach and I paired on this together, and we were able to return the `created_at` time back to the main thread, so the DB would still have a record for file uploads that did not finish. Thank you Sven for working on this and coming up with a good solution in the meantime before bug-1652835 is done.

PR #3298, opened 2026-04-01
bug-2028541: Install PostgreSQL client in the Docker image.

smarnach:
This makes `psql` available inside the Docker image, which is useful to run ad-hoc queries against the database.

The uncompressed image size increased by 5 MB. Before the change:
```
$ docker image ls tecken:build
                                                                                                                                                                    i Info →   U  In Use
IMAGE          ID             DISK USAGE   CONTENT SIZE   EXTRA
tecken:build   847bde8b7671        775MB             0B        
```
After the change:
```
$ docker image ls tecken:build
                                                                                                                                                                    i Info →   U  In Use
IMAGE          ID             DISK USAGE   CONTENT SIZE   EXTRA
tecken:build   cf52fb1cb835        780MB             0B        
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -37,10 +37,11 @@ RUN apt-get update && \
         apt-transport-https \
         build-essential \
         curl \
-        git \
-        libpq-dev \
         gettext \
-        libffi-dev && \
+        git \
+        libffi-dev \
+        libpq-dev \
+        postgresql-client && \
     apt-get autoremove -y && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*


biancadanforth (approved):

biancadanforth (commented):
Ah, one thought: Socorro has a `psql` recipe in its justfile and also some [docs](https://github.com/mozilla-services/socorro/blob/210b2d6543190f7e0066644809dc131ae29cbc4d/justfile#L72-L74) that discuss the command. Can you add these as well?

smarnach:
> Ah, one thought: Socorro has a `psql` recipe in its justfile and also some [docs](https://github.com/mozilla-services/socorro/blob/210b2d6543190f7e0066644809dc131ae29cbc4d/justfile#L72-L74) that discuss the command. Can you add these as well?

Both the docs and the just recipe are only for local development. The recipe runs psql in the postgres container, and the first example in the docs runs it on the host machine. This PR adds psql to the application container, so it's unrelated to the recipe and the docs. It's probably a good idea to add these to Tecken as well, but that would be a separate PR.

PR #3311, opened 2026-05-07
bug-2037388: Implement the /upload/auth_info/ endpoint.

smarnach:
This adds an /upload/auth_info/ endpoint intended to be used by the upload-symbols CLI tool.

The code is mostly straight-forward. The client-otel.py module exists for two reasons. First, as a container to hold the global OTel configuration provider, so we don't have to reinitialize the `IAMCredentialsClient` on every request. Second, probably more importantly, as a way to inject a mock version of the `IAMCredentialsClient` for tests.

We will need to create new service accounts with permission to submit telemetry and give Tecken permissions to create access tokens for these service accounts, which will happen later. In the meantime, the /upload/auth_info/ endpoint won't include any OTel configuration.

As a minor deviation from the proposed JSON response in the bug, the token expiry time is returned as a Unix timestamp rather than a ISO-formatted time string. This API is meant for machine consumption, and a Unix timestamp is easier to consume for machines than a formatted string.

This PR depends on #3310. The migration needs to be deployed and manually run in both stage and prod before we can deploy this PR.

### DIFF
--- a/tecken/tests/test_upload.py
+++ b/tecken/tests/test_upload.py
@@ -19,7 +19,7 @@ from django.utils import timezone
 
 from tecken.libstorage import ObjectMetadata, StorageBackend, StorageError
 from tecken.tokens.models import Token
-from tecken.upload import utils
+from tecken.upload import client_otel, utils
 from tecken.upload.forms import UploadByDownloadForm, UploadByDownloadRemoteError
 from tecken.upload.models import Upload, FileUpload
 
@@ -1182,3 +1182,31 @@ class Test_remove_orphaned_files:
             metricsmock.assert_not_incr(
                 "tecken.remove_orphaned_files.delete_file_error"
             )
+
+
+def test_upload_auth_info(client, db, uploaderuser):
+    token = Token.objects.create(user=uploaderuser, preferred_upload_api_version=2)
+    token.permissions.add(Permission.objects.get(codename="upload_try_symbols"))
+
+    client_otel.init(
+        service_account="otel-service-account@my-gcp-project.iam.gserviceaccount.com",
+        gcp_project="my-gcp-project",
+        log_level="info",
+        iam_client=client_otel.MockIAMCredentialsClient("some-access-token"),
+    )
+
+    url = reverse("upload:upload_auth_info")
+    response = client.post(url, HTTP_AUTH_TOKEN=token.key)
+    assert response.status_code == 200
+    assert response.json() == {
+        "email": uploaderuser.email,
+        "try_symbols": True,
+        "token_expires_at": int(token.expires_at.timestamp()),
+        "upload_api_version": 2,
+        "opentelemetry": {
+            "endpoint": "https://telemetry.googleapis.com/",
+            "headers": {"Authorization": "Bearer some-access-token"},
+            "resource_attributes": {"gcp.project_id": "my-gcp-project"},
+            "log_level": "info",
+        },
+    }
--- a/tecken/tests/test_upload.py
+++ b/tecken/tests/test_upload.py
@@ -8,6 +8,7 @@ import logging
 import os
 from unittest import mock
 
+from google.cloud import iam_credentials
 from markus.testing import AnyTagValue
 import pytest
 from requests.exceptions import ConnectionError, RetryError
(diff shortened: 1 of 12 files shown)


tecken/tests/test_upload.py
```diff
@@ -1182,3 +1182,31 @@ def adjusted_getmtime(fn):
...
+        "try_symbols": True,
+        "token_expires_at": int(token.expires_at.timestamp()),
+        "upload_api_version": 2,
+        "opentelemetry": {
+            "endpoint": "https://telemetry.googleapis.com/",
+            "headers": {"Authorization": "Bearer some-access-token"},
+            "resource_attributes": {"gcp.project_id": "my-gcp-project"},
+            "log_level": "info",
+        },
+    }
```
biancadanforth, line 1243: Can you add a test for the case that the `opentelemetry` key isn't in the response? I.e. when `CLIENT_OTEL_SERVICE_ACCOUNT` is empty.

tecken/tests/test_upload.py
```diff
@@ -1182,3 +1182,31 @@ def adjusted_getmtime(fn):
...
+def test_upload_auth_info(client, db, uploaderuser):
+    token = Token.objects.create(user=uploaderuser, preferred_upload_api_version=2)
+    token.permissions.add(Permission.objects.get(codename="upload_try_symbols"))
+
+    client_otel.init(
+        service_account="otel-service-account@my-gcp-project.iam.gserviceaccount.com",
+        gcp_project="my-gcp-project",
+        log_level="info",
+        iam_client=client_otel.MockIAMCredentialsClient("some-access-token"),
+    )
```
biancadanforth: I don't see that the `client_otel.CONFIG` gets reset after this test runs -- we may want to reset this afterwards? E.g. if you were to add a test case for a response without the `opentelemetry` field.

tecken/upload/client_otel.py
```diff
@@ -0,0 +1,74 @@
...
+    iam_client=None,
+):
+    """Initialize the global OTel config provider."""
+    global CONFIG
+    CONFIG = ClientOTelConfigProvider(
+        service_account, gcp_project, log_level, iam_client
+    )
+
+
+class MockIAMCredentialsClient:
```
biancadanforth: Why isn't this in a test helper file like `conftest.py`?

tecken/upload/client_otel.py
```diff
@@ -0,0 +1,74 @@
...
+        gcp_project: str,
+        log_level: str,
+        iam_client=None,
+    ):
+        self.service_account = service_account
+        self.gcp_project = gcp_project
+        self.log_level = log_level
+        self.iam_client = iam_client or iam_credentials.IAMCredentialsClient()
+
+    def get(self) -> dict | None:
```
biancadanforth: Nit: This method doesn't ever return `None` AFAICT.

tecken/upload/client_otel.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def init(
+    service_account: str,
+    gcp_project: str,
+    log_level: str,
+    iam_client=None,
+):
+    """Initialize the global OTel config provider."""
+    global CONFIG
```
biancadanforth, line 50: TIL: Python would assign to a local variable without this.

biancadanforth (approved):
LGTM! I suggested adding one more test case, and there are some small things.

tecken/upload/client_otel.py
smarnach: I kept this here based on the precedent in https://github.com/mozilla-services/tecken/blob/main/tecken/libcache.py, but I actually agree it's a bit weird, and it's not exactly the same case as libcache. It's only used in a single test, so I'll move it thither. (The conftest.py file can't be imported in tests, so I can't move it there.)

tecken/upload/client_otel.py
smarnach: You are right. An earlier revision of the code checked whether a service account is set inside this method, so it could return `None`, and I forgot to change it.

tecken/upload/client_otel.py
smarnach: It's one of Python's design flaws that you don't declare variables. You just start using them, so the compiler has no way of figuring out whether you want to create a local variable or change a global variable. There's also the `nonlocal` statement to access variables from an outer scope that's not the global scope. JavaScript used to be like this, but they fixed it by adding variable declarations at some point. Python unfortunately didn't.

tecken/tests/test_upload.py
smarnach: Ideally tests shouldn't modify global state at all, but I agree that we should at least clean up afterwards if we do.

PR #3314, opened 2026-05-28
bug-2037414: Remove all AWS-specific code.

smarnach:
We moved out of AWS more than a year ago, and the AWS-specific code is no longer useful for us, so we don't want to maintain it, and we don't want to deal with the unused boto dependencies anymore, which have already caused problems in Dependabot requests.

The new upload API will add the capability to generate an upload URL to the `StroageBackend`. We only need this functionality for GCS, and we don't want to implement it for S3.

This PR also removes contribute.json and the endpoint to serve it. The file has been out of date for a while, and the contribute.json standard has been abandanoed.


Haseeb702 (approved):
Looks good to me.

PR #3322, opened 2026-06-01
bug-2037432: Add initiate_upload() method to the StorageBackend interface.

smarnach:
This PR is the first step towards implementing the new upload API in Tecken.

In the new API, the server no longer uploads the files directly to the storage backend. Instead, it initiates upload sessions and returns the session URL to the client, which in turn performs the actual upload. This PR adds the `initiate_upload()` method to the `StorageBackend` interface for this purpose, and it implements it for the `GCSStorage` backend – the only backend currently supported by the codebase.

The `initiate_upload()` implementation for GCS is very similar to the original `upload()` implementation. It does not accept the request body as a parameter, and calls the `create_resumable_upload_session()` instead of `upload_from_file()` on the `Blob`, but otherwise they are the same. I factored out the common code into a helper method, which can be inlined again once we remove the `upload()` method from `StorageBackend` at the end of the migration to the new upload interface.

The tests I added initially failed due to a bug in the GCS emulator. The emulator ignores the Content-Type header that is passed when the upload session is created, so I had to add code to pass it again when actually performing the upload. This won't be required when using the real GCS.

### DIFF
--- a/tecken/ext/gcs/storage.py
+++ b/tecken/ext/gcs/storage.py
@@ -22,6 +22,10 @@ class GCSStorage(StorageBackend):
     An implementation of the StorageBackend interface for Google Cloud Storage.
     """
 
+    # Upload seesions are GCS resumable uploads as documented in
+    # https://docs.cloud.google.com/storage/docs/resumable-uploads.
+    upload_session_protocol = "gcs-resumable"
+
     def __init__(
         self,
         bucket: str,
@@ -130,15 +134,10 @@ class GCSStorage(StorageBackend):
         )
         return metadata
 
-    def upload(self, key: str, body: BufferedReader, metadata: ObjectMetadata):
-        """Upload the object with the given key and body to the storage backend.
+    def _prepare_upload_blob(self, key: str, metadata: ObjectMetadata) -> storage.Blob:
+        """Helper function for upload() and initiate_upload().
 
-        :arg key: the key of the symbol file not including the prefix, i.e. the key in the format
-            ``<debug-file>/<debug-id>/<symbols-file>``.
-        :arg body: A stream yielding the symbols file contents.
-        :arg metadata: An ObjectMetadata instance with the metadata.
-
-        :raises StorageError: an unexpected backend-specific error was raised
+        This function can be inlined once we remove the upload() method.
         """
         bucket = self._get_bucket()
         blob = bucket.blob(f"{self.prefix}/{key}")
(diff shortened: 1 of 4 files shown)


tecken/ext/gcs/storage.py
```diff
@@ -22,6 +22,10 @@ class GCSStorage(StorageBackend):
     An implementation of the StorageBackend interface for Google Cloud Storage.
     """
 
+    # Upload seesions are GCS resumable uploads as documented in
```
biancadanforth: Nit: typo s/seesions/sessions

tecken/ext/gcs/storage.py
```diff
@@ -153,9 +152,43 @@ def upload(self, key: str, body: BufferedReader, metadata: ObjectMetadata):
         blob.content_encoding = metadata.content_encoding
         # Prevent "decompressive transcoding" on download. See bug 1827506.
         blob.cache_control = "no-transform"
+        return blob
+
+    def upload(self, key: str, body: BufferedReader, metadata: ObjectMetadata):
+        """Upload the object with the given key and body to the storage backend.
+
+        :arg key: the key of the symbol file not including the prefix, i.e. the key in the format
```
biancadanforth, line 160: The prefix here being either v1/ or try/v1/?

tecken/libstorage.py
```diff
@@ -38,6 +38,11 @@ class StorageBackend:
     # Whether the backend handles try symboles
     try_symbols: bool
 
+    # The name of the protocol for upload sessions started with initiate_upload. This name is
+    # passed on to clients together with the upload session URL, and each protocol name should
+    # be documented in the service documentation.
+    upload_session_protocol: ClassVar[str]
```
biancadanforth, line 44: For the documentation -- are you referring to Tecken's readthedocs? I.e. should I expect to see some doc updates in this PR, or is that covered by a later ticket?

tecken/tests/utils.py
```diff
@@ -75,6 +77,21 @@ def upload_to_backend(self, backend: StorageBackend):
         self.backend = backend
         backend.upload(self.key, BytesIO(self.body), self.metadata)
 
+    def upload_session_to_backend(self, backend: StorageBackend):
```
biancadanforth: Nit: The name if this method is a little confusing to me. I see it's modeled after `upload_to_backend`, but this makes it sound like you're uploading a session, where you're uploading a symbols file with a session URL. Maybe `upload_to_backend_with_session`? Or `resumable_upload_to_backend`?

tecken/tests/utils.py
```diff
@@ -75,6 +77,21 @@ def upload_to_backend(self, backend: StorageBackend):
...
+        self.backend = backend
+        url = backend.initiate_upload(self.key, self.metadata)
+        self.upload_to_session_url(url)
+
+    def upload_to_session_url(self, url: str):
+        # NOTE(smarnach): The GCS emulator incorrectly ignores the Content-Type header in the
+        # request initiating the upload session. As a mitigation, we pass it again in the upload
+        # request. This isn't required when using the actual GCS.
```
biancadanforth: This took me a bit to understand. IUC, the resumable uploads protocol has two phases/requests. The first one initiates the session (and includes metadata about the object like its content type) and the second one actually uploads the object to GCS (using the object content-type from the first request). The data saved in the upload ID (included in the session URL returned by the first request) is what the server uses to keep track of the upload while it's in progress.

tecken/tests/test_storage_backends.py
```diff
@@ -12,17 +12,25 @@
...
 ):
     backend = get_storage_backend(storage_kind, try_storage)
     backend.clear()
     assert backend.exists()
 
-    upload.upload_to_backend(backend)
+    if use_upload_session:
+        upload.upload_session_to_backend(backend)
+    else:
+        upload.upload_to_backend(backend)
```
biancadanforth, line 33: I know this wasn't there before, but is it worth testing the error case?

biancadanforth (approved):
Had some small nits and a couple of other minor suggestions. It's nice that we can use the GCS emulator to still test the upload by session URL even though the new upload client will do that in the future.

tecken/ext/gcs/storage.py
smarnach: In practice it is, but that's a configuration setting. [The prefix is part of the `StorageBackend` interface.](https://github.com/mozilla-services/tecken/blob/9c7156ef10c07f5ce1a9f79ddafa2a939d639ddb/tecken/libstorage.py#L35-L36)

tecken/libstorage.py
smarnach: Yes, I'm referring to Tecken's docs. We have bug-2037438 to write that documenation. It's easier for me to write documentation if I do in in one go. Moreover, this change doesn't publicly expose any changes, so currently it's not really useful to have user documentation for this.

tecken/tests/utils.py
smarnach: I noticed it's an awkward name, but wasn't able to come up with anything better. I'd like to stick with the "session" naming. GCS's resumable uploads are a specific implementation of an upload session. We could also used signed URLs for GCS, or whatever else might be available in other object storage backends. At least conceptually this interface shouldn't be specific to GCS. I'll go with `upload_to_backend_with_session`.

tecken/tests/utils.py
smarnach: With the actual GCS, you can inlcude headers like the content-type either with the request to initiate the resumable upload, or with the request to upload the data. In the emulator only the latter works, which is why I had to send the header again in the upload request.

If we wanted, we could do the same with the real GCS. We control the client side of the upload as well, after all, so we could include the header with the actual data upload, but I prefer setting the header on the server side.

smarnach: I'll try and reword the comment to make it clearer.

tecken/tests/test_storage_backends.py
smarnach: I'm not sure what to test here for the error case. The expected behaviour of `StorageBackend` implementations is to raise an exception in case of an error in the HTTP request. I think it makes more sense to have an end-to-end test for that case once we have an API endpoint actually using this, so we can test that we return a reasonable response in case of an error.

PR #3324, opened 2026-06-12
bug-2038605: set private: true and use @mozilla namespace for package names

biancadanforth:
See mozilla-services/socorro#7215

The same motivation and fixes are made here.

Some differences worth noting:
* Unlike for the Socorro PR referenced above, the root `package.json` has no `package-lock.json`, since it has no dependencies, so I didn't update/generate a `package-lock.json`.
* Also, unlike Socorro, Tecken uses `yarn` as its `/frontend` package manager, and unlike `package-lock.json`, `yarn.lock` doesn't contain these `private` and `name` fields, so `yarn.lock` doesn't need to be updated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "tecken",
+  "name": "@mozilla/tecken",
   "description": "The Mozilla Symbol Server",
   "private": true,
   "repository": {
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -1,7 +1,7 @@
 {
-  "name": "tecken",
+  "name": "@mozilla/tecken-frontend",
   "version": "0.1.0",
-  "private": false,
+  "private": true,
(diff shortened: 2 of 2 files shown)


fkiriakos07 (approved):

package.json
smarnach, line 1: I honestly don't know what this file is even doing. We may be able to simply remove it, but for now this change looks fine.

smarnach (approved):

package.json
biancadanforth: Yes good point. I don't see that it's referenced anywhere, and there are no `dependencies` at the top level. `npm publish` requires a `package.json` in the `CWD`, so it would fail in that case as well. I'll try removing it as a separate PR.

biancadanforth: I opened #3326 for removing the `package.json`. I can close or revert it that way without affecting the other changes in this PR.

PR #3325, opened 2026-06-15
bug-2046068: Clean up storage key handling.

smarnach:
This cleans up some of the code handling storage keys.

* For downloads, keys are validated directly by Django's router, so we no longer need to validate the key in the view code.
* We use the same regular expression for the Django converters as we do for validating keys in uploaded ZIP archives.
* Keys are represented as a single string as much as possible, rather than storing the three parts separately.
* The variable naming for the three parts is made more consistent (though there are still some weird names left in the test code).

This could probably be cleaned up further, but I think this is already an incremental improvements, and it cleans up the parts I need for implementing the new upload API.

### DIFF
--- a/tecken/base/symbolstorage.py
+++ b/tecken/base/symbolstorage.py
@@ -59,25 +59,10 @@ class SymbolStorage:
             return self.try_upload_backend
         return self.upload_backend
 
-    @staticmethod
-    def make_key(symbol: str, debugid: str, filename: str) -> str:
-        """Generates a symbol file key for the given identifiers.
-
-        :arg symbol:
-        :arg debugid:
-        :arg filename:
-
-        :returns: A key suitable for use with StorageBackend methods.
-        """
-        # There are some legacy use case where the debug ID might not already be
-        # uppercased. If so, we override it. Every debug ID is always in uppercase.
-        return f"{symbol}/{debugid.upper()}/{filename}"
-
     def get_metadata(
-        self, symbol: str, debugid: str, filename: str, try_storage: bool = False
+        self, key: str, try_storage: bool = False
     ) -> Optional[ObjectMetadata]:
         """Return the metadata of the symbols file if it can be found, and None otherwise."""
-        key = self.make_key(symbol=symbol, debugid=debugid, filename=filename)
         for backend in self.get_download_backends(try_storage):
             with METRICS.timer("symboldownloader_exists"):
                 metadata = backend.get_object_metadata(key)
--- a/tecken/base/utils.py
+++ b/tecken/base/utils.py
@@ -15,78 +15,56 @@ def filesizeformat(bytes):
     return dj_filesizeformat(bytes).replace("\xa0", " ")
 
 
-# See the docstring in invalid_key_name_characters for an explanation
-# of this regex.
-INVALID_CHARS_REGEX = re.compile(
-    # \x00-\x1f is the "ASCII control characters". That's 00 (Null character)
-    # until 31 (Unit separator) (whatever that is).
-    r'[\x00-\x1f\x80-\xff\\\^`><{}\[\]#%"\'\|]'
+# Characters valid in the first and last part of a symbols file key. The allowed
+# characters were originally based on what's valid in AWS S3 object keys:
+#
+# https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html
+#
+# While we are no longer using S3, and the requirements for GCS are a bit more relaxed,
+# we currently don't have a use case for relaxing our requirements.
+#
+# https://cloud.google.com/storage/docs/objects#naming
+#
+# This is the list of "characters to avoid" from the S3 docs:
+#
+# * Backslash ("\")
+# * Left curly brace ("{")
+# * Non-printable ASCII characters (128–255 decimal characters)
+# * Caret ("^")
+# * Right curly brace ("}")
+# * Percent character ("%")
+# * Grave accent / back tick ("`")
+# * Right square bracket ("]")
+# * Quotation marks
+# * 'Greater Than' symbol (">")
+# * Left square bracket ("[")
+# * 'Less Than' symbol ("<")
+# * 'Pound' character ("#")
+# * Vertical bar / pipe ("|")
+#
+# We also exlucde the slash character, since it separates the components of the key, so
+# it can't occur within a component.
+VALID_KEY_CHARS = r"[^\x00-\x1f\x80-\xff\\\^`><{}\[\]#%\"'\|/]+"
+
+# The debug id in the key can only consist of hex characters.
+VALID_HEX_CHARS = "[0-9A-Fa-f]+"
+
+# A regular expression to validate a complete symbols file key.
+VALID_KEY_REGEX = re.compile(
+    f"""
+    (?P<debug_name>{VALID_KEY_CHARS})
+    /
+    (?P<debug_id>{VALID_HEX_CHARS})
+    /
+    (?P<symbols_file>{VALID_KEY_CHARS})
+    """,
+    re.X,
 )
 
 
-def invalid_key_name_characters(key):
-    """Return true if there are characters in the key name that is
-    considered invalid.
+def validate_key(key: str) -> bool:
+    """Indicate whether the given key is valid.
 
-    Invalid characters list is based on the official S3 documentation.
-    https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html
-
-    In particular, they list a set of "Characters to Avoid". They are
-    as follows:
-
-    * Backslash ("\")
-    * Left curly brace ("{")
-    * Non-printable ASCII characters (128–255 decimal characters)
-    * Caret ("^")
-    * Right curly brace ("}")
-    * Percent character ("%")
-    * Grave accent / back tick ("`")
-    * Right square bracket ("]")
-    * Quotation marks
-    * 'Greater Than' symbol (">")
-    * Left square bracket ("[")
-    * 'Less Than' symbol ("<")
-    * 'Pound' character ("#")
-    * Vertical bar / pipe ("|")
-
-    This function can be used also to avoid or outright refuse to bother
-    looking up if we have a symbol. This is useful in the Download part
-    where we don't want clients to be able to try to break the underlying
-    XML API by attempting lookups on really "strange" control characters.
-    E.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1427730
-
-    Because this function gets used a lot, by the Download view functions,
-    it's important that this function is as fast as it can be. Therefore,
-    consider the following pre-optimization observations:
-
-    * Doing `invalid_key_name_characters('foo.pdb' + hex + 'foo.sym')`
-      is 25% slower than doing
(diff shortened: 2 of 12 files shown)


tecken/base/symbolstorage.py
```diff
@@ -59,25 +59,10 @@ def get_upload_backend(self, try_storage: bool) -> StorageBackend:
             return self.try_upload_backend
         return self.upload_backend
 
-    @staticmethod
-    def make_key(symbol: str, debugid: str, filename: str) -> str:
```
smarnach, line 63: This is basically inlined at the beginning of the `download_symbol()` view function, since that's the only place requiring it. Uppercasing the debug id is now handled by the hex converter in `urls.py`.

tecken/base/utils.py
```diff
@@ -15,78 +15,56 @@ def filesizeformat(bytes):
...
-    * 'Pound' character ("#")
-    * Vertical bar / pipe ("|")
-
-    This function can be used also to avoid or outright refuse to bother
-    looking up if we have a symbol. This is useful in the Download part
-    where we don't want clients to be able to try to break the underlying
-    XML API by attempting lookups on really "strange" control characters.
-    E.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1427730
-
-    Because this function gets used a lot, by the Download view functions,
```
smarnach, line 58: I removed the entire part about how performance-critical this is because it really isn't. This is used _once_ per download, and Tecken does not spend any significand fraction of its CPU time on evaluating this regular expression.

tecken/download/views.py
```diff
@@ -103,56 +102,41 @@ def is_maybe_codeinfo(some_file, some_id, filename):
 
     """
     return (
-        bool(filename)
```
smarnach, line 106: I removed all the `booll(…)` calls from the condition because the regular expressions used for the route already make sure all three parts are non-empty.

tecken/download/views.py
```diff
@@ -103,56 +102,41 @@ def is_maybe_codeinfo(some_file, some_id, filename):
...
+    key = f"{debug_file}/{debug_id}/{symbols_file}"
+
+    if _ignore_symbol(debug_file, debug_id, symbols_file):
+        logger.debug("Ignoring symbol %s", key)
         response = http.HttpResponseNotFound("Symbol Not Found (and ignored)")
         if request._request_debug:
             response["Debug-Time"] = 0
         return response
 
-    if invalid_key_name_characters(debugfilename + filename):
```
smarnach, line 138: The key is validated by the router, so no need to validate it here.

tecken/download/views.py
```diff
@@ -103,56 +102,41 @@ def is_maybe_codeinfo(some_file, some_id, filename):
...
-    #
-    # Not only can we avoid doing a SymbolStorage call, we also don't have to bother
-    # logging that it could not be found.
-    if _ignore_symbol(debugfilename, debugid, filename):
-        logger.debug(f"Ignoring symbol {debugfilename}/{debugid}/{filename}")
+def download_symbol(request, debug_file, debug_id, symbols_file, try_storage=False):
+    # Assemble the key from the components. The individual components have already been
+    # validated by the converters in download/urls.py, and the debug ID is already
+    # converted to uppercase.
+    key = f"{debug_file}/{debug_id}/{symbols_file}"
```
smarnach, line 126: The `debug_id` gets upper-cased by the hex converter, so we don't need to to that here.

tecken/upload/views.py
```diff
@@ -47,46 +48,38 @@
...
 # can simply be ignored.
-_ignorable_filenames = (".DS_Store",)
+_ignorable_filenames = [".DS_Store"]
 
 
-def check_symbols_archive_file_listing(file_listings):
+def check_symbols_archive_file_listing(file_listings: list[FileMember]) -> str | None:
     """return a string (the error) if there was something not as expected"""
     for file_listing in file_listings:
+        key = file_listing.name
```
biancadanforth, line 57: This looks to have also been the case before your patch, but the storage `key` here is taken directly from the file name, and its validation allows for lowercase letters in the `debug_id`, while the download validation converts the `debug_id` to uppercase. IIUC, objects in GCS are case-sensitive (so an object with the same storage key where the `debug_id` is lowercase is treated as a different object than if it were uppercased). If that's true, then this means we wouldn't find the file on download if it were uploaded in with a lowercase `debug_id`. Worth filing a separate bug?

tecken/tests/test_download.py
biancadanforth: Nit: Since we can't get empty values anymore, do we need this test case?

tecken/tests/test_download.py
```diff
@@ -160,43 +160,20 @@ def test_client_404(client, db, symbol_storage):
...
-        args=(
-            debug_filename,
-            debug_id,
-            filename,
-        ),
-    )
+def test_client_with_invalid_keys(client, url):
     response = client.get(url)
-    assert response.status_code == 400
+    assert response.status_code == 404
```
biancadanforth, line 176: Why change this to a 404? A 400 seems more accurate to me if fails validation. I get we'll be doing object look-ups, and we wouldn't find an object with an invalid name, but we're not at that part yet.

tecken/upload/views.py
```diff
@@ -47,46 +48,38 @@
...
-            f"(First unrecognized pattern was {file_listing.name})"
-        )
+        if key.rpartition("/")[2] in _ignorable_filenames:
+            continue
+        if not validate_key(key):
+            return (
+                "Unrecognized file pattern. Should only be <module>/<hex>/<file> "
+                "or <name>-symbols.txt and nothing else. "
+                f"(First unrecognized pattern was {key!r})"
+            )
```
biancadanforth, line 83: This seems less descriptive compared to the previously returned error message: "Invalid characters in ..."?

tecken/upload/views.py
```diff
@@ -47,46 +48,38 @@
...
     """return a string (the error) if there was something not as expected"""
     for file_listing in file_listings:
+        key = file_listing.name
+        # TODO(smarnach): This looks wrong to me. The documentation of
+        # `DISALLOWED_SYMBOLS_SNIPPETS` appears to imply this should be looked up in
+        # the contents of symbols files rather than in their names. I don't have any
+        # context on this, though.
         for snippet in settings.DISALLOWED_SYMBOLS_SNIPPETS:
-            if snippet in file_listing.name:
+            if snippet in key:
```
biancadanforth, line 64: Probably good to file a follow up bug and reference it here.

tecken/base/utils.py
```diff
@@ -15,78 +15,56 @@ def filesizeformat(bytes):
...
-    * Left square bracket ("[")
-    * 'Less Than' symbol ("<")
-    * 'Pound' character ("#")
-    * Vertical bar / pipe ("|")
-
-    This function can be used also to avoid or outright refuse to bother
-    looking up if we have a symbol. This is useful in the Download part
-    where we don't want clients to be able to try to break the underlying
-    XML API by attempting lookups on really "strange" control characters.
-    E.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1427730
```
biancadanforth, line 56: May be worth keeping some version of this comment around? It's still relevant IIUC and why this validation was added initially based on the `git blame` for this utility function.

tecken/urls.py
```diff
@@ -29,10 +30,10 @@ def to_url(self, value):
 
 
 class MixedCaseHexConverter:
-    regex = "[0-9A-Fa-f]+"
+    regex = VALID_HEX_CHARS
 
     def to_python(self, value):
-        return value
+        return value.upper()
```
biancadanforth, line 36: Does this change have any test coverage? Not a unit test but a new test in `test_download.py`: a download request with a lowercase `debug_id` still resolves to the stored uppercase symbol.

biancadanforth (approved):
TIL: You can validate a URL path upstream of the view code in Django.

Nothing blocking, but I have some suggestions around tests and code comments and some possible follow-up bugs to file.

tecken/upload/views.py
smarnach: Yes, this is true, and I've been aware of it for a while. If you upload a file with a lowercase debug id, you can't download it. Similarly, if you upload a file with an extension tha's not in the list of approved extensions for download, you also can't download it.

My plan was to phase this behaviour out with the new upload API. Fixing it for the current API isn't as it seems. I actually tried during the GCP migration, but eventually gave up. It also looks like all uploads we actually get use uppercase debug ids, so this isn't a problem in practice.

tecken/tests/test_download.py
smarnach: Validation now happens in the Django router, and the paths simply don't match any route now. If the router doesn't find a matching route, it will throw a 404, and I can't change that.

I also disagree that a 400 is the right status code here. The correct status code for an invalid path is a 404. A 400 is a more generic status code for any client-side error that isn't covered by one of the specific 4xx status codes.

tecken/upload/views.py
smarnach: The code checking the key used to be a bit convoluted. It split the key into three parts and then used different regular expressions for the different parts. Since there were multiple steps in the validation, the code was able to give more specific error messages. The new, simplified version of the code matches the entire key against a single regular expression, and it either matches or it doesn't. From the result we can no longer tell why the regex didn't match, so we indeed get a less specific error message.

I think this is an acceptable tradeoff in general for an error message that's basically never hit in practice. There is also another background to this change: The check for valid keys will be [performed on the client side](https://github.com/mozilla-services/upload-symbols/blob/05cbbaaac5e2f02de9bb42e8886defa1fedc2f87/crates/upload-symbols/src/sym_files.rs#L66-L73) in the future, and the error messages are even more specific than they used to be on the server side. As an example, I believe that uploading a ZIP archive with a member path that's not valid UTF-8 currently causes an unhandled exception resulting in an internal server error, but the new client-side code returns a specific error message for that case.

tecken/base/utils.py
smarnach: We no longer use the S3 XML API, and GCS has far fewer restrictions on valid characters in keys. So I don't think there's any security aspect to it anymore. As I mentioned in the new comment, we could probably relax the regular expression a lot, but since there's no necessity to do so, we simply keep the current strict checking.

tecken/tests/test_download.py
smarnach: Good catch. We don't, I removed it.

tecken/upload/views.py
smarnach: Probably, but I was too lazy to do it. I filed https://bugzilla.mozilla.org/show_bug.cgi?id=2048257 and referenced it in the comment.

tecken/urls.py
smarnach: This doesn't really change any behaviour. We used to convert the debug id to uppercase in `SymbolStorage.make_key()`, and now we do it here. The behaviour should be exactly the same as before.

We actually do have test ZIP archives with lowercase debug ids, probably unintentionally. All of that would require a lot more cleanup, but I also need to make progress on my actual goals, so I have to find a balance between cleaning up the code and implementing the new upload API.

tecken/tests/test_download.py
biancadanforth: Ah, right. It's a validation check that's also in the path.

PR #3328, opened 2026-06-23
bug-2049671: Fix missing migration test, add missing migration

smarnach:
This PR contains three small fixes:

* Add a migration we missed in #3303. The change is only about how Django sets the default value. This won't result in any actual changes on the database side. The only thing applying the migration does is convincing Django that everything is up to date.
* Fix the test for missing migrations. The current test kept passing although we were missing a migrations. The fix consists of simply removing the `--no-migrations` flag from the pytest configuration. I don't know why we had that flag there in the first place. The documentation states that it can speed up tests in some cases. I wasn't able to detect any preformance degradation for the tests.
* Calling `pytest --help` inside the application container has been broken since forever. I never considered this important enough to fix, but got annoyed by this when debugging the migration test issue, so I fixed it as well. The problem here was that `pytest --help` will import `conftest.py` to discover options, but won't set up Django before doing so, so importing models or settings would fail. moving the respective imports into the functions where they are needed fixed the issue.


tecken/upload/migrations/0025_bug_2049671_fileupload_created_at.py
```diff
@@ -0,0 +1,24 @@
...
+    dependencies = [
+        ("upload", "0024_remove_bucket_region"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="fileupload",
+            name="created_at",
+            field=models.DateTimeField(
+                db_index=True, default=django.utils.timezone.now
```
biancadanforth, line 21: In #3310 , you had another kwarg of `db_default`, mentioning that there is an ORM-level default and a DB-level default. I imagine we would also want to add a `db_default` to the ORM and recreate this migration?

But that would also mean the migration would also need to be run in the DB since that'd be a DB default change, whereas as-is, this wouldn't require running migrations in the DB?

biancadanforth (approved):
I tested this locally on all three points, and it all looks good! Had one question about possibly wanting to also set `db_default` for the migration.

tecken/upload/migrations/0025_bug_2049671_fileupload_created_at.py
smarnach: The reason I added `db_default` in that other PR was that I needed to add the migration before actually adding the model change. That way, Django didn't know about the new field yet, while the DB already did. When creating a new token, Django would not provide a value for `upload_api_version`, and without a database default this would fail.

It's rather unusual to need `db_default` though, and in this case I don't think we need it. I'll probably also remove it for the `upload_api_version` field once I flip the default to v2.

biancadanforth: Okay, thanks. Yeah the `db_default` is for direct SQL INSERT statements and things that bypass our app IIUC.

PR #3333, opened 2026-07-07
bug-2037433: Implement the upload v2 endpoint.

smarnach:
This PR implements the upload v2 endpoint as specified in the [proposal](https://docs.google.com/document/d/1aspK254WuHdEo1W8LuzSeevHQzSb87iCJ_7IaGWPQb8/edit?tab=t.0).

In the new upload protocol, clients sent requests with structured JSON data describing the files to be uploaded. I didn't want to write the validation code for the JSON manually. Using JSON schema validation seemed cumbersome, and using the Django REST framework seemed overkill, in particular if we are using it for just this one endpoint, so I settled on using the [msgspec](https://pypi.org/project/msgspec/) library, which I wasn't familiar with before, but I'm happy with how the relevant part of the code turned out. I modelled the Rust structs on the client side (see https://github.com/mozilla-services/upload-symbols/pull/27) after the msgspec `Struct` subclasses in this PR.

The `ObjectMetadata` instance returned for an existing object in GCS now always sets the `original_md5_sum` and `original_content_length` attributes, regardless of whether the file is stored gzip-compressed. For uncompressed files we simply use the length and MD5 sum of the file itself. This simplifies the code for the upload endpoint, since it can rely on the `original_*` attributes in all cases.

This PR replaces the symbols files used in tests with a more representative sample of symbols files, covering [all supported file extensions](https://github.com/mozilla-services/tecken/blob/6801656b5fd0f3263d11614b49c6194516f15c78/tecken/settings.py#L662). Before, we had three Breakpad symbols files that we sometimes uploaded with gzip compression and sometimes without, but this did not represent the actual production use, where we determine whether to compress a file based on whether it has the extension .sym, which all three of the test files had. Before this PR this was less of a problem since the actual upload endpoint was tested with the .zip sample archives. The new upload API no longer consumes ZIP archives, but operates on individual symbols files, so I needed new test data to properly cover all cases.

The tests in `test_upload_v2()` should provide complete test coverage for the new upload v2 endpoint added in this PR. We will expand these tests, or add further ones, once we implement the finalization worker that creates the database records for the file uploads.


tecken/upload/executor.py
```diff
@@ -28,3 +29,7 @@ def init(synchronous: bool, max_workers: int):
 def submit(fn: Callable[..., T], /, *args, **kwargs) -> Future[T]:
     """Submit a job to the executor."""
     return EXECUTOR.submit(fn, *args, **kwargs)
+
+
+def map(fn: Callable[..., T], /, *args, **kwargs) -> Iterator[T]:
```
biancadanforth, line 34: IIUC, you added `map` here to preserve the order of the jobs being run in other threads (which doesn't happen with `submit`)?

tecken/upload/views.py
```diff
@@ -361,3 +366,127 @@ def upload_auth_info(request):
...
+            functools.partial(initiate_file_upload, backend=backend), payload.files
+        )
+    )
+    upload_obj = Upload.objects.create(
+        user=request.user,
+        bucket_name=backend.bucket,
+        try_symbols=try_storage,
+        skipped_keys=[f.key for f in files if isinstance(f.action, ActionSkip)],
+        size=sum(f.size for f in payload.files),
+    )
```
biancadanforth, line 450: May be a concern for the finalization worker/step, but how/when will `completed_at` be populated with upload/v2?

tecken/upload/views.py
```diff
@@ -361,3 +366,127 @@ def upload_auth_info(request):
...
+
+
+@METRICS.timer_decorator("upload_v2")
+@api_require_POST
+@api_login_required
+@api_any_permission_required("upload.upload_symbols", "upload.upload_try_symbols")
+def upload_v2(request):
+    # TODO: Return a 400 on error
+    # TODO: return a 400 if there are more than 1_000 or so file specs.
+    payload = msgspec.json.decode(request.body, type=UploadRequest)
```
biancadanforth: Can you explicitly handle these 400 cases and add test coverage? Right now the request fails with a 500 IIUC.

Why cap at 1000 files? Is that coming from GCS?

tecken/upload/views.py
```diff
@@ -361,3 +366,127 @@ def upload_auth_info(request):
...
+    # to upload try symbols. We determine what the user wants to do based on the permissions of
+    # the token they used.
+    try_storage = not request.user.has_perm("upload.upload_symbols")
+    backend = symbol_storage().get_upload_backend(try_storage)
+    files = list(
+        executor.map(
+            functools.partial(initiate_file_upload, backend=backend), payload.files
+        )
+    )
+    upload_obj = Upload.objects.create(
```
biancadanforth, line 442: If `backend.initiate_upload` raises an unhandled exception for even one file, this will re-raise the exception and results in a 500 response for the whole request, right? Is that the behavior we want? It seems like this could happen with transient/network issues e.g. trying to create a resumable session with GCS. Does that get retried somewhere downstream? It'd be good to have a test case for whatever the desired behavior is here, as I didn't see one.

tecken/upload/views.py
```diff
@@ -361,3 +366,127 @@ def upload_auth_info(request):
...
+        return FileSpecResponse(key, ActionSkip())
+
+    metadata = ObjectMetadata(content_type=get_key_content_type(key))
+    if should_compressed_key(key):
+        metadata.content_encoding = "gzip"
+        metadata.original_content_length = file_spec.size
+        metadata.original_md5_sum = file_spec.md5_hash
+    else:
+        metadata.content_length = file_spec.size
+    url = backend.initiate_upload(key, metadata)
```
biancadanforth, line 496: Probably a good idea to record when this fails? The only type of error we have a metric for is invalid storage `key` (aka filename) FWICT.

Related: What's the retry policy to create a resumable upload session?

tecken/upload/views.py
```diff
@@ -361,3 +366,127 @@ def upload_auth_info(request):
...
+        executor.map(
+            functools.partial(initiate_file_upload, backend=backend), payload.files
+        )
+    )
+    upload_obj = Upload.objects.create(
+        user=request.user,
+        bucket_name=backend.bucket,
+        try_symbols=try_storage,
+        skipped_keys=[f.key for f in files if isinstance(f.action, ActionSkip)],
+        size=sum(f.size for f in payload.files),
```
biancadanforth, line 449: I don't know how easy this was for you to measure when writing your RFC, but it'd probably also be helpful to have a metric for the overall upload size and number of files in the upload? That'd help for seeing how this changes over time and sizing things like the thread pool.

Also, this looks to include the size of all files, including any that were skipped or may have errored (only reason would be an invalid `key` name IIUC). Is that what we want here?

tecken/tests/test_upload_v2.py
```diff
@@ -0,0 +1,198 @@
...
+    upload_response = perform_uploads(client, token, file_specs)
+    assert upload_response["try_symbols"] is False
+    assert upload_response["upload_protocol"] == "gcs-resumable"
+    assert upload_response["user"] == uploaderuser.email
+    for file_spec_req, file_spec in zip(
+        file_specs, upload_response["files"], strict=True
+    ):
+        assert file_spec["key"] == file_spec_req.key
+        assert file_spec["action"]["type"] == "error"
+        assert file_spec["action"]["msg"] == "invalid key"
```
biancadanforth: Aside: I noticed some small differences here between the spec and the implementation (e.g. `message` versus `msg`), the response now has an `id`, the format for `created_at` is slightly different... Good to keep track of those subtle differences between the RFC and the actual implementation from a docs perspective (or rely on the implementation, which is what you're probably planning to do).

tecken/upload/views.py
```diff
@@ -361,3 +366,127 @@ def upload_auth_info(request):
...
+    if not validate_key(key):
+        METRICS.incr("upload_file_upload_error", 1)
+        return FileSpecResponse(key, ActionError("invalid key"))
+
+    with METRICS.timer("upload_file_exists"):
+        existing_metadata = symbol_storage().get_metadata(key, backend.try_symbols)
+    if (
+        existing_metadata
+        and existing_metadata.original_content_length == file_spec.size
+        and existing_metadata.original_md5_sum == file_spec.md5_hash
```
biancadanforth, line 484: Are these case sensitive?

biancadanforth (approved):
Happy path looks good to me. Had some questions about some failure scenarios and metrics. IIUC this new route won't be used yet until we change a flag to tell the client which upload API version to use. So while I don't think any of my feedback blocks merging, it would make sense to address the error handling parts before this starts being used.

tecken/upload/executor.py
smarnach: `submit()` starts a single job, and then you get a future for the result back. So it's perfectly feasible to first create a list of futures with a for loop or a list comprehension, and then collect all the results in a second loop. This would also give you the results in the same order the jobs were submitted. The `map()` interface basically has these two loops internally, so it's just an easier way to achieve the same thing.

The `map()` interface also offers a `buffersize` to limit the number of jobs submitted at once. We could consider using that to limit the queue size, but since the number of tasks actually executed at once is limited by the size of the thread pool, there probably isn't a need to use the buffer size.

tecken/upload/views.py
smarnach: My plan is to add a number of pending uploads to each upload row in the database, and whenever an upload is finished that number gets decremented. Once it reaches zero, the upload is marked as completed. This will also give us an easy way to find uploads that were not completed, to estimate how common that is.

tecken/upload/views.py
smarnach: Oh, right, I didn't do that yet! Yes, I think at the moment it will simply through an unhandled exception for malformed requests.

The main reason to limit the number of files in a single request is preventing denial of service attacks. If we happily start processing an arbitrary number of files, we can easily hit resource limits, or end up with very long running requests.

But even for non-malicious requests we want some limit. The issues we've had with Tecken for the last decade are all a result of upload requests taking too long, and to a lesser degree of the resource consumption of upload requests and download requests differing by too many orders of magnitude. We currently sometimes do get uploads with 10,000 files. If we'd get the same number of files in a v2 request, we'd need to check whether they already exist, and then iniate an upload for each of them, so there could be 20,000 requests to GCS. Even when run in four threads, this could still take more than a minute to complete, and I simply no longer want requests that are that slow. Capping at about a 1,000 should keep requests generally under 10 seconds, which seems much more friendly for default timeouts and the general assumptions of web infrastructure.

That said, we also [cap the number of files we submit on the client side](https://github.com/mozilla-services/upload-symbols/pull/27/changes#diff-163be98fd2086e1f4cac0d9697d3798eae4379cb2a5489473c16080d778482fbR30). I decided to go with 128 as the number of files in each upload batch, which appears to be a reasonable balance between the overhead of a request and the time it takes to respond. So the server side limitation would only affect people either writing their own upload clients (unlikely) or adjusting the batch size via the `--batch-size` command-line flag. We should limit the values that flag accepts to the server side limit as well.

tecken/upload/views.py
smarnach: GCS requests fail rarely, so I thought the retry behaviour on the client side should be enough to handle this.

The current upload v1 code also doesn't retry any GCS requests, and we rarely receive Sentry tracebacks resulting from that.

tecken/upload/views.py
smarnach: We'll et a Sentry traceback if this fails. This is the same behaviour as the current v1 upload, and it looks like it's hardly ever a problem. If an error happens, it results in a 500, and the client-side retry loop will try again.

If it turns out we are getting a significant amount of these tracebacks in Sentry, we can still add metrics or retry behaviour on the server side.

tecken/upload/views.py
smarnach: Our general security model is that we trust anyone with upload permission to not send malicious requests, so maybe limiting the number of files on the client side is enough? Unauthenticated upload requests will be rejected before processing the request payload.

tecken/upload/views.py
smarnach: The size used to be the size of the ZIP archive, so it did include all skipped files as well. With the new design, the value isn't really that useful anymore, but it's still a required field, so we need to put in _some_ value. Once we drop version 1 of the upload API, we can probably remove the field from the upload.

I decided to include skipped files as well because the sum of the sizes of all actually uploaded files is readily available by simply summing the `size` fields of all associated `FileUpload` records. The only way to actually add information here is to include the sizes of other files as well.

Note that the we check for valid keys on the client side as well, so it should only be possible to have any files that are marked as `error` when using the official `upload-symbols` client tool.

tecken/tests/test_upload_v2.py
smarnach: Except for `msg` vs `message`, these differences are intentional. The timestamp format in the proposal was inconcistent with the timestamp format returned by the `/upload/auth_info/` endpoint, and also more difficult to parse on the client side than a simple Unix timestamp. The `id` is included so we can include it in the telemetry on the client side, and link all traces to the upload ids in the database.

I'll definitely base the documentation on the actual implementation. I think the proposal is still similar enough to document the design rationale of the upload API, and I don't plan to update the proposal itself.

tecken/upload/views.py
smarnach: The terminology isn't great, unfortunately. We use the terms "MD5 sum" and "MD5 hash" intercahngeably to mean "hexadecimal string representation of the MD5 message digest". The hexadecimal representation could of course be either uppercase or lowercase. We always use lowercase letters, both on the server side and on the client side.

We validate the key provided by the client, in spite of that also being validated in `upload-symbols`, so we should probably also validate the MD5 hash. I'll add code checking that the hash consists of exactly 32 lowercase hex digits, and otherwise reject the filespec. That way, things will fail loudly if we ever accidentally submit MD5 hashes that are not the same format.

GCS uses the more compact base64 representation for the MD5 hash, which would also be a reasonable choice for us, but it's probably not worth the effort to migrate to that.

tecken/upload/views.py
smarnach: Regarding whether to add metrics for this: I think this makes sense. It wasn't too bad to extract this information from the database, but metrics are definitely even easier to query. I think the metric that makes most sense is the size of actually uploaded files, so that metric needs to go in the finalization worker. The size of an "upload" is rather meaningless with the new upload API, so I wouldn't add a metric for that.

tecken/upload/views.py
biancadanforth: Probably enough, but if we do actually get 10k files in a single, authorized upload request sometimes, and we don't want that, then this is the only place that enforces that.

tecken/upload/views.py
smarnach: For now I just added a comment about removing the upload size field at some point.

PR #3338, opened 2026-08-11
bug-2037435: DB migrations for the finalization worker.

smarnach:
This PR adds two migrations that are needed by the finalization worker.

* Adding a uniqueness constraint for `FileUpload`s. Each key can occur in each upload only once. I double-checked that he current data already fulfils this constrained. The constraint only applies to `FileUpload`s that are actually attached to an upload, i.e. with `upload_id != NULL`. (We don't cascade deletes for this table, so during database cleanup it's possible to delete uploads before the corresponding file uploads, leaving `FileUpload` rows dangling.) Adding the constraint implicitly creates an index for those fields.

  <details>
  <summary>SQL used to check for uniqueness violations</summary>

  ```sql
  SELECT *
  FROM (
      SELECT
          file_upload.*,
          COUNT(*) OVER (
              PARTITION BY upload_id, key
          ) AS duplicate_count
      FROM upload_fileupload AS file_upload
      WHERE upload_id IS NOT NULL
  ) AS duplicates
  WHERE duplicate_count > 1
  ORDER BY upload_id, key, id;
  ```
  </details>

  This constraint is need for two reasons:
  * Processing messages in the finalization worker must be idempotent. This means that we can only create a new `FileUpload` record for a given (upload id, key) pair if this record doesn't exist yet. The only way to ensure that we don't create two such records concurrently is a uniqueness constraint enforced at the database level.
  * We need to be able to look up `FileUpload`s by (upload id, key) efficiently, so we need the implicit index created by the constraint.

* Adding a counter for the number of outstanding file uploads to the `Upload` model. This number is used to track whether an upload is complete. Whenever the finalization worker creates a new `FileUpload` record for an `Upload`, it will decrement the `outstanding_file_uploads` counter. The field is added as nullable, so it's safe for the v1 upload endpoint to simply ignore the field, letting it default to NULL. This also means the migration doesn't have to change existing `Upload`s at all. The field will simply be assumed to be NULL for these rows, so the migration doesn't require a full table scan.

Django originally generated this migration for adding the uniqueness constraint:
```py
class Migration(migrations.Migration):
    dependencies = [
        ("upload", "0025_bug_2049671_fileupload_created_at"),
    ]

    operations = [
        migrations.AddConstraint(
            model_name="fileupload",
            constraint=models.UniqueConstraint(
                fields=("upload", "key"), name="upload_fileupload_upload_key_unique"
            ),
        ),
    ]
```
This would have resulted in this SQL:
```sql
$ ./manage.py sqlmigrate upload 0026_fileupload_upload_key_unique
BEGIN;
--
-- Create constraint upload_fileupload_upload_key_unique on model fileupload
--
ALTER TABLE "upload_fileupload" ADD CONSTRAINT "upload_fileupload_upload_key_unique" UNIQUE ("upload_id", "key");
COMMIT;
```
According to the [Postgres 17 documentation](https://www.postgresql.org/docs/17/sql-altertable.html#SQL-ALTERTABLE-DESC-ADD-TABLE-CONSTRAINT), this command would result in an `ACCESS EXCLUSIVE` lock on the `upload_fileupload` table, blocking all other read and write requests on the table, and essentially meaning we would have to take Tecken online while the migration is running. The table contains 25 million rows, and creating the index will definitely take minutes, if not hours.

Since I did not want to deal with managing this downtime, I asked Codex to suggest a way of creating the index concurrently and rewriting the migration accordingly, and the code in this PR is what it came up with. I read the [Postgres documentation on concurrent index creation](https://www.postgresql.org/docs/17/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY), and to the best of my knowledge, this approach seems sound. It creates the index in the background as an "invalid" index, and only if the index creation succeeds, it will attach the index to the `upload_fileupload` table in a separate transaction. That separate transaction does not have to perform a full table scan and is expected to finish very quickly, so it should not affect Tecken operations. We will definitely have to test this in stage.


tecken/upload/migrations/0026_fileupload_upload_key_unique.py
```diff
@@ -0,0 +1,51 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+# Generated by Django 5.2.16 on 2026-08-10 13:18
```
biancadanforth: This is misleading, since you originally generated this file, but manually edited most of it. Can you add a comment that explains that? I.e. that the Django-generated migration would require an `ACCESS EXCLUSIVE` lock etc.

FWICT this is the only DB migration like this.

tecken/upload/migrations/0026_fileupload_upload_key_unique.py
```diff
@@ -0,0 +1,51 @@
...
+
+class Migration(migrations.Migration):
+    atomic = False
+
+    dependencies = [
+        ("upload", "0025_bug_2049671_fileupload_created_at"),
+    ]
+
+    operations = [
+        migrations.SeparateDatabaseAndState(
```
biancadanforth, line 37: Talking out loud: Since Django doesn't know how to update its state from a RunSQL command, you have to explicitly tell it what the state is after that command is run (`state_operations`).

tecken/upload/migrations/0026_fileupload_upload_key_unique.py
```diff
@@ -0,0 +1,51 @@
...
+                    ),
+                    reverse_sql=migrations.RunSQL.noop,
+                ),
+                migrations.RunSQL(
+                    sql=(
+                        'ALTER TABLE "upload_fileupload" '
+                        'ADD CONSTRAINT "upload_fileupload_upload_key_unique" '
+                        "UNIQUE USING INDEX "
+                        '"upload_fileupload_upload_key_unique"'
+                    ),
```
biancadanforth, line 55: This second operation is fast once it has a lock, but since it's ACCESS EXCLUSIVE, it has to wait for all other queries to finish before it can run. IIUC if there's a long-running query that it's waiting on, this could create a queue of other queries behind it, since Postgres processes queries in the order they come in.

Consider adding a `lock_timeout`, so this operation retries if it has to wait longer than some reasonable amount of time (perhaps 5s)?

Also, it seems unlikely you'd be running the migrations at the same time as the DB cleanup job, but it would be good to avoid that.

tecken/upload/migrations/0026_fileupload_upload_key_unique.py
```diff
@@ -0,0 +1,51 @@
...
+
+    operations = [
+        migrations.SeparateDatabaseAndState(
+            database_operations=[
+                migrations.RunSQL(
+                    sql=(
+                        "CREATE UNIQUE INDEX CONCURRENTLY "
+                        '"upload_fileupload_upload_key_unique" '
+                        'ON "upload_fileupload" ("upload_id", "key")'
+                    ),
```
biancadanforth, line 44: What happens if this step fails? E.g. if there's a duplicate `FileUpload` record in the table before the migration is run. IIUC this creates an invalid index that needs to be dropped. Maybe part of a migration "runbook" rather than anything to be done in the patch itself.

tecken/upload/models.py
```diff
@@ -101,6 +102,12 @@ class FileUpload(models.Model):
     """
 
     class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["upload", "key"],
+                name="upload_fileupload_upload_key_unique",
+            ),
```
biancadanforth, line 112: Probably worth a comment here that this is for the case where `upload != null`?

biancadanforth (approved):
LGTM, though I suggested adding some comments and making one change to include a lock timeout for the ACCESS EXCLUSIVE operation.

Definitely good to do some stage testing, and ideally not just for the happy path, though I have also confirmed in stage and prod that there are currently no duplicate `FileUpload` records, so the chances of this happening in the interim/during the migrations running seems very small.

tecken/upload/models.py
```diff
@@ -50,8 +50,9 @@ class Upload(models.Model):
     # If the upload by download URL triggered 1 or more redirects, we
     # record that trail here.
     redirect_urls = ArrayField(models.URLField(max_length=500), null=True)
-    # One increment for every attempt of processing the upload.
     created_at = models.DateTimeField(auto_now_add=True, db_index=True)
+    # The number of outstanding file uploads for this upload
+    outstanding_file_uploads = models.PositiveIntegerField(null=True)
```
biancadanforth, line 55: I assume this value will be set in the forthcoming and much anticipated finalization worker PR? And this finalization worker will also update `Upload.completed_at`? I have more questions around that, but they will likely be answered in that PR.

smarnach: We try to organize migrations in a way that they can be run either before or after deploying the code. This usually requires splitting the migration out in a separate PR. This PR only contains the model changes and the migrations. All uses of these fields are in the big finalization worker PR. In particular, it will set both `outstanding_file_uploads` and `completed_at`.

tecken/upload/migrations/0026_fileupload_upload_key_unique.py
smarnach: I'm honestly a bit out of my depth here. I don't think we generally have any long-running queries, unless someone is browsing the file upload page in the web interface (which happens extremely rarely).

My gut feeling is that if we don't add a timeout, the worst that can happen is that we try running the migration while we happen to have such a long-running query. This could queue other requests for 20 seconds or so. However, if we do add the timeout, the worst that can happen is that we hit the timeout and we have to manually clean up some database mess because Django doesn't know what to do with that half-applied migration. The latter scenario definitely feels more scary.

I think I will add a timeout, but with a relatively high value, say a minute. Then we risk a partial outage of one minute (which shouldn't cause much harm) while minimizing the chance of actually hitting that timeout and potentially creating a situation that requires manual cleanup.

tecken/upload/migrations/0026_fileupload_upload_key_unique.py
smarnach: My understanding from the [documentation](https://www.postgresql.org/docs/17/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY) is that the invalid index can't be directly dropped. The actual recommendation is to just try again (after fixing the problem that caused the failure):

> The recommended recovery method in such cases is to drop the index and try again to perform `CREATE INDEX CONCURRENTLY`.

tecken/upload/migrations/0026_fileupload_upload_key_unique.py
smarnach: The migrations generated by Django take care of both the state and the SQL. If we want to replace the SQL with our own commands, we also need to handle the state update manually, and that's exactly what `SeparateDatabaseAndState` is for.

smarnach (commented):
Thanks for the thorough review!

> Definitely good to do some stage testing, and ideally not just for the happy path, though I have also confirmed in stage and prod that there are currently no duplicate `FileUpload` records, so the chances of this happening in the interim/during the migrations running seems very small.

The code currently enforces that there can't be any duplicates within one upload, or more accurately, if there are duplicate member in the ZIP archive, [they will be deduplicated](https://github.com/mozilla-services/tecken/blob/17a3e098b59a57ff792eec55e846b4b94db21bc3/tecken/upload/utils.py#L70). So I don't think it's possible that any duplicates are created before we apply this migration.

PR #3339, opened 2026-08-14
Implement the finalization worker

smarnach:
This PR adds the finalization worker that records file uploads in the database once the client has fully uploaded them to GCS.

The worker is a long-running process implemented as a Django management command. The application logic assumes at-least-once delivery of messages and makes sure reprocessing the same message later is safe. I haven't implemented any graceful shutdown, since I don't think it's needed; if the worker gets killed in the middle of processing some message, the message will either get nacked or times out, and either way it will be resent to a different worker. If the database transaction for the message was already complete, the worker processing the message will notice and do nothing. (It's relatively straightforward to add graceful shutdown if it turns out it's actually needed.)

To make the application code as cloud-independent as possible, we introduce new `Notification` and `NotificationQueue` interfaces that abstract the pub/sub queue, and we implement these interfaces for Google Pub/Sub. The application logic of the finalization worker itself consumes `Notification` instances, which only contain the information extracted by the `NotificationQueue` implementation, so the application logic is cloud-independent.

Writing proper integration tests for this turned out to be rather tricky. We create a unique bucket in the GCS emulator per test node, so we will also need to create a pub/sub topic, a subscription and a notification config for each test. Moreover, we need a way of starting and stopping the finalization worker in the background, and a way of propagating errors from the background threads to the tests. I went through multiple iterations on this, and I think this final version is quite reasonable.

To test this in the local development  environment, you can start the local services as usual with the `slick.sh` script, which will also create auth tokens. You can then use the upload-symbols binary to upload files, e.g. by running this from the root directory of the [upload-symbols repo](https://github.com/mozilla-services/upload-symbols), using the [fake token from the distributed version of slick.sh](https://github.com/mozilla-services/tecken/blob/7477856270c3ba57f1261b1df3e9181ca1a966d4/docker/config/slick.sh-dist#L20):
```bash
export SYMBOLS_AUTH_TOKEN=c7c1f8cab79545b6a06bc4122f0eb3cb
cargo run -r -- --server-url http://localhost:8000/ --upload-api-version v2 tests/data/linux/
```


smarnach:
I added a [small commit](https://github.com/mozilla-services/tecken/pull/3339/changes/7aa29b5a00a6e20d3c038ecd4ed4a9316b74b96d) to move the finalization worker to its own container in the dev stack (and out of the Procfile).

tecken/libstorage.py
```diff
@@ -106,3 +109,48 @@ def __init__(self, msg: str, backend: StorageBackend):
...
+
+
+@dataclass
+class Notification:
+    """Notification data for a storage object event."""
+
+    # The type of the storage event. Currently supported values:
+    #     finalize_new: a new file was fully uploaded.
+    #     finalize_update: an file overwriting an existing object was fully uploaded.
+    event_type: Literal["finalize_new"] | Literal["finalize_update"]
```
biancadanforth: Style nit:
```suggestion
    event_type: Literal["finalize_new", "finalize_update"]
```

tecken/ext/gcp/utils.py
```diff
@@ -0,0 +1,45 @@
...
+
+import base64
+import binascii
+
+from tecken.libstorage import ObjectMetadata
+
+
+def build_object_metadata(
+    size: int | None, md5_hash: str, gcs_metadata: dict[str, str]
+):
```
biancadanforth: Nit: Specify the return type?

tecken/ext/gcp/utils.py
```diff
@@ -0,0 +1,45 @@
...
+    """Helper function to build an ObjectMetadata instance from information provided by GCS.
+
+    This is used both in GcsStorage.get_object_metadata() and GooglePubSub.subscribe(), so it's
```
biancadanforth: Do you mean GCSStorage and GoogleNotificationQueue?

biancadanforth:
I know it doesn't look like it, but I am working on the review. This is a very large and complex PR, and I haven't worked on Tecken in a while. I will continue tomorrow, and let's plan to do something like a live review in our 1:1 on Monday.

tecken/upload/management/commands/finalization_worker.py
```diff
@@ -0,0 +1,218 @@
...
+        # will rerun the `SELECT` to query it.
+        _, created = FileUpload.objects.get_or_create(
+            upload=upload,
+            key=notification.key,
+            defaults=dict(
+                # Since the upload is happening on the client side now, recording two timestamps
+                # here no longer makes sense, and we should remove one of them once we
+                # decommission version 1 of the upload API.
+                created_at=notification.event_time,
+                completed_at=notification.event_time,
```
biancadanforth, line 110: Are these kinds of clean-up tasks being tracked somewhere outside of a code comment (e.g. a clean-up milestone)?

tecken/libstorage.py
```diff
@@ -106,3 +109,48 @@ def __init__(self, msg: str, backend: StorageBackend):
...
+
+    # The timestamp of the event
+    event_time: datetime.datetime
+
+    # The key of the uploaded object.
+    key: str
```
biancadanforth, line 127: pubsub.py does key.removeprefix("try/").removeprefix("v1/"), so this isn't the full storage key path. Worth clarifying?

tecken/ext/gcp/pubsub.py
```diff
@@ -0,0 +1,127 @@
...
+        process: Callable[[Notification], None],
+        stop_event: Event | None = None,
+    ):
+        """Subscribe to storage object notifications.
+
+        This function calls the provided function for each received notification. If the callback
+        finishes without error, the message is automatically acknowledged.
+
+        :arg process: a function taking and processing a Notification instance.
+        :arg stop_event: an optional event used to stop the subscription gracefully.
```
biancadanforth, line 47: Worth mentioning this is a test-only argument?

tecken/ext/gcp/pubsub.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def subscribe(
+        self,
+        process: Callable[[Notification], None],
+        stop_event: Event | None = None,
+    ):
+        """Subscribe to storage object notifications.
+
+        This function calls the provided function for each received notification. If the callback
+        finishes without error, the message is automatically acknowledged.
```
biancadanforth: Can you document what happens in the error cases as well? I can see what happens in the handled exception case below, but what happens in the unhandled exception case? I assume the message times out and is redelivered? Is this related to your graceful shutdown comment in the PR description?

tecken/ext/gcp/pubsub.py
```diff
@@ -0,0 +1,127 @@
...
+        :arg stop_event: an optional event used to stop the subscription gracefully.
+        """
+
+        def callback(message):
+            """Callback called for each pub/sub message.
+
+            This function needs to be careful to not throw any exceptions for permanent errors.
+            An exception means the message won't be acked, so we will receive it again. If we
+            throw an exception for a permanent error, i.e. an error that will definitely occur
+            again if the messgae is processed again, we will create an infinite reprocessing loop.
```
biancadanforth: Isn't there a max_delivery_attempts of 10 before the message goes into the dead-letter queue? From the pubsub infra PR: mozilla/webservices-infra#12187. So we wouldn't have an infinite reprocessing loop in this case, right?

tecken/ext/gcp/pubsub.py
```diff
@@ -0,0 +1,127 @@
...
+                try_symbols = key.startswith("try/")
+                key = key.removeprefix("try/").removeprefix("v1/")
+                storage = symbol_storage()
+                for backend in storage.get_download_backends(try_symbols):
+                    if (
+                        isinstance(backend, GCSStorage)
+                        and backend.try_symbols == try_symbols
+                        and backend.bucket == message.attributes["bucketId"]
+                    ):
+                        base_url = backend.get_download_url(key)
```
biancadanforth: Technically this could make a network request as part of setting up a storage client, which could experience transient errors, but you have a thread pool that reuses storage clients, so this seems rare.

tecken/ext/gcp/pubsub.py
```diff
@@ -0,0 +1,127 @@
...
+                logger.exception("error extracting data from pub/sub message")
+                message.ack()
+                return
+            notification = Notification(
+                event_type=event_type,
+                event_time=event_time,
+                key=key,
+                metadata=metadata,
+            )
+            process(notification)
```
biancadanforth: Is it worth wrapping this in a `try..catch` block as well? Otherwise if this errors, it's unhandled, and IIUC the finalization worker process exits with a non-zero exit code. IIUC, if the finalization worker exits, it nacks the current message but also its buffer of unacked messages, which means we lose a delivery attempt on any messages in that buffer in addition to the message that was currently being processed. Is that right?

I guess that's what the dead letter queue is there to catch, but are there any likely permanent failures that could occur during this part? E.g. possible but I don't know how likely: A .sym file whose MODULE line has a debug ID longer than 40 characters

tecken/ext/gcp/utils.py
```diff
@@ -0,0 +1,45 @@
...
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import base64
+import binascii
+
+from tecken.libstorage import ObjectMetadata
+
+
+def build_object_metadata(
+    size: int | None, md5_hash: str, gcs_metadata: dict[str, str]
```
biancadanforth, line 12: Can `md5_hash` be `None`? There's an existence check for it in this function.

biancadanforth (commented):
I've reviewed some of this. Sharing my comments so far.

One overall piece of feedback (that you already separately acknowledged to me you will open a separate PR for) is to include some metrics for monitoring the finalization worker. I imagine we'd at least want to have a counter for messages that were processed successfully versus unsuccessfully (ideally with a breakdown by error category) and for how long the whole finalization process takes end to end.

tecken/ext/gcp/pubsub.py
smarnach: If an exception propagates to the pub/sub client library, the message will be "nacked" by the library. This means it will be redelivered after the current back-off period, and the pub/sub server won't have to wait for the message to time out. I expanded the docstring.

tecken/ext/gcp/pubsub.py
smarnach: It's currently test only, but as I mentioned in the PR description, it's possible we start using this in the actual code as well, if that turns out to be necessary. I think it's a natural feature for a function that otherwise doesn't return, even outside of tests. So I consider this to be a part of the interface, and the fact that it's currently only used in test code is incidental.

tecken/ext/gcp/pubsub.py
smarnach: You are right. I just didn't know that yet when I wrote this comment. I'll update it.

tecken/ext/gcp/pubsub.py
smarnach: This is the call-back we pass to the [subscribe()](https://docs.cloud.google.com/python/docs/reference/pubsub/latest/google.cloud.pubsub_v1.subscriber.client.Client?_gl=1*38krkl*_ga*MTE1NDgyMDkzLjE3NDY3MTMwNDg.*_ga_WH2QY8WWF5*czE3ODc1NzgzNzIkbzEzNiRnMSR0MTc4NzU3ODQyOCRqNCRsMCRoMA..#google_cloud_pubsub_v1_subscriber_client_Client_subscribe) method in Google's pub/sub client library. The library takes care of exceptions:

> If an exception occurs in the callback during processing, the exception is logged and the message is `nack()`ed.

I guess we could just do this ourselves here for clarity, so I'll do that.

tecken/ext/gcp/utils.py
smarnach: I don't think it can be `None`. I'll remove the check.

tecken/ext/gcp/utils.py
smarnach: Yes. I renamed the class at some point during development.

tecken/upload/management/commands/finalization_worker.py
smarnach: Not really. I added a comment to https://mozilla-hub.atlassian.net/browse/CRINGE-283 for now, but I think I'll go through all models again anyway once I work on that. (It also doesn't really matter if I miss a field that's no longer needed. We've had unused database fields in Tecken for years, and I wasn't even aware of them, so they don't really cause any problems.)

tecken/libstorage.py
smarnach: That's the way `key` is used throughout the codebase now. It used to be a mess when I took over the codebase – `key` included the prefix in about 60 precent of cases, and didn't have the prefix in the other 40 percent of cases. I standardized on storing all keys without prefix.

The wording here is confusing, though, so I'll update it.

tecken/ext/gcp/pubsub.py
smarnach: Thanks. I didn't realize that creating a GCS client could make network requests, but I guess it will contact the GCE metadata service to retrieve the credentials.

This code is running in a thread pool managed by the pub/sub client library, so I can't really run any setup code at the time of thread creation, but I can handle exceptions here and explicitly nack the message.

tecken/ext/gcp/pubsub.py
```diff
@@ -0,0 +1,142 @@
...
+            notification = Notification(
+                event_type=event_type,
+                event_time=event_time,
+                key=key,
+                metadata=metadata,
+            )
+            try:
+                process(notification)
+            except Exception:
+                message.nack()
```
biancadanforth: It's good to see this explicitly handled -- but perhaps also add a `logger.exception` to help detect/diagnose if needed?

Also consider a test case for this new path?

tecken/upload/management/commands/finalization_worker.py
```diff
@@ -0,0 +1,218 @@
...
+
+SESSION: requests.Session | None = None
+
+
+def run_finalization_worker(stop_event: Event | None = None):
+    """Consume finalization notifications until the optional stop event is set."""
+    global SESSION
+
+    logger.info("Finalization worker starting...")
+    SESSION = session_with_retries()
```
biancadanforth: IIUC when you call `subscriber.subscribe`, the underlying pubsub client uses a [`ThreadPoolExecutor` to create 10 threads by default](https://github.com/googleapis/google-cloud-python/blob/d3d684007a429dc64d8480ceda04658306c52842/packages/google-cloud-pubsub/google/cloud/pubsub_v1/subscriber/scheduler.py#L84-L87). You are currently using the same `SESSION` across all of these threads. I'm not sure if that's a problem in this case, but I don't see that `request.Session` is documented as being thread-safe.

tecken/ext/gcp/pubsub.py
```diff
@@ -0,0 +1,142 @@
...
+                        and backend.try_symbols == try_symbols
+                        and backend.bucket == message.attributes["bucketId"]
+                    ):
+                        try:
+                            base_url = backend.get_download_url(key)
+                        except Exception:
+                            # Exceptions during this call are potentially temporary, so we should
+                            # nack the message and bail out.
+                            logger.exception("exception while building download URL")
+                            message.nack()
```
biancadanforth: Consider adding a test case for this new nack?

tecken/upload/management/commands/finalization_worker.py
```diff
@@ -0,0 +1,218 @@
...
+    # connections after processing is required for the tests, but also mirrors Django's behaviour
+    # for HTTP requests.
+    close_old_connections()
+    try:
+        _process_notification(notification)
+    finally:
+        close_old_connections()
+
+
+def _process_notification(notification: Notification):
```
biancadanforth, line 53: We talked about this in our 1:1 today, but it'd be great to have some unit tests for this function. The most important one is to test for idempotency when the same message is delivered twice.

docker-compose.yml
```diff
@@ -24,12 +24,28 @@ services:
...
+        condition: service_started
       gcs-emulator: *condition_service_healthy
+      pubsub: *condition_service_healthy
       redis-cache: *condition_service_healthy
       oidcprovider: *condition_service_healthy
       statsd: *condition_service_healthy
     command: ["web"]
 
+  # Finalization worker is a prod-like background worker container.
+  finalization_worker:
```
biancadanforth, line 37: You mentioned this in our 1:1 this morning -- can you attach this container to the `just run` command, so it outputs logs to stdout like the other services do?

docker/images/gcs-emulator/Dockerfile
```diff
@@ -1 +1 @@
-FROM us-docker.pkg.dev/moz-fx-cavendish-prod/cavendish-prod/gcs-emulator:v2024.11.19-1@sha256:a02a86372ffe26c9f07cb55bf2d1551935389f0469d7e7dd0b69580879a00731
+FROM us-docker.pkg.dev/moz-fx-cavendish-prod/cavendish-prod/gcs-emulator:v2026.08.14@sha256:22e237dc4b0d1ef36cf0423edf83183920989771f5144194fb43699e12b09c58
```
biancadanforth: You'd mentioned this wasn't getting updated via Dependabot for some reason -- any idea why?  Is this something you can take care of as a follow-up?

tecken/upload/management/commands/finalization_worker.py
```diff
@@ -0,0 +1,218 @@
...
+
+    logger.info("processing upload id %s, key %s", upload_id, notification.key)
+    metadata = notification.metadata
+    sym_data = {}
+    if is_sym_file(notification.key) and metadata.download_url:
+        try:
+            prefix = download_sym_file_prefix(metadata.download_url)
+            sym_data = extract_sym_header_data(iter(prefix.splitlines()))
+        except (UnicodeDecodeError, zlib.error, SymParseError) as exc:
+            logger.debug("error while parsing .sym file: %s", exc)
```
biancadanforth: Do we see these in stage/prod? In `settings.py`, I see the log handler is set to INFO, and I don't see LOGGING_DEFAULT_LEVEL set in webservices-infra config for Tecken. I'm a bit short on time, but I can check this for myself later too in GCP.

biancadanforth (approved):
We talked about this PR for about 1.5 hours sync today. It really is a lot to get through and understand. Thank you for explaining most of the changes at a high level file-by-file and showing me how to start up and run the finalization worker locally (including those steps in the PR description would be good for posterity).

I don't see any major issues remaining. I was worried about a potentially missing `nack` for `process(notification)` in `pubsub.py`, but you addressed that in your most recent update. You cover the overall behavior well with the integration tests for the finalization worker.

As we discussed in our 1:1, the unit tests are a bit light, but I think the most important ones are for `_process_notification` (especially the idempotency test case).

- Things I'm expecting outside of this PR
	- Separate PR to cover finalization worker metrics in the app
	- Separate PR for finalization worker deployment in the infra repo with PubSub config: this is already up in https://github.com/mozilla/webservices-infra/pull/12222 and next on my list.

docker/images/gcs-emulator/Dockerfile
smarnach: I don't know why this is happening. I filed https://bugzilla.mozilla.org/show_bug.cgi?id=2066376 to track this.

tecken/upload/management/commands/finalization_worker.py
smarnach: `Session` instances are not thread-safe (though they were incorrectly documented as thread-safe at some point in the past). I made the sessions thread-local.

tecken/upload/management/commands/finalization_worker.py
smarnach: We won't see the message in stage or prod. I used `debug()` here in analogy to https://github.com/mozilla-services/tecken/blob/7477856270c3ba57f1261b1df3e9181ca1a966d4/tecken/upload/utils.py#L198 but I agree `exception()` would be better here, so we see the error in Sentry as well. If it turns out we get a lot of exceptions that we can't do anything about, we can still make it less noisy later.

tecken/ext/gcp/pubsub.py
smarnach: Thanks! This definitely needs `logger.exception()` so the exception gets sent to Sentry (in addition to being logged). I'll also add a test case.

smarnach: I added a test as well.

justfile
```diff
@@ -78,7 +75,7 @@ service-status *args: _env
     docker compose run --rm --no-deps web bash service-status {{args}}
 
 # Open psql cli.
-psql *args: _env
+psql: _env
```
biancadanforth, line 78: I guess we never used `args` in the underlying command...

justfile
```diff
@@ -29,7 +26,7 @@ setup: _env
     docker compose run --rm web bash ./bin/setup-services.sh
 
 # Run webapp and services.
-run *args='--attach web --attach frontend --attach fakesentry web frontend fakesentry': _env
+run *args='--attach web --attach frontend --attach fakesentry --attach finalization_worker web frontend fakesentry finalization_worker': _env
```
biancadanforth, line 29: I can confirm I now see the finalization_worker logs in stdout along with the other container logs. \o/

tecken/upload/management/commands/finalization_worker.py
```diff
@@ -195,7 +205,8 @@ def download_sym_file_prefix(url: str, gzipped_bytes: int = 2048) -> str:
         "Range": f"bytes=0-{gzipped_bytes - 1}",
         "Accept-Encoding": "gzip",
     }
-    with SESSION.get(url, headers=headers, stream=True) as response:
+    session = SESSION.get()
+    with session.get(url, headers=headers, stream=True) as response:
```
biancadanforth: It's a little confusing seeing two different variables, `SESSION` and `session` calling their `.get` methods in two consecutive lines. `SESSION` now is no longer a `requests.Session`.

You could use a functionally equivalent approach more similar to how you handled thread local storage clients elsewhere in Tecken as one option to make it a bit more readable, but otherwise I don't have a great idea for any alternatives. This is a readability thing not a functionality thing, so it's not blocking.

This should make it easier to add unit tests to `_process_notification` (which I see you've done in a subsequent commit).

tecken/ext/gcp/pubsub.py
```diff
@@ -141,3 +63,78 @@ def callback(message):
...
+        logger.exception("error extracting data from pub/sub message")
+        message.ack()
+        return
+    notification = Notification(
+        event_type=event_type, event_time=event_time, key=key, metadata=metadata
+    )
+    try:
+        process(notification)
+    except Exception:
+        logger.exception("exception while processing pub/sub notification")
```
biancadanforth, line 137: For all of these `logger.exception` calls, can we also include the name of the exception? E.g. `AttributeError`. That may make it easier to debug if we do end up hitting one of these exceptions. Or better still, perhaps also the storage key and upload ID if available?

tecken/upload/management/commands/finalization_worker.py
```diff
@@ -89,8 +89,8 @@ def _process_notification(notification: Notification):
         try:
             prefix = download_sym_file_prefix(metadata.download_url)
             sym_data = extract_sym_header_data(iter(prefix.splitlines()))
-        except (UnicodeDecodeError, zlib.error, SymParseError) as exc:
-            logger.debug("error while parsing .sym file: %s", exc)
+        except (UnicodeDecodeError, zlib.error, SymParseError):
+            logger.exception("error while parsing .sym file: %s")
```
biancadanforth, line 93: Nit: re-add `exc` or remove `%s`? My preference would be to include the error name in the exception log line.

Also, IIUC, Sentry will collect every log line for `logger.exception` and report a Sentry error. Is that what we want in this case? Would a warning be more appropriate?

tecken/ext/gcp/pubsub.py
biancadanforth: Nit

> If the callback throws an exception, ...

This now refers to code outside of this function. May be better as part of the docstring for `_callback`?

tecken/ext/gcp/pubsub.py
```diff
@@ -141,3 +63,78 @@ def callback(message):
             finally:
                 future.cancel()
                 future.result(timeout=10)
+
+
+def _callback(process: Callable[[Notification], None], message: Message):
```
biancadanforth: Nit: Now that this is at the module level and imported for tests, can we rename it to something more descriptive like `handle_message`?

bin/setup-services.sh
```diff
@@ -14,34 +14,7 @@ pubsub-cli create-subscription "${PUBSUB_GCP_PROJECT}" "${PUBSUB_TOPIC_NAME}" "$
...
-{
-    "topic": "projects/${PUBSUB_GCP_PROJECT}/topics/${PUBSUB_TOPIC_NAME}",
-    "event_types": ["OBJECT_FINALIZE"],
-    "payload_format": "JSON_API_V1"
-}
-EOF
-curl --fail-with-body --silent --show-error -X POST \
-    -H 'Content-Type: application/json' --data "${notification_config}" \
-    "${STORAGE_EMULATOR_HOST}/storage/v1/b/${UPLOAD_GCS_BUCKET}/notificationConfigs"
+gcs-cli notification "${UPLOAD_GCS_BUCKET}" "${PUBSUB_GCP_PROJECT}" "${PUBSUB_TOPIC_NAME}"
```
biancadanforth, line 17: Yay!

tecken/tests/test_pubsub.py
```diff
@@ -0,0 +1,147 @@
...
+    with pubsub_v1.SubscriberClient() as subscriber:
+        subscription_path = subscriber.subscription_path(
+            settings.PUBSUB_GCP_PROJECT, gcs_pubsub_subscription
+        )
+        (received,) = subscriber.pull(
+            subscription=subscription_path, max_messages=2
+        ).received_messages
+        process_mock = Mock(side_effect=process_error)
+        message = MessageWrapper(subscriber, subscription_path, received)
+        pubsub._callback(process_mock, cast(Message, message))
```
biancadanforth: Good idea to base this on the notification structure from the library in case it ever changes.

tecken/tests/test_pubsub.py
```diff
@@ -0,0 +1,147 @@
...
+    expect_call: bool = True,
+    process_error: Exception | None = None,
+) -> tuple[MessageWrapper, Notification | None]:
+    with pubsub_v1.SubscriberClient() as subscriber:
+        subscription_path = subscriber.subscription_path(
+            settings.PUBSUB_GCP_PROJECT, gcs_pubsub_subscription
+        )
+        (received,) = subscriber.pull(
+            subscription=subscription_path, max_messages=2
+        ).received_messages
```
biancadanforth, line 70: Is there risk of a race condition where the messages haven't yet been pushed to the queue? Worried about intermittent test failures with indirect/misleading errors.

tecken/tests/test_pubsub.py
```diff
@@ -0,0 +1,147 @@
...
+    assert metadata.upload_id is None
+
+    # Upload the file again to verify that updates are detected correctly.
+    upload.upload(symbol_storage)
+    message, notification = pull_and_process(settings, gcs_pubsub_subscription)
+
+    assert notification.event_type == "finalize_update"
+
+
+def test_callback_backend_error(
```
biancadanforth, line 116: Consider also asserting the log calls?

tecken/tests/test_finalization_worker.py
```diff
@@ -0,0 +1,60 @@
...
+):
+    upload = Upload.objects.create(
+        user=fakeuser,
+        size=200,
+        outstanding_file_uploads=2,
+        bucket_name="publicbucket",
+    )
+
+    # Send notifications for two uploaded files.
+    for i, key in enumerate(["a/b/c", "d/e/f"]):
```
biancadanforth, line 29: Consider including a test for the sym file path as well, since that code is pretty complicated in parsing the header and adds additional data to the DB?

tecken/tests/test_finalization_worker.py
```diff
@@ -0,0 +1,60 @@
...
+import pytest
+
+from tecken.libstorage import Notification, ObjectMetadata
+from tecken.upload.management.commands.finalization_worker import _process_notification
+from tecken.upload.models import FileUpload, Upload
+
+
+@pytest.mark.parametrize("event_type", ["finalize_new", "finalize_update"])
+@pytest.mark.django_db
+def test_process_notification(
```
biancadanforth, line 18: It'd probably be good to have a few more test cases for some of the other non-happy paths -- but especially the one where upload_id is None, since that tests the behavior during the v1 to v2 transition., and IIUC wouldn't get to Sentry. What do you think?

biancadanforth (commented):
Test code feedback -- some requests to add a few more test cases/assertions and one question about a possible test race condition. Otherwise some nits and suggestions to include more information in the log lines for debugging purposes.

tecken/ext/gcp/pubsub.py
smarnach: `logger.exception()` already logs the entire traceback of the exception, including its name. The exeception is also reported to Sentry, which will receive the values of local variables as well.

tecken/upload/management/commands/finalization_worker.py
smarnach: See above – `logger.exception()` already logs a full traceback of the current exception, which is why I removed the `exc` variable.

While warnings end up in the logs, we generally won't go looking for them, so we won't see them. If we regularly get errors here, I'd like to understand what they are. If there are some we can't do anything about, we can silence these errors later. So in the first version I'd like this to be noisy, since I currently don't know why this would regularly error.

tecken/tests/test_pubsub.py
smarnach: I was worried about that as well, but `pull()` waits for up to a minute for a message to arrive if none is immeditely available, and only after that timeout it will return an empty response. The timeout is configurable, and there's also a `return_immediately` flag that can be set, but I figured that the defaults are fine.

tecken/tests/test_finalization_worker.py
smarnach: I'm generally not aiming for 100 percent test coverage. Full test coverage requires writing a lot of tedious tests, which takes time, and in my experience it also doesn't help future maintenance work, but rather slows it down, so it's kind of a lose-lose approach.

Before writing a new test case, I'm trying to judge whether it's actually worth it. Of course this can't be an exact science. These are some questions I tend to ask myself when deciding whether to test a specific code path:

* Which parts of the behaviour do we actually care about?
* How complex is the code?
* What will happen if the code path breaks?

For the example of `upload_id is None`, the answers to these questions are

* If `upload_id` is unset, we want the finalization worker to do nothing. We don't care about the exact way it does nothing.
* The code for that case is trivial – log a message and return.
* If the code path ever breaks, it can at worst log an error. Without an upload id, it _can't_ do any of its work, even if it tries.

So for this particular case, the _value_ of the test seems rather low. On the other hand, the _cost_ of that particular test is low as well – easy to write, unlikely to cause maintenance problems, fast run time – so we might just as well add it. Which I will go ahead an do now, but I'll basically only test it returns without error in that case.

I don't think tests for the `except` branches in `_process_notification()` provide much value either, but they are more cumbersome to write, so I'll omit them.

tecken/tests/test_finalization_worker.py
smarnach: The sym file path is covered by the integration tests, which feels like the right place.

Testing that path here is a bit tricky. We'd basically have two options:

1. Upload actual objects to the bucket so we can download the first few kilobytes of them. This would make the test more of an integration test, which we already have.
2. Use the `requestmock` fixture to mock the download. For this to be useful in any way, we'd need to simulate a lot of behaviour – range requests, gzip encoding, streaming reads. I'm not sure the mock can do that, but even if it can, I'm not sure how much real behaviour we'd even be testing.

Overall it seems best to cover the header parsing only in the integration tests. The assertions for the data extracted from the header should in the integration tests should be improved, though, so I'll do that.

tecken/upload/management/commands/finalization_worker.py
smarnach: I'll rename `SESSION` to `THREAD_LOCAL_SESSIONS`, and I think that's enough to remove the confusion.

tecken/ext/gcp/pubsub.py
smarnach: The docstring explains the interface of the `subscribe()` method. The word "callback" here refers to the `process` parameter of the method. I'll clarify that.

tecken/ext/gcp/pubsub.py
smarnach: Names don't need to be globally descriptive. They only need to be descriptive within their context and namespace. As an example, basically all client classes in the Google Cloud Python client libraries are called `Client`, which isn't a globally descriptive name:

* https://docs.cloud.google.com/python/docs/reference/pubsub/latest/google.cloud.pubsub_v1.subscriber.client.Client
* https://docs.cloud.google.com/python/docs/reference/pubsub/latest/google.cloud.pubsub_v1.subscriber.client.Client
* https://docs.cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.client.Client

This is a naming convention that's often recommended by API guidelines, and a convention I like. It's always possible to rename items during import with `from google.cloud.storage import Client as StorageClient` if the name would otherwise be confusing, or to import the module rather than the individual item, so the name (e.g. `storage.Client`) is still qualified when it's actually used.

The `_callback()` function is generally considered private to the module, as indicated by the leading underscore, and within the context that name is descriptive. In the tests it's called as `pubsub._callback()`, which also seems descriptive. I'd like to keep the `callback` part in the name, so I'll rename it to `_subscription_callback()`, which is the only more descriptive name I can think of under this constraint.

tecken/tests/test_pubsub.py
smarnach: Yeah, I first tried fabricating messages, which quickly got cumbersome, so I decided to leave that part to the GCS emulator, which already has an implementation to fabricate these messages that's maintained by someone else.

tecken/tests/test_pubsub.py
smarnach: We only have log calls in places where we log exceptions (which also sends these exceptions to Sentry). I don't care about the log messages that much, and care more about the exceptions being sent to Sentry. We have this `fillmore` Sentry fixture thing, so I'll try that out to test that the exceptions are reported as expected.

biancadanforth: Ah true. Previously they were log.debug, but that didn't seem to show up in stage/prod logs, so you made them exceptions which based on Sentry config go to Sentry. That influences a number of similar comment I made in my review pass yesterday.

biancadanforth (commented):
I reviewed the commits added since my third review pass yesterday (thanks for adding them as separate commits -- that makes it much easier to review). The changes largely cover improving names of variables/functions and increasing test coverage. It all looks good to me. My previous R+ stands!

tecken/ext/gcp/pubsub.py
```diff
@@ -0,0 +1,140 @@
...
+
+        :arg process: a function taking and processing a Notification instance.
+        :arg stop_event: an optional event used to stop the subscription gracefully.
+        """
+        with pubsub_v1.SubscriberClient() as subscriber:
+            subscription_path = subscriber.subscription_path(
+                self.project_id, self.subscription_name
+            )
+            callback = functools.partial(_subscription_callback, process)
+            future = subscriber.subscribe(subscription_path, callback=callback)
```
biancadanforth, line 54: One thing to consider -- set flow_control to be closer to the max number of threads processing messages simultaneously (I think 10?). The default is 1000, and if there's no graceful shutdown and the worker is killed, any messages queued up in the client lose a delivery attempt IIUC.

smarnach: I can't really see a situation where we will have anything close to 1,000, or even just 100, messages in the queue. On average, we receive like one file upload every two seconds. Of course the distribution is anything but uniform, and during spikes it can easily be tens per second, but creating database records is really fast, so almost always the messages are trickling in much slower than they are processed. Having _any_ messages queued up in a worker will be an exceptional situation, so most of the time killing a worker won't consume any delivery attempts. The maximum number of delivery attempts is set to ten, which I expect to be generous. And if it turns out that I'm wrong, we'll catch any messages that fail more than ten times in the dead-letter queue, so we'll notice and can do something about it.

I should probably just implement graceful shutdown – we already have the `stop_event` parameter to facilitate that, and all that's missing is a signal handler. I won't do that in this PR, though, but rather in a follow-up.

